PR #2, opened 2018-12-19
Add መጽሐፈ፡ አርድእት፡

smaugustine:
Budge includes a negative facsimile of ff. 1-25 in his book. As he notes in the introduction and as one can see in the facsimile, after the ጸሎት፡ ዘመንገደ፡ ሰማይ፡, a part of ልፋፈ፡ ጽድቅ፡, but before the ትርጓሜ፡ ፊደል፡ (? this is absent in the facsimile) is መጽሐፈ፡ አርድእት፡. There are three entries for this text in BM, but I can't tell which is which since they don't have the text, so I went with the most generic.

### DIFF
--- a/LondonBritishLibrary/add/BLadd16204.xml
+++ b/LondonBritishLibrary/add/BLadd16204.xml
@@ -46,15 +46,24 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         <summary/>
                         
                         <msItem xml:id="ms_i1">
-                            <locus from="1" to="25"/>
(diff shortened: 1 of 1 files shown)


thea-m:
Hi @smaugustine ! Sorry for getting back to you so late.
Yes, you made the right choice, the generic records are created for cases just as this, where the exact recension is not known to the encoder. For the Maṣḥafa ʾardəʾət, there would be an even more generic record (LIT1915Mashaf) in cases where it is not known whether it is the Christian or Beta ʾƎsrāʾel recension.

PR #18, opened 2019-09-17
various small changes

thea-m:
various small changes in d'Abbadie 203 and other manuscripts from the last weeks


DenisNosnitsin1970:
ok.

PR #24, opened 2019-09-25
Additions of miniatures, varia and notes, with some corrections

Ralph-Lee-UK:
Additions of miniatures, varia and notes, with some corrections to EMIP 151-162, and EMIP 163-186 (some just minor edits)


PietroLiuzzo:
I did not actually check but I do not want to hold your work as I will be away until wednesday

PR #32, opened 2019-10-22
Emip metadata ralph

Ralph-Lee-UK:


EMIP/1001-2000/EMIP01950.xml
```diff
@@ -57,6 +64,35 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                    <foreign xml:lang="gez">ማርያም፡</foreign>
+                    <foreign xml:lang="gez">ሚካኤል፡</foreign>
+                    <foreign xml:lang="gez">ገብርኤል፡</foreign>
+                    <foreign xml:lang="gez">መስቀል፡</foreign>
+                    <foreign xml:lang="gez">ስላሤ፡</foreign>
+
+                  </seg>
+                </handNote></handDesc>
+                  <decoDesc>
+                    <decoNote type="frame" xml:id="d01">
```
PietroLiuzzo: Sorry if I did not notice this before. I am not sure about the reason for which here a 2 digit number is used, so d01 instead of simply d1

EMIP/1001-2000/EMIP01950.xml
```diff
@@ -57,6 +64,35 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                      Decorative designs: <locus target="#1r"/> (haräg);
+                      <locus target="#7v"/> (multiple full stops and a line of alternating red and black dots);
+                      <locus target="#31v #34r #43r #48r #59v #68r #71v #80r #81v #83r #83v #88r"/> (multiple full stop</decoNote>
+                  </decoDesc>
+                  <additions><list>
+                    <item xml:id="a01"><desc><locus target="#iir"/> Excerpt from various prayers used as pen trials.</desc></item>
```
PietroLiuzzo: same note as above about the xml:id a01, which should be a1 only

PietroLiuzzo: let me know if this has been done systematically and I will fix is throughout myself

Ralph-Lee-UK: I am pretty sure that I have done it like this since the start, yes! I have put this right in the spreadsheets that I use to generate code, so it shouldn't happen again! Thanks for all your support

Ralph-Lee-UK: I think that we have all done this for "d" "a" and "e" xml:ids

PietroLiuzzo:
Dear @Ralph-Lee-UK could you please remove the .gitignore from your PR? if you made changes to those files, that you do not want to commit, simply do not commit them and move them to another branch or your local file system.

PR #34, opened 2019-10-24
EMIP 1972/3/4

jonahsandford:
Added quire maps, margin dim, columns/lines, binding desc, dating info


PietroLiuzzo:
It would be way better to wait until you are finished with a task before opening a PR, instead of opening one and then committing changes to it afterwards

EMIP/1001-2000/EMIP01974.xml
```diff
@@ -48,19 +49,168 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                            <watermark/>
                         </support>
                         <extent>
-                           <measure unit="leaf"/>
+                           <measure unit="leaf">iv + 184</measure>
```
PietroLiuzzo: measure should contain only the number, not the letter and signs. https://betamasaheft.eu/Guidelines/?id=objectDescription 
this applies to all entries

EMIP/1001-2000/EMIP01974.xml
```diff
@@ -48,19 +49,168 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                           <measure unit="quire">Protection quire + 18</measure>
```
PietroLiuzzo: as in previous comment, measure should only be used for the figures

EMIP/1001-2000/EMIP01974.xml
```diff
@@ -48,19 +49,168 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                         <locus from="1r" to="165r"></locus>
+                       </layout>
+                       <layout columns="2" writtenLines="19">
+                         <locus from="165v" to="182r"></locus>
+                       </layout>
+                     </layoutDesc>
                   </objectDesc>
+                  <bindingDesc>
+                    <binding xml:id="binding" contemporary="true">
+                      <decoNote xml:id="b1">Parchment, 155 x 108 x 65 mm, four Coptic chain stitches attached with bridle attachments to rough-hewn boards of the traditional wood covered with tooled leather, headband and tailband.</decoNote>
```
PietroLiuzzo: I am not sure why material and measure are repeated here

EMIP/1001-2000/EMIP01975.xml
```diff
@@ -48,19 +49,117 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                         <locus from="1r" to="145v"></locus>
+                       </layout>
+                       <layout columns="2" writtenLines="20 22">
+                         <locus from="145v" to="161r"></locus>
+                       </layout>
+                     </layoutDesc>
                   </objectDesc>
+                  <bindingDesc>
+                    <binding xml:id="binding" contemporary="true">
+                      <decoNote xml:id="b1">Parchment, 123 x 100 x 80 mm, four Coptic chain stitches attached with bridle attachments to rough-hewn boards of the traditional wood.</decoNote>
```
PietroLiuzzo: see previous comment, the same applies to all files

jonahsandford:
Hi Pietro,

Thanks for your feedback. Regarding the inclusion of letters/symbols in the
<measure unit="leaf"/> code, I have included more than just numbers because
in our catalogues the leaves in the protection quires are not numbered in
the sense that the following leaves are. So a manuscript may have content
starting on page 1, but if there is a protection quire with four pages
those would be numbered as i-iv (like front matter for a book would be
separately numbered). Is that okay?

As for the prose in the binding descriptions, I am copying/pasting that
from our catalogues to give more detail. However, the measurements should
probably be taken out of that, so I'll go back in and correct those
portions.

Thank you,

--
Jonah Sandford, MATS
Adjunct Professor of Greek
Portland Seminary
Phone: (208) 651-0225

PR #68, opened 2019-11-15
Emip metadata

Ralph-Lee-UK:
First of Tweed collection to try image links.


PietroLiuzzo:
dear @Ralph-Lee-UK, I hope this is going to move now to a common emip branch. 
Please only make a PR when ready. commits to the branch with the PR after the creation of the PR should be corrections only. thank you

PR #74, opened 2019-11-20
check

nafisa-valieva:


nafisa-valieva:
but it is not what I want to pull

PietroLiuzzo:
this file has been updated and renamed on the master branch. this generates a conflict which is not solvable here.

PietroLiuzzo:
what do you want to send as a pull request then?

nafisa-valieva:
I want to have my 'second start', but it is on my laptop only! I am trying to pull it but it says there is a fatal error: unable to access ....

PietroLiuzzo:
I cannot help you this way. do you have a land line to call if skype does not work? this internet troubles affect your computer as well? that might be the core issue to solve. you need some connection at least.

PietroLiuzzo:
you can call me in my office, you know the extension or can find it in my emails

PR #83, opened 2019-11-22
Update ESqsm017.xml (introduced msParts, IDs, corrected collation)

eu-genia:
No IDs for names supplied yet
Miracles (162 + 1, treated as 163 by E-S but wrongly as these are two MSS (four actually)) are not identified in E-S

### DIFF
--- a/ES/ESdmb003.xml
+++ b/ES/ESdmb003.xml
@@ -45,7 +45,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                      25vb-46vb)&#xD; &#xD; Colophon.</summary>
                   <msItem xml:id="ms_i1">
                      <locus from="6ra" to="106va"/>
-                     <title type="complete" ref="LIT5673AkkonuColl "/>
+                     <title type="complete" ref="LIT5673AkkonuColl"/>
                      <textLang mainLang="gez"/>
                                           
                      <msItem xml:id="ms_i1.1">
--- a/ES/ESdmb003.xml
+++ b/ES/ESdmb003.xml
@@ -44,51 +44,97 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                      bǝʾsi ... "Does not the man ..." (distributed among the miracles of Mary, fols.
                      25vb-46vb)&#xD; &#xD; Colophon.</summary>
                   <msItem xml:id="ms_i1">
+                     <locus from="6ra" to="106va"/>
+                     <title type="complete" ref="LIT2384Taamme"/>
+                     <textLang mainLang="gez"/>
+                                          
+                     <msItem xml:id="ms_i1.1">
+                        <locus from="6ra" to="10rb"/>
+                        <title>Maṣḥafa śǝrʿāt and hymns</title>
+                        <textLang mainLang="gez"/>
+                        
+                        <msItem xml:id="ms_i1.1.1">
+                           <locus from="6ra" to="9rb"/>
+                           <title type="complete" ref="LIT1968Mashaf">Maṣḥafa śǝrʿāt, Introductory rite of al-Muʿallaqah</title>
+                           <textLang mainLang="gez"/>
+                        </msItem>
+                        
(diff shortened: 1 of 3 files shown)


DenisNosnitsin1970:
Miracle 163 is written by the same scribe who added a text on f. 5 r-v (a part of the 3 leaf protective quire in the front, into which miniatures were incerted), but I do not think the last quire was added later since the main text ends on its 1st leaf. That scribe added a "layer of writing", but not a production unit.

eu-genia:
I know if you see the description f. 5r-v and 253-256 are described together as one msPart and one hand

eu-genia:
oh i see it was 254 the beginning not 253 have to correct

ES/ESqsm017.xml
```diff
@@ -46,30 +46,15 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
-                     </explicit>
-                     <note/>
-                  </msItem>
-                  <!--The data in the items below refers to other items. please remove this items and move the information to the appropriate place.-->
-                  <msItem xml:id="ms_i2">
-                     <note>Maṣḥafa śǝrʿāt and Introductory exhortation are repeated twice.</note>
-                     <note>Significant portions of the text (omissions or amendments?) are written
-                        in the margin of fols.9v-10r, 12v-13r.</note>
+                     <locus from="5ra" to="256vb"/>
+                     <title ref="LIT2384Taamme">Taʾāmra Māryām "Miracles of Mary"</title>
```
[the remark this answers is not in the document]
eu-genia: ok

eu-genia:
added two more updated MSS - ESdmb003 and ESqsm008 - we need ESdmb003 and ESqsm017 well accessible because of the postcard print

ES/ESdmb003.xml
```diff
@@ -44,51 +44,97 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                     
+                     <msItem xml:id="ms_i1.3">
+                        <locus from="9ra" to="10ra"/>
+                        <title type="complete" ref="LIT3058RepCh338"/>
+                        <textLang mainLang="gez"/>
+                     </msItem>
+                     
+                     <msItem xml:id="ms_i1.4">
+                        <locus from="12va" to="106va"/>
+                        <title type="complete" ref="LIT3585Taamme">103 Miracles of Mary</title>
```
thea-m: LIT5673AkkonuColl (this would also solve the problem of having two msItems for the miracles and Akko-nu besi hymns

ES/ESqsm008.xml
```diff
@@ -44,14 +44,42 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                     
+                     <msItem xml:id="ms_i1.3">
+                        <locus from="8vc" to="9vc"/>
+                        <title type="complete" ref="LIT3058RepCh338"/>
+                        <textLang mainLang="gez"/>
+                     </msItem>
+                     
+                     <msItem xml:id="ms_i1.4">
+                        <locus from="10ra" to="156vc"/>
+                        <title type="complete" ref="LIT3585Taamme">179 Miracles of Mary</title>
```
thea-m: see above

ES/ESdmb003.xml
eu-genia: but here for 1.4 or above for 1? the record LIT5673AkkonuColl  seems also to have the intro parts

thea-m: I would add it as an msItem in ms_i1. Then the introductory textparts can still have their individual IDs within LIT5673AkkonuColl , and the miracles not belonging to this collection follow afterwards

eu-genia: done (instead of LIT2384Taamme)

eu-genia: please check also: LIT5673AkkonuColl says that the first stanza is followed by LIT5052SobaAshafa

Here it is followed by the very similar but seemingly different (?) LIT5049SobaA

PR #98, opened 2019-11-30
added title and keywords

thea-m:


thea-m:
I have copied the titles as given in the description by @ElianaDalSasso (to close https://github.com/BetaMasaheft/Documentation/issues/1275), since I don't have the catalogue. I would assume it to be the Relevation of John.

PR #124, opened 2019-12-13
Emip metadata

Ashlee-Benson:


PietroLiuzzo:
I am really not sure why this is being made in a PR again. the changes look the same as in the previous one.

PietroLiuzzo:
perhaps you have not update your branches after the merge I did not long ago?

Ashlee-Benson:
I update the branch at least once a day, at the start of the work day and
sometimes midway through.  Today I updated the branch before I started
working, which was about 5 hours ago.  Should I be updating the branch more
than once a day?

Please advise.

Thank you,

Ashlee

PietroLiuzzo:
yes, this looks like it is still the same stuff. There is something not right at all in the way this is being done, which I think has to do with the following. 
If you make a Pull Request, you should typically wait until this is approved and do changes only if requested by the review. My best guess is that while the previous PR was on hold, you have continued working on EMIP Metadata and now you want to commit those changes, which were not part of the previous PR, before having updated you branch from master, which includes the corrections I have made to the non valid EMIP 397. 
It is definitely not a matter of how often to update, it is a matter of understanding how this works.
The above, if this is what happened, cannot work well for several logical reasons.
Now, what I suggested also to  @Ralph-Lee-UK and @jonahsandford is that you all use the same branch and commit to that. if you do so, you can commit as much as you like to that branch, but then, once you decided to issue a PR with a reviewable number of edits, you have to wait for this to be reviewed and for the necessary edits to be requested. When this will happen, the branch will be deleted, you will update master, make a new branch and start a new session of work. I understand that this makes you have to wait a bit, as in any editorial process, however you cannot expect me to be always and only reviewing your PRs. If you really cannot wait, you can still: make another branch or continue editing and wait to commit to have a new branch.

PietroLiuzzo:
I will decide what to do with this PR next week, and after having received confirmation from all of you independently that this is clear. as Ralph knows I have also offered a specific session on this workflow to support you.

PietroLiuzzo:
Please note that this issue is entirely unrelated from
- the validation issue which triggered my commit to fix EMIP 397
- the issue about the wrong use of `</sic>`

Ralph-Lee-UK:
Pietro, thanks for clarifying these things. I have encountered some problems with the autocomplete package not working properly, which has, at least in my case, probably been associated with slow internet speeds. I have suggested a simple approach to use at the beginning of each editing session to check that it is functioning properly. You will also recall that I had an issue with it failing completely, and that then reinstalling the package needed some extra work. 

With the PR request issue, we are going to move to collecting work, and making PR perhaps once a week (unless there is a lot of work). As you have indicated, work can continue with the creation of a new branch, so I'll start dating the EMIP metadata branch so that it is clear which branch should be worked on after a PR has been made. 

We should review all of this when we have our training session!

PietroLiuzzo:
to work around your validation issues I will add in the review guidelines a recommendation to validate the data for the reviewer before doing anything else. this is what I am about to do now again on these files.

PR #133, opened 2019-12-20
manuscript description

thea-m:
See https://github.com/BetaMasaheft/Documentation/issues/1301 for the missing person records
There is quite a lot Geez text, I would be grateful if @sologebre could skim it for typos


thea-m:
accidentally commited with the ES manuscript, where I corrected the Gadla Hawaryat recension in a title

PR #137, opened 2020-01-03
IVefiopsk 19 RNBefns19

DenisNosnitsin1970:
Renaming, minor changes


eu-genia:
I personally would refer to the bibliographic entry and not to the person in descriptions such as 
"```
<note>According to <persName ref="PRS9602Turaev"/>,
                        partly coloured in the recent time.</note>
```"

DenisNosnitsin1970:
I will just take Turaev out. It is clear that the only source is his description.

PR #140, opened 2020-01-08
Update InEtn384127a.xml

DariaElagina:
added information from Platonov's catalog, corrected the locus (Turaev does not provide any)


DenisNosnitsin1970:
ok

PR #184, opened 2020-02-04
Bruce 82

thea-m:


DenisNosnitsin1970:
"The edges of the manuscript have been cut off before binding and gilded" - do you mean the that the textblock was trimmed, and its edges were gilded?
"One paper fly leaf each, containing a watermark with a pattern and Latin characters at the front and a watermark with Latin characters at the back". - Not very clear, what is "pattern"?

thea-m:
"do you mean the that the textblock was trimmed, and its edges were gilded?" Yes, much better, thank you! I will correct the other records as well.
Then maybe it would be better to just state the presence of the watermark without attempts of description?

DenisNosnitsin1970:
Just state the presence of a watermark, that will be fine.

thea-m:
It this ok now?

thea-m:
I think that the structure of the last quire was unclear to me, and the reason from the stub not obvious.

PR #187, opened 2020-02-04
Bruce 85

thea-m:
limited description due to lack of foliation


DenisNosnitsin1970:
Of course, I meant ""imprinted on the spine" - "printed on the spine", on there is a paper label?

PR #216, opened 2020-02-18
Neukirchen 004

DariaElagina:


DenisNosnitsin1970:
Did you propose a dating? I would always indicate if the leather cover is "blind-tooled" (most of the cases) or "untooled" (rare, all exemplars are valuable). Otherwise there is an ambiguity.

DariaElagina:
blind-tooled added, I propose no dating, I am not trained enough in this aspect.

PR #220, opened 2020-02-19
Emip metatdata 13 feb 2020

Ralph-Lee-UK:
Weekly pull request with final edits to some of the Tweed Collection.


PietroLiuzzo:
as I am busy in a workshop, it may take me some time to get to this.

Ralph-Lee-UK:
Pietro, perhaps then I should cancel this pull request, and make another may early next week when you are more available? At the moment we have several instances of people working on the same files, so i think that this would be the best way to avoid conflicts?

PietroLiuzzo:
no, please, just be patient. I will try to get it done at a break

Ralph-Lee-UK:
OK, I'll wait!

PR #233, opened 2020-02-24
DS Ethiop1

thea-m:
Improved the encoding, added available incipits and explicits. 
Still missing:
- description of the hand (do you want to do it @DenisNosnitsin1970 ?)
- I have no more images available which would allow the transcription of lenghtier incipits and explicits. Do you have any by chance?
- the issues mentioned in comments in the file (particularly regarding the date)


thea-m:
Is the `<origDate>` maybe a copy paste error from DSEthiop6?

PR #241, opened 2020-02-27
new record

thea-m:


thea-m:
(also here, sorry for the later corrections! will stop committing to this branch now :) )

LondonBritishLibrary/orient/BLorient682.xml
```diff
@@ -0,0 +1,491 @@
...
+                                <title type="complete">Acts and Martyrdom of St Paul</title>
+                                <textLang mainLang="gez"/>
+                                
+                                <msItem xml:id="ms_i1.3.1">
+                                    <locus from="31r"/>
+                                    <title type="complete" ref="LIT1016Actsof"/>
+                                    <textLang mainLang="gez"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.3.19">
```
DariaElagina: 19?

LondonBritishLibrary/orient/BLorient682.xml
```diff
@@ -0,0 +1,491 @@
...
+                                    </q>
+                                </item>
+                            </list>
+                        </additions>
+                        
+                    
+                        
+                        <bindingDesc>
+                            <binding contemporary="true" xml:id="binding">
+                                <decoNote xml:id="b1">In the original boards, lined with striped cloth and covered with stamped leather.</decoNote>
```
DariaElagina: stamped or tooled?

thea-m: yes, you are right, I just took Wright's terminology. Since there are digitized manuscripts of his catalogue I could safely confirm that stamped means blind tooled. I will correct the other records in a separate branch.

DenisNosnitsin1970:
I prefer to use the formula which uses the term "(textile) inlays", which I put in a separate "deconote b". Nearly all items from Ethio-Spare are described like this. Both ways are not wrong, just different. I wonder, when  someone will look for this technique if and how he will get all the cases if they are described in different terms.

thea-m:
Haha, I was counting the seconds until someone would say this. According to the guidelines, (in case of a more elaborate binding description), this would be included in a `<decoNote type="Boards">`. Maybe you could open a new issue if you think there should be a keyword or something similar for this feature.

PR #243, opened 2020-03-01
Emip1859

thea-m:
added text for my thesis


thea-m:
(Sorry for the later commits, it seemed the easier solution. I will not commit anymore when it is reviewed)

PR #249, opened 2020-03-03
BL orient 679

thea-m:


thea-m:
Thank you, that is a good point! do we know that no manuscripts at all were produced in Madḫāne ʿĀlam? I.e., could this line be added to all manuscripts having a Madḫāne ʿĀlam ownership note, or would it be safer to restrict it to those manuscripts in which we have internal evidence to their provenance before they came to Madḫāne ʿĀlam (like this one)?

DenisNosnitsin1970:
We can be 99.5% sure that no manuscript were produced in Medhane alam. If we have the ownership note of the specific shape in a book (you know what I mean), we can be sure that the book was brought there by Tewodros from Gondar (the full list of Medhane alam books is a desideratum, by the way). Of course, it is not the end of the story, as the book might have been brought to that church in Gondar from elsewhere as well (e.g. produced in Gojjam (in some cases still before the foundation of Gondar!), gone through several hands or churches, before landing in one of the Gondar churches).

thea-m:
Can this be merged?

PR #264, opened 2020-03-09
210325

DariaElagina:


StPetersburg/InEtn210325.xml
```diff
@@ -0,0 +1,160 @@
...
+                <msIdentifier>
+                    <repository ref="INS0524InEtn"/>
+                    <idno>2103-25</idno>
+                    <altIdentifier>
+                      <idno>Platonov cat. III.116</idno>
+                    </altIdentifier>
+                </msIdentifier>
+                <msContents>
+                  <summary>
+                    The manuscript was used to train reading or copying.
```
thea-m: Interesting information! I assume that the information is given like this in the catalogue, without explaining why the cataloguer came to this conclusion? I am wondering whether this is the best place for it, but have no other idea either. What do you think @PietroLiuzzo @DenisNosnitsin1970

DenisNosnitsin1970: I would say, "according to "ref to the note", the manuscript was meant for disciples of a nebab-school. I think the passage is exactly the one with what they start, to learn how to read letters (it can /should be checked).

DariaElagina: In this case, there is a note which says that the manuscript was produced for learning, I can add a ref. But in the case of InEtn210321 this information is just mentioned in the catalogue description.

thea-m: I am sorry, but I really don't think that `<summary>` is the appropriate place for this. It should be used for a summary of the content, if at all. At least in this case, the information could be added in the `<desc>` of a1. As for the manuscript where it is just mentioned by the catalogue, maybe an extra? Or a note in the msItem?

DenisNosnitsin1970:
Isn't it smth for the origin? If a ms was meant as donation, it is indicated under origin as the main motivation behind its creation, isn't it? Here is the same.

DariaElagina:
@thea-m  According to TEI guidlines, <summary> contains an overview of the available information concerning some aspect of an item or object (for example, its intellectual content, history, layout, typography etc.) as a complement or alternative to the more detailed information carried by more specific elements. In my opinion, it is a good place to state that the manuscript was used as a student book. But I don't insist on it.

DariaElagina:
@DenisNosnitsin1970  you are right, but on the other hand there is an aspect of usage, not only production as well

DariaElagina:
I am open to any suggestion, I have another such scroll in the catalogue, so I think it wouzld be good to encode this information in a consistent way

thea-m:
@DariaElagina that's fine, but anything (even TEI guidelines) contradicting our guidelines needs to be discussed first (https://betamasaheft.eu/Guidelines/?q=summary&id=manuscriptContents). What do you think @PietroLiuzzo ? Our place for anything hard to place elsewhere has been extras, but this is obviously not ideal either.

DenisNosnitsin1970:
Now I think that the aspect of usage remains unclear to me, as the manuscript was landed in the hands of a foreigner. If it is Kohanowski collection, there is a big possibility that the item was produced for the collector on purpose, or somehow not for the indigenous users. In other words, it is an early "manuscript for tourists", fir which the manuscript-maker went far beyond the conventions. So, we probably have to distinguish what the manuscript-maker stated in a subscription and how the item fits into the traditional context. The form of the scroll (associated exclusively with magic) is actually not fitting for the context of a church school. "Lange Rede kurzer Sinn" - I would not state that the ms was used in school (we did not see it, we cannot estimate if it is used or not, by the way), I would only reproduce the statement of the scribe.

DariaElagina:
What should I do with two other such scrolls, which are described as "student books" by the cataloguer, but have no scribal notes in them? Three scrolls with the same content, without decoration are not something spectacular to be produced for "touirists". On the other hand, we can not stay that the scrolls were not typical for church schools at the end of 19th -early 20th century. Now, we have three examples (or even four examples) of scrolls which might have been used in schools. By the way they are cheaper and more handy, which makes sence.

DenisNosnitsin1970:
I wouldn't say that the opinion of the cataloguer is wright or wrong, I would just underline that this is his opinion (for which he does not propose arguments, in my opinion, there are many other scenarios as to why these items were produced). The scrolls as teaching devices are indeed not typical, at least to my knowledge, and generally the idea of a special book production for the needs of the nebab schools (the lowest level) was not common. We cannot/should not overturn his opinion, but we should express reservations at least, or somehow leave it open. Someone might look at the items again to see if they are written in the same hand, or if they have traces of intensive use etc.

DariaElagina:
Ok, I agree that this is just the opinion of a cataloguer, which however should be present in the description. We just have to decide where to put this information. In extras?

DenisNosnitsin1970:
In my opinion, "Origin" is more suitable than extras (as this information points to the reasons of the creation of the manuscripts).

PR #266, opened 2020-03-09
210327

DariaElagina:


DariaElagina:
So, the drawing is neither Net of Solomon, nor the Seal of Solomon. I would prefer to leave it as it is by now.

DenisNosnitsin1970:
You could at least "dissimilate" it from AT1140SealSol, not to leave it without any context (but I still believe that it is a very liberally or rather unprofessionally realized talsam-motives). By the way, 
if its origDate notBefore="1850" notAfter="1900", there is a gap of 5 years till the moment it was bought by Koh in 1905-1913.

DariaElagina:
I think that this is a talsam-motive as well, though I can't be sure about that. But  AT1140SealSol is a very specific and good identifiable talsam-type. I don't think I schould provide its id for the description of a drawing which is quite different. This would be even more misleading.
The origDate is based on the catalogue description, but I see ypur point, I will change: notAfter="1913"

thea-m:
I don't know these drawings well or the manuscript at all, but as I understand @DenisNosnitsin1970 , the ideal is to mention the AT1140SealSol ID in order to state that the case of this manuscript is quite different - readers would otherwise probably have the same question.

DariaElagina:
Well, from my experience many/most such drawings are different to AT1140SealSol. Ideally, we should have a key word for talsam, and a repertoire of them. But the problem is that we can't always identify talsams basing just on the catalogue descrition.

DariaElagina:
If we follow this rule, we would need to add this information to practically every drawing in magic scrolls different to AT1140SealSol

DenisNosnitsin1970:
I am not certain as to what is meant with "many/most such drawings". For me, here is rather an individual case of of a talismanic image that remotely reminds the common motif AT1140SealSol and seems to be in place of it. As a cataloguer, you can feel free to comment on this (it is, in fact, a minor thing) without compelling yourself to create a comprehensive rule (in the last case, I would go rather after concrete examples of the same treatment of talsam-motives, not after the abstract principle "A dissimilar B").

DariaElagina:
I think, that it might be a case of individual perception. The seal of Solomon is an eight-pointed star with a face inside. Sometimes the points af the star have eyes on them. In the case of this drawing there are two faces in a square, no cross.  Optically there is not so much similarity for me.

thea-m:
If you feel that a keyword for talsam would be helpful, you can always create one :)

DenisNosnitsin1970:
Well, this is exactly the case to comment upon and to remark this, as the Seal would be more expected in this place, isn't is?

DariaElagina:
@thea-m No, I think that it would be very misleading, because the real definition of talsam is not clear. One can't acsribe this definition randomlly to any drawing in a magic scroll. I think that if someone is interested in talsams or sth like that, the easiest way is to look for any drawings in scrolls containing magic texts. 
@DenisNosnitsin1970 I would not say so, speaking from my experience. There are many different motives which can occur on this place.

thea-m:
I have no opinion on this.

DariaElagina:
Can this be merged, @DenisNosnitsin1970 ?

PR #271, opened 2020-03-11
BLadd18994

thea-m:


thea-m:
Excellent idea! Would you suggest an ID for this particular version (including the New Testament readings, or a general "Psalter that does not follow the common arrangement" ID? Since I do not yet know the tradition well, I would tend towards the latter, which would ease the encoding.

DenisNosnitsin1970:
As a practical temporary solution, we can have a summary ID for peculiar Psalter versions (they all seem to be linked to Horologium, in one way or another). But two versions can get IDs now, or as soon as we have encountered witnesses: the Psalter without Anqasa berhan at the end, and Psalter with other texts distributed between the psalms (like Mazmura Krestos in DD-050).

thea-m:
Very good! So for "Psalter without Anqasa berhan" you mean with Weddase Maryam, but without Anqasa berhan? Just to be sure.

DenisNosnitsin1970:
Yes, just the last one, Anqasa berhan, is missing.

thea-m:
@DenisNosnitsin1970 is it ok like this? It's just one non-traditional Psalter ID now, we can create others in the future

thea-m:
Sure, it is in the manuscript description in the record. Thank you!

PR #279, opened 2020-03-16
Transliteration

DariaElagina:


thea-m:
I am sure that the change of format is related to your find/replace work in atom - is there any possibility of doing the find/replace without changing the format? Like this, it just takes very long for me to skim the records, because everything is coloured, and there are these header conflicts. But if this is the only way to do it, don't worry

DariaElagina:
Well, the working procedure was the same as before, those conflicts appeared just for this two commits only. I don't know what has been different

thea-m:
(Sorry, I think the format thing is related to changes Pietro has carried out in all records for the latest release!)

PietroLiuzzo:
The data freeze was not lifted yet. Please refrain from any further action until I give a green light. Do not attempt to revert this or other merges, this will just push forward the end of data freeze until I will have managed to check this as well.

DariaElagina:
@PietroLiuzzo sorry, I have completely forgotten about this.

PR #280, opened 2020-03-16
new record

thea-m:
Wright always states that after the Wəddāse Māryām, there is "another Wəddāse Māryām" - I strongly suppose that this is ʾAnqaṣa bərhān, but found none of the concerned manuscripts digitised so could not confirm.

### DIFF
--- a/LondonBritishLibrary/add/BLadd19960.xml
+++ b/LondonBritishLibrary/add/BLadd19960.xml
@@ -1,0 +1,220 @@
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"
+?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng" 
+   schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI xmlns="http://www.tei-c.org/ns/1.0" type="mss" xml:id="BLadd19960" xml:lang="en">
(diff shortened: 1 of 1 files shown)


DenisNosnitsin1970:
I think Wright says so because Anqasa berhan actually starts with words ()Nesehef...) Weddaseha la-Maryam. Anqasa berhan appears as a methaphor for St. Mary later in the text, and was picked as "supplied title".

thea-m:
Yes, that makes sense

thea-m:
@DenisNosnitsin1970  @DariaElagina

LondonBritishLibrary/add/BLadd19960.xml
```diff
@@ -0,0 +1,220 @@
...
+                        </objectDesc>
+
+                        <handDesc>
+                            <handNote script="Ethiopic" xml:id="h1">
+                                <date notBefore="1600" notAfter="1699"/>
+                                <desc>Written in a neat hand.</desc>
+                            </handNote>
+                            <handNote script="Ethiopic" xml:id="h2" corresp="#a2">
+                                <date notBefore="1699" notAfter="1877"/>
+                                <desc>Written in a later hand.</desc>
```
eu-genia: and the hand of a1?

thea-m: I believe that many years ago we decided not to add hands for additions if we have no information on them, because it was assumed that an addition is by definition in another than the main hand. My principle was to only add a hand for an addition if the catalogue gives information on it (such as here h2).

thea-m: I will merge this now, since it is already approved, and the hand issue is larger than this singe PR. If you would like to start a discussion or think that I misremember something, please open an issue. THank you!

PR #283, opened 2020-03-18
BLorient542 and BLorient78

thea-m:


thea-m:
hmm, I am happy to do it, my last memory from this discussion was that it was not so useful for manuscript records which have other datings. I'll add them now, it can be useful for filtering etc.

DenisNosnitsin1970:
Yes, but, in a way, there is no manuscript with no "other dating" (each manuscript can be dated palaeographically), maybe only if the manuscript can be more or less precisely dated or precisely datable. But also in this case it may be useful for filtering since the periods are supposed to be historic.

thea-m:
Can this be merged? @DenisNosnitsin1970 @DariaElagina

PR #286, opened 2020-03-19
added content, additions, binding

thea-m:
I am unsure about assigning the ID LIT1850Malkea, since this manuscript does not follow the typical sequence of hymns of the Malkǝʾa gubāʾe. On the other hand, this is the title given in the ownership note, and we follow the indigenous tradition. What do you think?

### DIFF
--- a/LondonBritishLibrary/orient/BLorient579.xml
+++ b/LondonBritishLibrary/orient/BLorient579.xml
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
     <teiHeader xml:base="https://betamasaheft.eu/">
         <fileDesc>
             <titleStmt>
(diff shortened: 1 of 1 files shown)


thea-m:
@DenisNosnitsin1970 @DariaElagina

eu-genia:
Could it make sense to have a generic work ID for non-standardized collections of malk that one could refer to here?
This would also be important for our discussion of traditional "labels" vs what we understand under titles of works (then under the MG label you would find everything the tradition may refer to as such, even if it is not THE MG)

thea-m:
I would prefer to keep this open for the moment, instead of maybe too quickly creating the illusion of a standard 
ttps://github.com/BetaMasaheft/Documentation/issues/1381

thea-m:
I'm quite sure that whoever works on https://github.com/BetaMasaheft/Documentation/issues/1381 (which does not seem to happen anytime soon), will find this record then and update it. Can we merge it?

PR #287, opened 2020-03-19
BLadd19342

thea-m:


thea-m:
Yes, I don't know about it's origin...I wonder whether the "pontificale" means that the church in question was one regularly visited by the pope - or do all churches have such pontificial clothes?...

DenisNosnitsin1970:
No one knows, one has to see the manuscript.

thea-m:
you can :) http://www.bl.uk/manuscripts/Viewer.aspx?ref=add_ms_19342_fs001r 
but it did not help me much

DenisNosnitsin1970:
Mysterious object! It should be left for future generations. :)

PR #294, opened 2020-03-24
40555

DariaElagina:


DariaElagina:
Could you please mark the place in the code to which you refer?

StPetersburg/InEtn40555.xml
```diff
@@ -0,0 +1,151 @@
...
+                      </supportDesc>
+                      <layoutDesc>
+                          <layout columns="1" writtenLines="18">
+                          </layout>
+                      </layoutDesc>
+                  </objectDesc>
+                  <handDesc>
+                    <handNote script="Ethiopic" xml:id="h1">
+                        <seg type="ink">Black, red</seg>
+                        <date>Second half of the 19th-cent - beginning of the 20th century according to <bibl><ptr target="bm:Chernetsov1975Svitki"/>
```
thea-m: It would be good to express the dating in `<date>`, I think

StPetersburg/InEtn40555.xml
```diff
@@ -0,0 +1,151 @@
...
+                        <desc>Ornamental band with three <term key="cross">crosses</term>, published in <bibl><ptr target="bm:Chernetsov1975Svitki"/>
+                        <citedRange unit="page">221</citedRange></bibl>.</desc>
+                    </decoNote>
+
+                  </decoDesc>
+
+                </physDesc>
+                <history>
+                  <origin>
+                    <origDate notBefore="1850" notAfter="1909"></origDate>
```
thea-m: you could give evidence type="lettering" (it this is true)

StPetersburg/InEtn40555.xml
```diff
@@ -0,0 +1,151 @@
...
+                  </origin>
+                  <acquisition>The manuscript was acquired by <persName ref="PRS9602Turaev">Boris Turaev</persName> in <date>1909</date> in
+                  <placeName ref="LOC2804Egypt"></placeName>
+                  and was kept in <ref type="place" corresp="INS0526Hermitage"/>.
+                  In <date>1930</date> it was granted to <ref type="place" corresp="INS0524InEtn"/>.</acquisition>
```
thea-m: donated?

StPetersburg/InEtn40555.xml
DariaElagina: done

StPetersburg/InEtn40555.xml
DariaElagina: done

StPetersburg/InEtn40555.xml
DariaElagina: done

thea-m:
From my point of view, this can be merged, I also could not follow @DenisNosnitsin1970 's comment :)

DenisNosnitsin1970:
In InEtn40555.xml, the formula "The text is written on the reverse side" is repeated in line 121, 127, 133, 139, 150 etc. I mean, that it will not be easily understood (and writing on the reverse side is, again, not typical), a couple of words of explanation would not be redundant.

thea-m:
Do you mean in InEtn40553?

DariaElagina:
Apparentely in 40553, the writing on the reverse side is of course not typical. But what do you mean by explanation?

DenisNosnitsin1970:
A user will not quickly understand that the reverse side usually has no writing (there are cultures with scroll where both sides are used for writing). The fact that the reverse side has normally no text in Ethiopian scrolls is not obvious for everyone. I ask you to help the future user and add a short explanatory note, e.g. under "Layout" (and this is a feature of layout.)

DariaElagina:
Well, the problem is that those texts are clearly additions (I have seen this scroll) and are encoded like additions. I do not think that we describe in the layout-element the layout of additions.

DenisNosnitsin1970:
The fact that this scroll has additions is even more interesting and worth of highlighting. :)

DariaElagina:
So, I will prepare a contribution on this scroll :) I have added a note, in another branch

DenisNosnitsin1970:
Ok.

PR #302, opened 2020-03-26
continued description

thea-m:
I am not entirely sure whether ʾAmməstu ʾaʿmāda məśṭir is really an addition here, or should be a msItem. It is not clear to me from the catalogue.

### DIFF
--- a/LondonBritishLibrary/add/BLadd24992.xml
+++ b/LondonBritishLibrary/add/BLadd24992.xml
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
     <teiHeader xml:base="https://betamasaheft.eu/">
         <fileDesc>
             <titleStmt>
-                <title>Dāwit, Mazmura Dǝngǝl</title>
+                <title>Dāwit with Mazmura dǝngǝl, ʾAmmǝstu ʾaʿmāda mǝśṭir, prayers</title>
                 <editor key="DR"/>
                 <editor role="generalEditor" key="AB"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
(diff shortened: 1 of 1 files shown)


thea-m:
sorry, I'm done committing to this now

DenisNosnitsin1970:
I think it is an addition, it is written on the margins of the folios.

LondonBritishLibrary/add/BLadd24992.xml
```diff
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
     <teiHeader xml:base="https://betamasaheft.eu/">
         <fileDesc>
             <titleStmt>
-                <title>Dāwit, Mazmura Dǝngǝl</title>
+                <title>Dāwit with Mazmura dǝngǝl, ʾAmmǝstu ʾaʿmāda mǝśṭir, prayers</title>
```
DariaElagina: I think you can just write Mazmura dǝngǝl, because Dawit is included into MD

LondonBritishLibrary/add/BLadd24992.xml
```diff
@@ -39,57 +39,64 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
                     <msContents>
                         <summary/>
-
+                        
                         <msItem xml:id="ms_i1">
+                            <locus from="2r"/>
+                            <title type="complete" ref="LIT2701Dawit"/>
+
+                        <msItem xml:id="ms_i1.1">
                             <locus from="2r"/>
```
DariaElagina: I am not sure that you should use LIT2701Dawit here, it is Mazmura Degel plus Weddase and Anqasa. Maybe we could create a LIT for MD accompanied by these texts, which is not so uncommon.

thea-m: I think you worked most on these texts, so please decide that. My thinking was that someone looking for Dāwits would want to find also those which include Mazmura dəngəl, but of course the same could be achieved with a new ID and the appropriate recensions.

DariaElagina: I see, let's leave it like that by now; I would need time to come with a further solution on MD.

thea-m: I meant relations instead of recensions in my last comment...

eu-genia:
@DariaElagina @DenisNosnitsin1970 can this be merged?

PR #304, opened 2020-03-26
Create InEtn40556.xml

DariaElagina:


StPetersburg/InEtn40556.xml
```diff
@@ -0,0 +1,223 @@
...
+                          <layout columns="1" writtenLines="300">
+                          </layout>
+                      </layoutDesc>
+                  </objectDesc>
+                  <handDesc>
+                    <handNote script="Ethiopic" xml:id="h1">
+                        <seg type="ink">Black, red</seg>
+                        <seg type="script">The script is archaic, angular according to <bibl><ptr target="bm:Chernetsov1975Svitki"/>
+                        <citedRange unit="page">221</citedRange></bibl>.</seg>
+                        <date><date>Second half of the 18th-cent - beginning of the 19th century</date> according to <bibl><ptr target="bm:Chernetsov1975Svitki"/>
```
thea-m: There is a double  `<date><date>` :) And it could be useful to encode the date within `<date>`

StPetersburg/InEtn40556.xml
```diff
@@ -0,0 +1,223 @@
...
+                      <ref type="authFile" corresp="AT1140SealSol"></ref> and has been published in <bibl><ptr target="bm:Chernetsov1975Svitki"/>
+                      <citedRange unit="page">221</citedRange></bibl>.</desc>
+                    </decoNote>
+
+                  </decoDesc>
+                  <additions>
+                    <list>
+                      <item xml:id="e1">
+                        <desc>
+                          A paper shit with a description of scroll by an unknown cataloguer has been glued to the reverse
```
thea-m: sheet :)
description of the scroll?

StPetersburg/InEtn40556.xml
```diff
@@ -0,0 +1,223 @@
...
+                          side of the scroll. This description reads as follows:
+                          <foreign xml:id="de">Aethiopisches Manuskript aus dem 16. Jahrh. - 300 Zeilen in
+                          aethiopischener Schrift sauber in schwarz und rot quer über einen 208 cm.
```
thea-m: is aethiopischener in the catalogue?

thea-m:
oh sorry, everyone started working at the same time...

StPetersburg/InEtn40556.xml
DariaElagina: I will correct that as soon as I can fetch the origin :)

StPetersburg/InEtn40556.xml
DariaElagina: I think, this is what I did, there are two date-elements there. Or I do not get the point

thea-m: What I see here is:
` <date><date>Second half of the 18th-cent - beginning of the 19th century</date> according to <bibl><ptr target="bm:Chernetsov1975Svitki"/><citedRange unit="page">221</citedRange></bibl>.</date>`
There are two date elements, which is a bit redundant, but the date itself is not marked up precisely:
`<date notBefore="1750" notAfter="1849">` or whichever dates seem best to you.

PR #305, opened 2020-03-26
Create InEtn40557.xml

DariaElagina:


DariaElagina:
The cataloguer makes difference between parchment and leather in the description of the scrolls. The usage of leather, even of some exotic and wild animals, is well documented by the contributions on magic scrolls and Ethiopian magic in general.

DenisNosnitsin1970:
It sounds very strange; the use is rather reported (and then repeated) by not properly documented. I assume there is a kind of misunderstanding or smth like this, but go ahead.

PR #310, opened 2020-03-30
Changes

DariaElagina:
The manuscript 40559 has a very interesting layout, which I tried to describe as precise as possible. But I suppose that the enhancements will be needed.

### DIFF
--- a/StPetersburg/InEtn40558.xml
+++ b/StPetersburg/InEtn40558.xml
@@ -1,0 +1,251 @@
+<?xml-model href="https://raw.githubusercontent.com/SChAth/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/SChAth/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="InEtn40558" xml:lang="en" type="mss">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>Collection of Magic Texts</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="DE"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens: Eine multimediale
+                            Forschungsumgebung / beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+              <msDesc type="scroll">
+                <msIdentifier>
+                    <repository ref="INS0524InEtn"/>
+                    <idno>4055-8</idno>
+                    <altIdentifier>
+                      <idno>Platonov cat. III.126</idno>
+                    </altIdentifier>
+                </msIdentifier>
+                <msContents>
+                  <summary>
+                  </summary>
+                  <msItem xml:id="ms_i1">
+                    <locus from="1r" n="1"></locus>
+                    <locus to="1r" n="34"></locus>
+                    <title type="complete" ref="LIT4599Salot"><certainty locus="value" match="../@ref" cert="medium"></certainty></title>
+                    <textLang mainLang="gez"></textLang>
+                    <incipit xml:lang="gez">
+                      በስመ፡ አብ፡ ወወልድ፡ ወመንፈስ፡ ቅዱስ፡ ፩ አምላክ፡ ጸሎት፡ በእንተ፡ ማዕሠሮሙ፡
+                    ለአጋንንት፡ ባርያ፡ ወሌጌዎን፡ ዘይሰልብ፡ ልበ፡ ሰብእ፡</incipit>
+                  </msItem>
+                  <msItem xml:id="ms_i2">
+                    <locus from="1r" n="35"></locus>
+                    <locus to="1r" n="69"></locus>
+                    <title type="incomplete" ref="LIT2709Matthew"></title>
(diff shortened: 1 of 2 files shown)


StPetersburg/InEtn40558.xml
```diff
@@ -0,0 +1,251 @@
...
+                <publisher>Die Schriftkultur des christlichen Äthiopiens: Eine multimediale
+                            Forschungsumgebung / beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+              <msDesc type="scroll">
```
thea-m: I don't think that this  encoding is mentioned in our guidelines, and there is already  `<objectDesc form="Scroll">`, so I see no need for it. But I see now that you have used this in all records. Maybe I missed a discussion, sorry. Was there a particular motivation behind having the information here?

StPetersburg/InEtn40559.xml
```diff
@@ -0,0 +1,224 @@
...
+                  ወጥቁሮ፡</incipit>
+                  </msItem>
+                  <msItem xml:id="ms_i2">
+                    <locus from="1ra" n="66"></locus>
+                    <locus to="1ra" n="76"></locus>
+                    <title type="incomplete" ref="LIT4156Prayer"></title>
+                    <textLang mainLang="gez"></textLang>
+                    <incipit xml:lang="gez">አሌፍ፡ ብሂል፡ ፩ እግዚአብሔር፡ አብ፡
+                    </incipit>
+                    <note>Begging of the prayer containig explanations of six letters: ʼAlef,
```
thea-m: Beginning, containing

StPetersburg/InEtn40559.xml
```diff
@@ -0,0 +1,224 @@
...
+                              <note>A parchment strip of 14,5 cm is attached to the top of the scroll.</note>
+                          </support>
+                          <extent>
+                              <measure unit="leaf">2</measure>
+                              <dimensions type="outer" unit="mm">
+                                <height>1445</height>
+                                <width>75</width>
+                              </dimensions>
+                          </extent>
+                          <condition key="deficient">The parchement was torn and repaired with a piece of white thread.</condition>
```
thea-m: parchment
and the material you gave was leather? Or does this refer only to the attached parchment strip? I remember there was a discussion on this in another PR

StPetersburg/InEtn40559.xml
```diff
@@ -0,0 +1,224 @@
...
+                      </supportDesc>
+                      <layoutDesc>
+                        <layout>First 14 lines are written in 1 column, further the text is layouted in 2 columns
```
thea-m: I think the use of different `<layout>` elements for each unit of layout, as you have done, is exactly right. I would further try to separate the information into each layout element, with reference to their place in the scroll and the strip on which they are contained. I am aware that the guidelines on all this are suboptimal and it all needs reworking. If you want to close this PR before the scroll issue is resolved, I think separating the information into each <layout> would be good, and could then be later corrected once we have better guidelines (this is of course true for all existing scroll descriptions, not only these!)

StPetersburg/InEtn40559.xml
```diff
@@ -0,0 +1,224 @@
...
+                      </layoutDesc>
+                  </objectDesc>
+                  <handDesc>
+                    <handNote script="Ethiopic" xml:id="h1">
+                        <seg type="ink" corresp="#ms_i1 #ms_i2">Faded black, red.</seg>
+                        <seg type="ink" corresp="#ms_i3 #ms_i4 #ms_i5 #ms_i6">Black, red.</seg>
+                        <seg type="script">The script is regular according to
+                          <bibl><ptr target="bm:Chernetsov1975Svitki"/>
+                        <citedRange unit="page">223</citedRange></bibl>.</seg>
+                        <date>Mid 19th century.</date>
```
thea-m: See my remark in a previous PR - i really think that adding the attributes `<date notBefore....notAfter>` would improve on this, and if you want the paragraphic date to be there, because you prefer it, it can stay exactly where it is.

StPetersburg/InEtn40559.xml
```diff
@@ -0,0 +1,224 @@
...
+                    </handNote>
+                  </handDesc>
+                  <decoDesc>
+                    <decoNote xml:id="d1" type="ornamentation">
+                      <desc>Black ornamental band.</desc>
+                    </decoNote>
+                    <decoNote xml:id="d2" type="drawing">
+                      <desc>
+                        Drawing of an <term key="angel">angel</term> with a <term key="sword">sword</term> attached
```
thea-m: It reads a bit funny, as if the sword were attached with faded ink :) Maybe "faded-ink drawing ...."  ....?

StPetersburg/InEtn40559.xml
```diff
@@ -0,0 +1,224 @@
...
+                    </decoNote>
+                    <decoNote xml:id="d3" type="drawing">
+                      <desc>
+                        Drawing of a flying <term key="angel">angel</term> with a <term key="sword">sword</term> with faded
+                        ink.
+                      </desc>
+                    </decoNote>
+                    <decoNote xml:id="d4" type="drawing">
+                      <desc>
+                        Drawing of a devil's face with eight horns with two rods on sides in black ink.
```
thea-m: Cool :) Mark up the devil so that everyone can find him?

thea-m:
See https://github.com/BetaMasaheft/guidelines/issues/87, where the layout information could maybe better take place than in a PR.

StPetersburg/InEtn40558.xml
DariaElagina: Actually I can not rememeber either, how this appeared in my encoding. I suppose that I have made a mistake. This should rather be: xml:id="ms", right?

thea-m: Yes, at least this is what we have everywhere. I never thought about it :)

DariaElagina: I have corrected this in this PR, I will go through my other records and correct it as well

StPetersburg/InEtn40559.xml
DariaElagina: in this case the material is parchment, another scrolls are made of leather (according to the catalogue)

StPetersburg/InEtn40559.xml
DariaElagina: done

DariaElagina:
corrected

PR #312, opened 2020-03-31
Adding refernce to the asmat prayer

DariaElagina:
Just added reference to the work file I have created


thea-m:
All seems wonderful to me, but for the future, it might make sense to assign those who worked on these manuscripts.

PR #317, opened 2020-04-02
Corrections of scroll descriptions

DariaElagina:
correction of msDesc element


StPetersburg/InEtn210325.xml
```diff
@@ -133,11 +133,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                     <term key="Translation"/>
                 </keywords>
             </textClass>
-            <langUsage><language ident="en">English</language><language ident="am">Amharic</language></langUsage>
```
thea-m, line 136: Amharic is used in the addition and should also be kept here, in addition to Geez

DariaElagina: corrected

PR #318, opened 2020-04-02
More e s mss 

eu-genia:


eu-genia:
where exactly? everywhere? the text is all from DomLib

eu-genia:
@DenisNosnitsin1970 - when you review a file you can click on the plus sign to the left of the reviewed line before commenting and then your comment will be attached to the line where the correction must be carried out

I corrected the same በለስ is also in the Miracle incipit provided in ESap046

PR #321, opened 2020-04-03
Update IVorlov35.xml

DenisNosnitsin1970:
Minor edits


PietroLiuzzo:
Did you ask my review with a particular concern in mind?

DenisNosnitsin1970:
No particular concerns (though encoding mistakes cannot be exluded). Should I ask you for review only if I have some specific worries?

PietroLiuzzo:
no, that is ok, I do not get PR typically, as I am in most cases not able to review. but if you think your edits are reviewable by me, that is ok, it just made me wonder

PR #330, opened 2020-04-06
Update IVEf47.xml

DenisNosnitsin1970:
Minor changes


StPetersburg/IVEf47.xml
```diff
@@ -73,24 +78,21 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                             <item xml:id="a1">
                               <locus target="#1v"/>
                               <desc type="GuestText">
-                                <title type="incomplete" ref="LIT4551Salam" cert="medium"/> 
+                                <title type="incomplete" ref="LIT4551Salam" cert="medium"/>
                            </desc>
-                              <q xml:lang="gez">
-                              <note>Following is the incipit</note> ሰላም፡ ለከ፡ ጻድቅ፡ ላሊበላ፡</q>
+                              <q xml:lang="gez">ሰላም፡ ለከ፡ ጻድቅ፡ ላሊበላ፡</q>
```
DariaElagina, line 83: I think we should leave the note that only the incipit is quoted. What do pu think?

DenisNosnitsin1970: It should be not there. But I leave the ms for you, Ok? As I understood it is not mine, I stopped.

PR #334, opened 2020-04-08
DSEthiop3

thea-m:
- I have not added the description of the script
- What would be your preferred English translation for malkəʾ?
- I cannot discern a ruling pattern for msPart p2. 
- I am not completely sure whether the measurements given on the metadata sheet for text area 2 really refer to text area 2 (for the second msPart) or to its folio size.

### DIFF
--- a/DayrSuryan/DSEthiop3.xml
+++ b/DayrSuryan/DSEthiop3.xml
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
-            <title>Organ of Mary, Malkǝʾ-hymn to Gabra Manfas Qǝddus, Malkǝʾ-hymn to Mary</title>
+            <title>Organ of Mary, Malkǝʾ-hymn to Gabra Manfas Qǝddus, Malkǝʾ-hymn to Sunday</title>
             <editor key="DR"/>
             <editor key="DN"/>
             <editor key="AB" role="generalEditor"/>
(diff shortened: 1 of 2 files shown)


DayrSuryan/DSEthiop3.xml
```diff
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
-            <title>ʾArgānona Māryām</title>
```
DenisNosnitsin1970, line 8: Don't we want to change to "Harp of Mary"?

DayrSuryan/DSEthiop3.xml
```diff
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+            <title>Organ of Mary, Malkǝʾ-hymn to Mary</title>
```
DenisNosnitsin1970: The simplest for Malke' is "Image of...", straight forward.

DenisNosnitsin1970: Otherwise just Malke'-hymn.

DayrSuryan/DSEthiop3.xml
thea-m: I would be in favour of "Organ of...", which is the literatl translation and title given in the EAE. I realize that the matter of musical instruments is complicated...

DayrSuryan/DSEthiop3.xml
thea-m: I used Malke'-hymn, because it is a specific genre, which this makes more obvious

PR #335, opened 2020-04-08
Update DSEthiop21.xml

DenisNosnitsin1970:
Revisions


thea-m:
I have edited small things directly in the file. 
(We have to decide how to spell text()block...Comst has "text block")
Are you planning to add information on the binding? 
I would suggest that, when you edit the records, you add change elements with your name, otherwise you will not be acknowledged appropriately in the files.

DenisNosnitsin1970:
Let us write "text block", it is more common. I do not want to add anything to the binding of 21.

thea-m:
I think it could be useful to add something very generic as "modern hard-cover binding", to give readers an impression (especially in the case there might be no image of this). Now they will only know that the textblock is produced industrially.

PR #339, opened 2020-04-14
46931

DariaElagina:


thea-m:
As I wrote, in this month I have rather no time for non-urgent PRs which do not depend on any particular qualifications only I might have (haha). Maybe you could consider assigning Zhenia if you want it to be dealt with? I will of course work through the PRs in 3-4 weeks, but you might not want to wait so long.

PR #341, opened 2020-04-14
DSEthiop4

thea-m:
- The binding is described in the record as "reddish brown" (I don't remember by whom originally) - to me, it looks rather light-medium brown on the photos.
- in the record it was stated that there is the note in pencil "شلاخرث" (not an Arabic word) on f. 172v. I cannot see this on the image, can you?
- regarding Gabra Səllāse: he is mentioned with Dabra Bizan and Dayr al-Baramus. I have a very slight doubt: could we identitfy them with the two famous monasteries? Or might "Dabra Bizan" (Byzantium) be a translation of Dayr al-Baramus (Monastery of the Romans)?
- regarding Hayla Gabreel: The Arabic metadata says that "it is said about him that he worked magic and was expulsed from his monastery". I can see nothing regarding this in the manuscript.

### DIFF
--- a/DayrSuryan/DSEthiop4.xml
+++ b/DayrSuryan/DSEthiop4.xml
@@ -352,6 +352,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         </collation>
                         
                         <condition key="good">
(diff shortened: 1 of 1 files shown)


thea-m:
@DenisNosnitsin1970

PR #343, opened 2020-04-15
DSEthiop5

thea-m:
- The binding description had "leather inner boards" - I'm not sure what this means (maybe I wrote it once), can we delete it? The boards look wooden to me.
- I included a `<layout>` where I recorded the ruling pattern of two leaves of the protective quire ruled differently from the rest of the manuscript, but I didn't record that other leaves of the protective quire are not ruled at all. Should this also be a `<layout>`, what do you think?
- I didn't describe the script for now.
- I summarized several of the varia `<item>`s and commented the single items out.

### DIFF
--- a/DayrSuryan/DSEthiop5.xml
+++ b/DayrSuryan/DSEthiop5.xml
@@ -26,10 +26,16 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                      <collection>Ethiopic</collection>
                       <idno>DS Ethiop. 5</idno>
                      <altIdentifier>
-                        <idno xml:lang="ar">مسلسل ٧١٥</idno> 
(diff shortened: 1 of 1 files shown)


DenisNosnitsin1970:
The binding looks normal, two wooden boards covered with.... Inlays etc. I think we should not attend specifically the ruling/pricking of the protective quires, at least now. Through their status of endleaf quires (clerly marked A and B in the quire formula), we say that they may have deviations as the were not meant for regular text. I think this is enough.

DenisNosnitsin1970:
In other words, we should not create a special layout zone for A and B quires, at least now. It is enough if we, under "layout", just limit the indication for the text quires regularly ruled and pricked. If there is smth really outstanding with A and B, a remark under "other details" will be enough.

PR #347, opened 2020-04-16
Ds ethiop6

thea-m:


DayrSuryan/DSEthiop6.xml
```diff
@@ -80,14 +88,15 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
-                           <measure unit="leaf">120</measure>
+                           <measure unit="leaf">I + 120</measure>
                            <dimensions type="outer" unit="mm">
                               <height>182</height>
                               <width>127</width>
                               <depth>25</depth>
                            </dimensions>
 
                         </extent>
+                        <foliation resp="DN DR">The manuscript is foliated in pencil in the centre of the recto.</foliation>
```
DenisNosnitsin1970, line 99: Foliation in pencil, in the center of the bottom margin. (If you say "foliation", it is clear).

DayrSuryan/DSEthiop6.xml
```diff
@@ -265,7 +276,8 @@ schematypens="http://relaxng.org/ns/structure/1.0"
 
                   <bindingDesc>
                      <binding xml:id="binding">
-                        <decoNote xml:id="b1">Non-Ethiopian binding </decoNote>
+                        <decoNote xml:id="b1">Non-Ethiopian binding. A paper on which the
+                           shelfmark <foreign xml:id="ar">٢٣ ب</foreign> is glued to the spine.</decoNote>
                           <decoNote xml:id="b2">Imprinted decorations (gilded in the center).</decoNote>
```
DenisNosnitsin1970: decoration on the boards (guilded frame and medallion).

thea-m:
I've added the corrections. I noticed that some of your comments referred to older versions of this file - not dramatic in this case, but in other records this might lead to confusion. I think the best thing to do in PRs like this is to check the "Files changed" tab and to read the file there, because all changes will be combined (and not click on the "commits" tab and then read the single commits)

PR #354, opened 2020-04-20
Ds ethio2correction

DenisNosnitsin1970:
Minor corrections


DayrSuryan/DSEthiop2.xml
```diff
@@ -252,16 +254,20 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
 
                         <handNote xml:id="h2" script="Ethiopic" corresp="p1_i2">
                              <seg type="ink">Black, red.</seg>
                            <seg type="rubrication">Names of the months.</seg>
-                           <date notBefore="1850" notAfter="1949">Late 19th or first half of the 20th century.</date>
+
+                           <date>Late 19th or first half of the 20th century.</date>
                            <desc>Mediocre, letters of bigger size,
-                           very similar or possibly identical to <ref target="#h3">Hand 3</ref>.</desc>
+                           very similar or possibly identical <ref target="#h3"/>.</desc>
```
thea-m, line 269: I suggested in the comments to the PDF that we visualize the content of `<ref @target>`. I think that will be the easiest, instead of always coming up with different rules on how to call the different entities we refer to. How would you like `<ref target="#h3"/>` to be printed in the article?

DayrSuryan/DSEthiop2.xml
```diff
@@ -359,6 +369,65 @@ schematypens="http://relaxng.org/ns/structure/1.0"
 
                            <item xml:id="e3">
                               <locus target="#ir"/>
+
```
thea-m, line 372: This and the following extra descriptions are older versions not present like this anymore in the current state of this record in the master branch. If it is just a synchronizing accident, I can fix it easily without complications. But maybe it is not an accident? Did you chose the older version on purpose?

thea-m:
Thank you! I just have some small question on potential visualisation and a little synchronizing issue.

DayrSuryan/DSEthiop2.xml
```diff
@@ -252,16 +254,20 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                            <seg type="rubrication">Incipits, chapter
                              titles, names Jesus and Christ, parts of the punctuation sign ፨,
                              days of the week written in the upper margin.</seg>
-                             <date notBefore="1850" notAfter="1949">Late 19th or first half of the 20th century.</date>
+
+                             <date>Late 19th or first half of the 20th century.</date>
```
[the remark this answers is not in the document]
DenisNosnitsin1970: I moved the bottom value. Late 18th is not the same as after mid-18th cent.

DayrSuryan/DSEthiop2.xml
DenisNosnitsin1970: "identical to Hand 3", this is ok.

DayrSuryan/DSEthiop2.xml
DenisNosnitsin1970: No, sorry, it was an accident. It was a late time.

thea-m: No problem! I'll do it now, if you aren't working in the file at the same time

PR #356, opened 2020-04-22
Update ESamq020 ESamq017

eu-genia:


thea-m:
Thank you! I am afraid Denis and I have to concentrate on the article for the next few weeks (2-3). We will of course work through any open pull requests immediately then.

PR #360, opened 2020-04-22
Ds ethiop 8

thea-m:
I think the collation should not be printed, but can be left in the record.
All the varia with editorial practices are also too detailed, but I wonder whether a general note on intensive scribal corrections etc. having taken place in the manuscript would be useful?


DenisNosnitsin1970:
We wanted to reproduce collation only for parchment codices. Yes, a remark on scribal practices can be useful. "Other details"?

PR #364, opened 2020-04-23
6607 15

DariaElagina:


StPetersburg/InEtn660715.xml
```diff
@@ -0,0 +1,268 @@
...
+                      <idno>Platonov cat. III.131</idno>
+                    </altIdentifier>
+                </msIdentifier>
+                <msContents>
+                  <summary>
+                  </summary>
+                  <msItem xml:id="ms_i1">
+                    <locus from="1r" n="1"></locus>
+                    <locus to="1r" n="13"></locus>
+                    <title type="complete" ref="LIT5889UndoingOfCharms" xml:lang="gez">መፍትሔ፡ ሥራይ፡</title>
```
eu-genia, line 38: what about LIT1824Mafteh?

StPetersburg/InEtn660715.xml
```diff
@@ -0,0 +1,268 @@
...
+                    <locus from="1r" n="14"></locus>
+                    <locus to="1r" n="38"></locus>
+                    <title type="incomplete" ref="LIT2711Mark"></title>
+                    <textLang mainLang="gez"></textLang>
+                    <note>1.23-28.</note>
+                  </msItem>
+                  <msItem xml:id="ms_i3">
+                    <locus from="1r" n="39"></locus>
+                    <locus to="1r" next="52"></locus>
+                    <title type="complete" ref="LIT5889UndoingOfCharms"></title>
```
eu-genia, line 51: See also above. Some disambiguation is also needed between LIT5127MagicPrayer (no text is provided there) and other Prayers of the same title

StPetersburg/InEtn660715.xml
eu-genia: (I thought LIT1824Mafteh was the general record for Undoing of Charms)

StPetersburg/InEtn660715.xml
DariaElagina: I would be happy to do that, but the catalogue provides NO text, so no kind of disambiguation can be done. That's ehy this general record was created.

eu-genia: LIT1824Mafteh  (used e.g in BLadd16243) was a general record for unspecified Maftehe seray I assumed. 
Unless we can disambiguate (e.g. say that LIT5889UndoingOfCharms is for single short prayers (?) and LIT1824Mafteh for longer collections) the two IDs must be merged.

DariaElagina: It is in fact like that, LIT1824Mafteh is for the collection of magic prayers which is called Maftehe Seray (the text occupies many folia and has more or less stable content). LIT5889UndoingOfCharms is for single prayers for undoing of charms (no clue whether it belongs to the collection LIT1824Mafteh or not), there is already a number of such identified prayers in the database.

DenisNosnitsin1970: Someone should take a good witness of Mäftahe seray and write out part-prayers.

DariaElagina: Once we are sure that this witness is really representative :)

DenisNosnitsin1970: Well, there is always a certain risk that it is not. Otherwise we can wait untill someone will prepare a critical edition.

eu-genia: #https://github.com/BetaMasaheft/Works/pull/112

PR #385, opened 2020-04-29
Update DSEthiop12.xml

DenisNosnitsin1970:
Updates. We have to see how we shape some fields, for instance I am not sure that writing out of "Layout" in full detail makes sense for the catalogue (perhaps a summary would be enough: the texts are copied on paper of several, ruled with mistara not very skilfully. At a certain point, the quires were gathered and trimmed, and Bound. See my explanations concerning the scribal hands). I unified the problematic "supplication" on fols 80v-81r to the text that follows (I would shape it as "addition", but it looks like the space for it on the leaves was left for it (or for smth else?), i.e. "planned" in a way.). Typologically, it is a rare case but it happens.

### DIFF
--- a/DayrSuryan/DSEthiop12.xml
+++ b/DayrSuryan/DSEthiop12.xml
@@ -57,8 +57,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
 
                   <msItem xml:id="ms_i2">
                      <locus from="24r" to="37v"/>
-                     <title type="complete" ref="LIT4960Taabbeyo">Excerpt from the initial part of the prayer, with
-                     some omissions.</title>
+                     <title type="complete" ref="LIT4960Taabbeyo"></title>
                      <!--should be compared with pictures of MNC008 to confirm that the work is identical-->
                      <textLang mainLang="gez"/>
                      <incipit xml:lang="gez">
@@ -82,6 +81,8 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         <hi rend="rubric">ወምስለ፡
                            ፍቁሩ፡</hi> በመንፈስ፡ ኄር፡ ወምዕመን፡ ለዓለመ፡ <hi rend="rubric">ዓለም፡ ዓሜን፡ ወዓሜን። ። ለይኩን፡ ለይኩን፨ ። ።</hi>
                      </explicit>
+                     <note>The text is an excerpt from the initial part of the prayer, with
+                     some omissions.</note>
                   </msItem>
 
                   <msItem xml:id="ms_i3">
@@ -105,7 +106,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
 
                   <msItem xml:id="ms_i4">
                      <locus from="44r" to="59r"/>
-                     <title type="complete" ref="LIT1716Kidanz" xml:lang="gez">Copied with some omissions.</title>
+                     <title type="complete" ref="LIT1716Kidanz" xml:lang="gez"></title>
                      <textLang mainLang="gez"/>
                      <incipit type="supplication" xml:lang="gez">
                         <locus target="#44r"/>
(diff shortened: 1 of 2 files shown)


DayrSuryan/DSEthiop12.xml
```diff
@@ -57,7 +57,8 @@ schematypens="http://relaxng.org/ns/structure/1.0"
 
                   <msItem xml:id="ms_i2">
                      <locus from="24r" to="37v"/>
-                     <title type="incomplete" ref="LIT4960Taabbeyo">The initial part of the prayer, with
+                     <title type="complete" ref="LIT4960Taabbeyo">Excerpt from the initial part of the prayer, with
```
thea-m, line 60: I think that writing this information inside `<title>` might lead to problems, Maybe it would be better to have a `<note>`.

DayrSuryan/DSEthiop12.xml
```diff
@@ -196,9 +213,10 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         <hi rend="rubric">ስብሃት፡ ለአብ፡</hi> መፍቀሬ፡ ሰብ፡ ዘመጽአ፡ ወአድሃነነ፡ እምእደ፡ ጸላኢ፡ ዘአልቦ፡ ምህረት፡ ዘሰበረ፡ ሆሃተ፡ ብርት፡ ወቀጥቀጠ፡ መናስግተ፡ ዘሃፂን፡ ወዘአድከመ፡ ሀይለ፡ መሥቴማ፡ ዘገብረ፡ ዘንተ፡ ኵሎ፡
                         <pb n="87v"/>ወዘወሃበነ፡ እንቲዓሁ፡ ስልጣን፡ በዘ፡ ንክል፡ ንፍታህ፡ ለኩሉ፡ ማእሰር፡ <hi rend="rubric">ይእዜነ፡ ንሰብሖ፡ ለዘሎቱ፡</hi> ይደሉ፡ ሥብሃት፡ ለአለመ፡ አለም<hi rend="rubric">።</hi> አሜን፡
                      </explicit>
+                     <note>Supplication at the beginning of the work is added by a different hand, possibly later.</note>
```
thea-m, line 216: The supplication

DayrSuryan/DSEthiop12.xml
```diff
@@ -143,6 +140,21 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         ሀብተ፡ ጸጋ፡ ወኃይል፡ ለከ፡ እንዘ፡ ንዌድስ፡ ወኢንዓርፍ፡ ለዝሉፉ፡ በአልባቢነ፡ ናስተማስል፡ አምሳለ፡ መንግሥትከ፡ ቦቱ፡ በእንቲአከ፡ ወበእንተ፡ ፍቁር፡ ወልድከ፡ እንግዚእነ፡ ኢየሱስ፡ ዘቦቱ፡ ለከ፡ ስብሐት፨
                         ወእኂዝ፡ ለዓለመ፡ ዓለም፡ አሚን። ። ። ። ።
                      </explicit>
+                     <note>The incipit (<foreign xml:lang="gez">ዘነገሮሙ፡ እግዚእነ፡</foreign>) is similar to that in some manuscript
```
thea-m, line 143: I am not sure, do you think this can be somehow meaningfully summarised? I can also look into it with more concentration later.

thea-m:
Maybe these summarised statements for the layout can be given in a `<note>` in the first `<layout>`, and we can then decide to print only this.

DayrSuryan/DSEthiop12.xml
DenisNosnitsin1970: ok

DayrSuryan/DSEthiop12.xml
DenisNosnitsin1970: ok

DenisNosnitsin1970: ok

DayrSuryan/DSEthiop12.xml
DenisNosnitsin1970: this is meaningful, in a classical printed catalogue this would be a footnote.

DenisNosnitsin1970: I can put it back into <-- -->, and leave only the summary statement. This would be pity, however.

thea-m: No, no, then better to keep it.

PR #386, opened 2020-04-29
DS Ethiop 13

thea-m:
I am not really satisfied with the content description, I don't know this book well and had trouble identifying the excerpts present here.


DayrSuryan/DSEthiop13.xml
```diff
@@ -133,105 +153,101 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
-                           <q xml:lang="ar"/>
+                           <desc>Arabic shelfmark, the place for the number has been left empty.</desc>
+                           <q xml:lang="ar">رقم    <space/> طقوس</q>
                         </item>
                      </list>
                   </additions>
 
                   <bindingDesc>
                      <binding xml:id="binding">
-                        <decoNote xml:id="b1">No binding.
```
DenisNosnitsin1970, line 233: I would say: No boards.

DayrSuryan/DSEthiop13.xml
```diff
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
-            <title>Maṣḥafa zǝq</title>
```
DenisNosnitsin1970, line 8: I think the standard title of the work is: "Maṣḥafa ziq"

DayrSuryan/DSEthiop13.xml
```diff
@@ -48,14 +52,17 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                   <msItem xml:id="ms_i1">
                         <locus from="1r" to="32r"/>
                      <title type="incomplete" ref="LIT1979Mashaf"/>
-                        <textLang mainLang="gez"/>
-                        <note> Chants for individual
```
DenisNosnitsin1970, line 52: If we say "incomplete" in the case of chant mss or anything arranged according to the calendar, t would be helpful to say, somewhere, what is the last feast.

DayrSuryan/DSEthiop13.xml
thea-m: Yes, I think you looked again at an outdated version, it was already corrected in the PR

DayrSuryan/DSEthiop13.xml
thea-m: I tried, but had trouble. The version you reviewed is the old one. Could you try to look at the complete file that is under "Files changed"?
image
Thank you :)

DayrSuryan/DSEthiop13.xml
thea-m: I actually corrected to "The Book of Ziq", as the title of the description.

DayrSuryan/DSEthiop13.xml
DenisNosnitsin1970: The title of the work can be arranged as: Mashafa ziq za-Gwandar (catalogued like this in other catalogues, the attribute "of Gwandar" appears in the incipit).

DayrSuryan/DSEthiop13.xml
```diff
@@ -48,14 +52,17 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
-                     <notatedMusic>
+                        <textLang mainLang="gez" otherLangs="am"/>
+                    <notatedMusic>
                         <desc>The manuscript contains extensive musical notation,
-                          which has not been transcribed here. </desc>
+                          which has not been transcribed here.</desc>
                      </notatedMusic>
+                     <note>The manuscript seems to be an unfinished compilation of excerpts of the work, beginning regularly with
+                        <date ref="ethiocal:Maskaram1">1 Maskaram</date> commemorating <persName ref="PRS5741Johnthe"/>, but continuing
+                        immediately with the commemoration on <date ref="ethiocal:Hedar7">7 Ḫǝdār</date> of Qʷǝsqʷām.
```
DenisNosnitsin1970, line 62: I would leave out indication to zemmare (I know it from a different context, it will not be clear); for the rest is ok. I  think the text was copied with some omissions. It would be good to quote at least the first few words of the portions for the last two feasts, at least , in a "Note" or somehow ( i see the portion in EMIP01019_MihurGedam_044 frame 9 left), and also to say that the dates - - which were supposed to be rubricated - were inserted after fol. 23. Therefore, we cannot say the exact dates (It appears that some versions of Ziq have non-linear organisation of the feasts.) I would reduce the time range to 1900 - 1950. Small, nasty manuscript!

DenisNosnitsin1970: ( i see the portion in EMIP01019_MihurGedam_044 frame 9 left) - sorry, disregard this.

DayrSuryan/DSEthiop13.xml
thea-m: In the work record, you mean?

DenisNosnitsin1970: Yes. The attribute "of Gwandar" relates to the work.

DayrSuryan/DSEthiop13.xml
thea-m: By the last two feasts, do you mean Giyorgis and Qwesqwam on 21v? The end of the manuscript seems differently organised than the beginning

thea-m: Qwesqwam and Giyorgis are included completely in the explicit

DayrSuryan/DSEthiop13.xml
DenisNosnitsin1970: Sorry, but also here, why not "The Book of Ziq of Gwandar"?

thea-m: Very deep reason: because I forgot

thea-m:
@DenisNosnitsin1970

PR #387, opened 2020-04-29
Emip metadata april 25 2020

Ralph-Lee-UK:
Standard edits to EMIP collection

PR #389, opened 2020-04-29
EMIP Metadata

Ralph-Lee-UK:
Recommitted files to address problems with today's pull request

### DIFF
--- a/EMIP/1-1000/EMIP00025.xml
+++ b/EMIP/1-1000/EMIP00025.xml
@@ -257,7 +257,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
           <change when="2018-08-08" who="RL">Corrected art theme references</change>
           <change when="2020-04-01" who="RL">Reformatted ms items, additions and extras, and added image links</change>
           <change when="2020-04-18" who="RL">Corrected ǝ character</change>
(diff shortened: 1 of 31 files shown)


Ralph-Lee-UK:
This is the replacement pull request for the EMIP metadata edits for this week.

Ralph-Lee-UK:
I haven't deleted the other branch, but we can do that once this is completed.

PietroLiuzzo:
https://github.com/BetaMasaheft/Manuscripts/pull/387

PietroLiuzzo:
there are some files which only have a `<change>` but not the described change. was that already in a previous branch and PR?

Ralph-Lee-UK:
Can you point me to an example?

PietroLiuzzo:
the second file in the files changed view for me is EMIP000037, where it looks like only a change element was subsituted

Ralph-Lee-UK:
Yes! This is where the conflict was that I had anticipated. The change to the code that I did and that Thea did was identical, as it only involved the substitution of the code for the work, replacing a work file that has been deleted. So I am confident that in these cases that this is correct, and that nothing has been lost.

PietroLiuzzo:
then dorothea's change should not be removed from the list of changes, please

Ralph-Lee-UK:
I think I have completed this. I have deleted my change statement where it described exactly the same thing that Dorothea had done.

PR #393, opened 2020-04-30
DS Ethiop 14

thea-m:


DayrSuryan/DSEthiop14.xml
```diff
@@ -162,11 +173,17 @@ schematypens="http://relaxng.org/ns/structure/1.0"
 
                <history>
                   <origin>
-                     <origDate notBefore="1900" notAfter="1980"/>
+                     The manuscript can be dated palaeographically to the
```
DenisNosnitsin1970, line 176: I would not write in each case "palaeographically". If there is no other indication, it goes by default that your dating is palaeographical. At some places we adopt different formulations that should actually not be very diverging.

DayrSuryan/DSEthiop14.xml
```diff
@@ -162,11 +173,17 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+                     <origDate notBefore="1900" notAfter="1980" evidence="lettering">early-middle 20th century</origDate>.
```
DenisNosnitsin1970, line 177: I would agree with "middle", but 1980 is not middle. Middle is maximum 1960.

DayrSuryan/DSEthiop14.xml
```diff
@@ -39,9 +39,9 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                      <title type="incomplete" ref="LIT2000Mazmur"/>
                         <textLang mainLang="gez"/>
                      <note>Contains Psalms 1-6 and 8-17; Psalm 7 is missing.
-                     Numeration of Psalms is not consequent. Nummer 1 is used two times;
-                   Psalms 5, 6, 8, 9, 11 have no numbers. Recent numbers written
-                   in pencil, Eastern Arabic numerals, are added to all Psalms.</note>
+                     The original numeration of Psalms is not consequent. Number 1 is used two times;
```
DenisNosnitsin1970, line 42: I a "note" after the title, you need to say that it is a portion of the Psalter (only Ps 1-6 and 8-17).

DayrSuryan/DSEthiop14.xml
thea-m: I had copied this formulation from the first example description of DSEthiop1...I agree that we should use the same formulations in the descriptions. I like this formulation because it is very clear and transparent.

DayrSuryan/DSEthiop14.xml
thea-m: So "A portion of the Psalter (Ps 1-6 and 8-17)" instead of the current formulation?

thea-m: So "A portion of the Psalter (Ps 1-6 and 8-17)" instead of the current formulation?

DayrSuryan/DSEthiop14.xml
DenisNosnitsin1970: Yes, but still better avoid this word, which may turn out provocative. A certain problem of this palaeographical formulations in the descriptions (esp. in Ethiopic catalogues) is that the argumentation is "locked" on itself. In most of them, no clear arguments are proposed as to why exactly the script is to be assigned to one or another time (Images are sometimes proposed, but mostly without clear explanations). It is a little bit like "It is 20th cent. because I say so". Actually, we evaluate a bit of everything, not only the hand. Historical context and internal evidence, condition of the parchment, ink, text arrangement.

DayrSuryan/DSEthiop14.xml
DenisNosnitsin1970: Yes, for instance.

thea-m:
@DenisNosnitsin1970

PR #396, opened 2020-05-02
Update DSEthiop16.xml

DenisNosnitsin1970:
Completing.


DayrSuryan/DSEthiop16.xml
```diff
@@ -116,39 +121,46 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
                </physDesc>
-                  
+
                <history>
-                  <provenance>
-                  </provenance>
-               </history>
-               
+                              <origin>
+                                 <origDate notBefore="1920" notAfter="1960" evidence="internal">First half of the 20th century.</origDate>
```
thea-m, line 159: by "internal", I understand that a date is specified in the manuscript, which does not seem to be the case here?

PR #403, opened 2020-05-05
Didactic mss

DariaElagina:
Concerning 3052887a and 3052887b I have not used msFrag because there is not enough evidence that they once constitued one solid ms.

### DIFF
--- a/StPetersburg/InEtn3052887b.xml
+++ b/StPetersburg/InEtn3052887b.xml
@@ -1,0 +1,142 @@
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/SChAth/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/SChAth/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="InEtn3052887b" xml:lang="en" type="mss">
+    <teiHeader xml:base="https://betamasaheft.eu/">
+        <fileDesc>
+            <titleStmt>
+                <title>Syllabary of the Ethiopic alphabet</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="DE"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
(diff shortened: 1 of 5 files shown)


StPetersburg/InEtn3052887b.xml
```diff
@@ -0,0 +1,142 @@
...
+          </langUsage>
+        </profileDesc>
+        <revisionDesc>
+            <change who="DE" when="2020-05-05">Created entity, added description and content.</change>
+        </revisionDesc>
+    </teiHeader>
+    <text xml:base="https://betamasaheft.eu/">
+        <body>
+            <div type="bibliography"/>
+         <!---->
```
eu-genia, line 139: a relation between a and b could be helpful, too

DariaElagina: I went through the list of relations but could not find an appropriate. Do you have a idea which one I could use?

eu-genia: if the sdc relations are not helpful you can always use a dc:relation with a desc

PietroLiuzzo: Sorry, I do not understand what you are trying to state a relation about. We can make a statement about a previous UniCirc with a hypothetic annotation, it does not need to be 100% sure (what is so, after all...)

PietroLiuzzo: If you explain in words briefly, I will be glad to try and translate in appropriate relations

DariaElagina: Thank you Pietro, it might have been that InEtn3052887a and InEtn3052887b once constitued one single scroll. However, this is just a hypothesis

PietroLiuzzo: I would add something like this
`<relation active="InEtn3052887a#UniCirc3" name="sdc:hasCertainty" passive="sdc:Hypothetique"/>
<relation active="InEtn3052887a#UniCirc3" name="sdc:undergoesTransformation" passive="InEtn3052887a#tr1"/>
<relation active="InEtn3052887a#tr1" name="sdc:hasCertainty" passive="sdc:Hypothetique"/>
<relation active="InEtn3052887a#tr1" name="sdc:hasTransformationModel" passive="sdc:D3"/>
<relation active="InEtn3052887a#tr1" name="sdc:produces" passive="InEtn3052887a"/>
<relation active="InEtn3052887a#tr1" name="sdc:produces" passive="InEtn3052887b"/>`

eu-genia: @DariaElagina do you want to add the relations? then the branch can be merged

DariaElagina: Done. Thank you @PietroLiuzzo !

PR #404, opened 2020-05-05
Update DSEthiop21.xml

DenisNosnitsin1970:
Finishing. Not much can be added.


thea-m:
I updated the ID to LIT2000Mazmur, because it is not a complete psalter

PR #408, opened 2020-05-06
DS Ethiop 22

thea-m:
No images from the middle of the manuscript...

### DIFF
--- a/DayrSuryan/DSEthiop10.xml
+++ b/DayrSuryan/DSEthiop10.xml
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
-            <title>Miscellaneous collection of hymns and prayers, Teaching of Mysteries</title>
+            <title>Miscellaneous collection of hymns and prayers, Doctrine of Mysteries</title>
             <editor key="DR"/>
             <editor key="DN"/>
             <editor key="AB" role="generalEditor"/>
--- a/DayrSuryan/DSEthiop10.xml
+++ b/DayrSuryan/DSEthiop10.xml
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
-            <title>Miscellaneous collection of hymns and prayers, Doctrine of Mysteries</title>
+            <title>Miscellaneous collection of hymns and prayers, Teaching of Mysteries</title>
             <editor key="DR"/>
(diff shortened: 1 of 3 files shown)


DayrSuryan/DSEthiop10.xml
```diff
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
-            <title>Miscellaneous collection of hymns and prayers, Teaching of Mysteries</title>
+            <title>Miscellaneous collection of hymns and prayers, Doctrine of Mysteries</title>
```
DenisNosnitsin1970, line 8: Please, let us live with "Teaching". It is the standard title. What is in the work is not a "doctrine".

DayrSuryan/DSEthiop22.xml
```diff
@@ -79,17 +92,34 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
 
                <history>
-                  <provenance>
-                  </provenance>
+               <origin>
+                  The manuscript is written in the same hand, same type of text block and binding as <ref type="mss" corresp="DSEthiop22"/>, 
+                  which according to its colophon was written by  <persName ref="PRS11944GabraM">Gabra Māryām of Māḫbara Səllāse</persName>
+                  for <ref target="INS0478DS"/>, in <date> 1951/52 CE</date>,
+                  in the tenure of <persName ref="PRS12955Tawfilus"/>,
+                  in <placeName ref="LOC6126WadinN">ʾAsqeṭəs</placeName>. It is therefore very probable that this manuscript was
```
DenisNosnitsin1970, line 119: the information "same type of text block and binding as" would do better as a separate sentence after "Asqetes".

thea-m: But the Asqetes part only refers to Ethiop. 21, it would make not much sense without the information on why it is relevant here?

DenisNosnitsin1970: We can say smth like "copied in the same time and place as DSEthiop21", and specify in brackets (or even not - 21 will be just above).

thea-m:
@DenisNosnitsin1970

DayrSuryan/DSEthiop22.xml
```diff
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
    <teiHeader xml:base="https://betamasaheft.eu/">
       <fileDesc>
          <titleStmt>
-            <title>Tǝmhǝrta ḫǝbuʾāt</title>
+            <title>Teaching of Mysteries and Supplications</title>
```
DenisNosnitsin1970, line 8: Better: Teaching of Mysteries, Supplications.

PR #427, opened 2020-05-14
Ds ethiop7revision

thea-m:
Could you please check the wording of history as it is now?


thea-m:
What do you mean? The one by PRS11959Pawlos is now a colophon

PR #442, opened 2020-05-18
Ds ethiop7

DenisNosnitsin1970:
Minor spelling changes.


DayrSuryan/DSEthiop7.xml
```diff
@@ -343,6 +343,23 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                            </q>
                         </item>
 
+                        <item xml:id="a3">
```
thea-m, line 346: If you decided now for an addition, it would be good to delete the colophon, right now I think we have both

PR #453, opened 2020-05-19
Ds ethiop13 dn

DenisNosnitsin1970:
A remark on Musical Notation in the Note which will be visualized.


thea-m:
I've moved the information to the musical notation element and will ask Pietro to visualise this.

PR #469, opened 2020-05-20
Emip metadata 14 may 2020

Ralph-Lee-UK:
Standard edits to EMIP collection


PietroLiuzzo:
only validation checks done

PR #476, opened 2020-05-25
De ethiop1correction

DenisNosnitsin1970:
minor changes


thea-m:
It seems that there were too many merges here. No there are no changes visible anymore. 
If you look through the commit history, can you remember which ones were the significant changes, then we can reconstruct them?

DenisNosnitsin1970:
There are a few slight changes, nothing big. I think it is better to merge and I will look though. No idea how the ygot stuck.

thea-m:
If I understand it correctly, if we merge now, there will be no changes, because none are visible anymore.

thea-m:
image

DenisNosnitsin1970:
Ok, let it be merged.

thea-m:
:(

PR #481, opened 2020-05-25
Update BLorient753.xml

DariaElagina:


LondonBritishLibrary/orient/BLorient753.xml
```diff
@@ -1,17 +1,14 @@
...
         <fileDesc>
             <titleStmt>
-                <title>Qalemǝnṭos, Ṭǝnta Haymānot, The religious Romance (The History of Baralām wa-Yǝwāsǝf)
-                    
-                    
-                    
-                    
+                <title>Qalemǝnṭos, Maṣḥafa ṭǝnta hāymānot, Baralām wa-Yǝwāsǝf
                 </title>
                 <editor key="SG"/>
```
eu-genia, line 10: I think in all those cases, SG should only be mentioned in the change element as he only created stubs, without the proper description

LondonBritishLibrary/orient/BLorient753.xml
```diff
@@ -133,13 +81,244 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+                        </physDesc>
+                    </msPart>
+                    <msPart xml:id="p3">
+                        <msIdentifier><idno></idno></msIdentifier>
+                        <msContents>
+                          <msItem xml:id="p3_i1">
+                            <locus target="#157"></locus>
+                            <title type="incomplete" xml:lang="en" ref="LIT1932Mashaf">
+                              <certainty locus="value" match="../@ref" cert="low"></certainty>
+                              Excerp from a Lectionary (New Testament)</title>
```
eu-genia, line 295: Excerpt

LondonBritishLibrary/orient/BLorient753.xml
```diff
@@ -133,13 +81,244 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+                          </msItem>
```
eu-genia, line 296: since you add xml lang en to the title it would be even more important to add an xml lang gez element for the text to avoid confusions

PR #483, opened 2020-05-25
Ds ethiop6 dn

DenisNosnitsin1970:
PAlaeography added, slight dating change.


DayrSuryan/DSEthiop6.xml
```diff
@@ -304,7 +304,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
 
                <history>
                   <origin>
-                     <origDate notBefore="1880" notAfter="1948" evidence="internal-date">Late 19th century to 1948.</origDate>
```
thea-m, line 307: the internal date evidence is based on the date 1948 in a1. If we change to a palaeographic dating, I would change it to lettering

DenisNosnitsin1970: Done

thea-m: Now it is again 1948? I am confused...

PR #510, opened 2020-05-28
Ds ethiop19 dn

DenisNosnitsin1970:
Palaeography filled, some minor changes.


DayrSuryan/DSEthiop19.xml
```diff
@@ -215,7 +216,14 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                               </item>
 
                               <item xml:id="fol2" resp="DN DR">
-                                 <desc>Foliation in pencil in the centre bottom of the recto. This foliation is used as reference in this record.</desc>
+                                 <desc>Foliation in pencil in the centre bottom of the recto. This foliation is used as
+                                   reference in this record. (transferred hereto from "Note" under ms_i1:
```
thea-m, line 220: I don't really understand why the part starting with "transferred hereto" has been included here? This refers to the texts contained in the manuscript, not the foliation.

DenisNosnitsin1970: Sorry, 
this was ununderstandable  comment that was in "Note". There is no way to solve it, but it contains some information. I just removed it from the printed version.

thea-m: Ah, so do I understand correctly that you want this information to be kept somewhere in the record, but not in note, because you don't want it printed in the article?

DenisNosnitsin1970: Yes. We found it ununderstandable, for the moment.

thea-m: Maybe the first part, 
```
Portions of the work for Tuesday and Wednesday corresponding to
                                      <bibl><ptr target="bm:Leander1922Arganona"/>
                                      <citedRange unit="page">34-103</citedRange></bibl>.
```
is not necessary to be kept, because the same information is given in the sub msItems. Could it be deleted?
 ```
The text on <locus target="#7"/> and <locus target="#8"/> is in disorder,
                                      the correct sequence of text is
                                      <locus target="#6"/>, <locus target="#7"/>, <locus target="#8"/> and <locus target="#9"/>.
```
could maybe be added as a note to collation, then it also would not be printed. What do you think?

Comment (DenisNosnitsin1970) [reply DayrSuryan/DSEthiop19.xml:220]: I also though in this way, but the appearance in this note should be harmonized with Ethiop. 18, it has the same. If you want, please do. 
Yes, sure, put it to the collation. I assume the mistery was in the difference verso and recto, which got lost.

Comment (thea-m) [reply DayrSuryan/DSEthiop19.xml:220]: Looking at the history of the file, something must have gone wrong on my commit of 5 May. The original sentence was
 ```
<note>The text from <locus target="#6v"/> continues on <locus target="#8r"/>, the text from <locus target="#8v"/>
                        continues on <locus target="#7r"/>, which continues on <locus target="#9r"/>.</note>
```
. I would trust my old self and correct to  
```
the correct sequence of text is  <locus target="#6"/>, <locus target="#8"/>, <locus target="#7"/> and <locus target="#9"/>.
```
Would you want this then to appear in the article or should it also be put in collation?

DenisNosnitsin1970: Aha! This is  a different matter. Then just reinstall where it was, or at a different convinient location.

thea-m:
If it's fine now, please merge.

PR #512, opened 2020-05-28
Ds ethiop22 dn

DenisNosnitsin1970:
Palaeography filled, minor changes.


DayrSuryan/DSEthiop22.xml
```diff
@@ -80,14 +80,12 @@ schematypens="http://relaxng.org/ns/structure/1.0"
 
                         <foliation>The manuscript is not foliated.</foliation>
                         <collation>
-                           <note>The text block is a multi-quire large format notebook
```
thea-m, line 83: As I understood the result of https://github.com/BetaMasaheft/Documentation/issues/1418, we agreed to put this in `<p>` in `<support>`. Can I move it there?

thea-m:
I think this is an old branch (the famous DS Ethiop. 120 is back!). It is not dramatic, but I don't want any complications, so after we have finished the conversation, I will copy the changes to a new branch and make a new PR

DenisNosnitsin1970:
Strange, i see the phantom 120 is back, but I did not touch it - at least I think i did not.

DayrSuryan/DSEthiop22.xml
DenisNosnitsin1970: OK, of course.

thea-m:
Must have to do with the resucitation of the old branch, I don't know either...

PR #529, opened 2020-05-30
Ds ethiop10 dn

DenisNosnitsin1970:


DenisNosnitsin1970:
Sorry, I mistakenly sent the record into false branch. I assume it cannot be discarded, cant it? Should it be merged, and should I make the same changes in the record in the correct branch?

thea-m:
The changes seem fine to me. You probably reactivated an old branch which led to a lot of merging, but whether that branch is called 10 or 11 is irrelevant. Are the changes visible here the ones that you want to have?

thea-m:
image

DenisNosnitsin1970:
Yes, they are.

thea-m:
Then it should be fine.

PR #534, opened 2020-05-30
Ds ethiop16 dn

DenisNosnitsin1970:
Reference to an edition of the text is added.


thea-m:
Now I think that maybe the title for the manuscript could be more precisely "Excerpt from the Canon of al-Muʿallaqa"?

DenisNosnitsin1970:
Done.

PR #547, opened 2020-05-31
Ds ethiop22 dn

DenisNosnitsin1970:
minor change


DenisNosnitsin1970:
Hier kommt mal wieder Phantom Ethiop. 120, wie können wir ihn bekämpfen?

thea-m:
image
 ;)
This happens, I think, because you reactivated an old branch (which happens because GitHub desktop does still show deleted branches, for which I don't have a solution...). 
For the next times, it could be better if you would create a new branch from the current master branch.

PR #571, opened 2020-06-05
further revisions

thea-m:


thea-m:
In cases where there is now "commissioner" or "commissioner and owner" or both?

PR #572, opened 2020-06-05
Ds small corrections

DenisNosnitsin1970:
Minor formal revisions.


DayrSuryan/DSEthiop3.xml
```diff
@@ -760,6 +753,8 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                               <locus from="101r" to="102v"/>
                            </extent>
                            <collation>
+                             In <ref target="#p1">Unit 1</ref>, the dimensions of the written area of and the number ow written lines slightly change
```
thea-m, line 756: I think that this information would be better kept in layout, not collation

DayrSuryan/DSEthiop1.xml
```diff
@@ -664,7 +664,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                   <bindingDesc>
                      <binding xml:id="binding">
                         <decoNote xml:id="b1">
-                           Two wooden boards covered with reddish brown blind-tooled leather.
+                           Bound in Ethiopia. Two wooden boards covered with reddish brown blind-tooled leather.
```
thea-m, line 667: is there a conceptual difference between "Bound in Ethiopia" here and "Ethiopian" elsewhere?

DenisNosnitsin1970: It should be "Ethiopian binding".

DenisNosnitsin1970:
S. DS Ethiop 3 - layout like this? we want that this info would be visible in the catalogue.

thea-m:
it will not be visible yet, but I wrote it down, we will find a solution

DenisNosnitsin1970:
Thanks!
We want to give the reader creadible information :)

PR #583, opened 2020-06-08
Ds small corrections

DenisNosnitsin1970:
minor changes


DayrSuryan/DSEthiop23.xml
```diff
@@ -253,7 +253,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                      <persName ref="PRS11985PetrosGS">Ṗeṭros Gabra Śǝllāse</persName> indicates
                      that he finished the first part of <ref target="#ms_i4">Text IV (Missal in Tigrinya)</ref>
                       on <date>21 Sana 1975 EC (= 29 June 1983 CE)</date>,
-                      in <origPlace ref="wd:Q61">Washington, D.C.</origPlace>. The last work contained in the manuscript,
```
thea-m, line 256: I would keep the . :)

DenisNosnitsin1970: Du meinst, den Punkt? Wir haben dann zwei Punkte im PDF.

PietroLiuzzo: please do not decide your encoding on the basis of the output. if there is an extra dot somewhere in an output where it should not be, we will fix it in the output. Thanks!

DenisNosnitsin1970: Ok, done.

PR #587, opened 2020-06-09
Update BLorient745.xml

DariaElagina:
Abuna Takla Haymanot from the a2 will be identified, I need to check the EAe.


eu-genia:
The only 18th century TH we have is the ǝccage (PRS9155TaklaHa), but I would not be at all sure that the ownership note speaks of a person and not of a church dedicated to the saint TH.

DariaElagina:
yes, you are right!

eu-genia:
(maybe a2 could still be provided with a note/translation, to make it better searchable, or at least I would link TH `<persName ref="PRS9162TaklaHa" type="tabot" cert="low">Takla Haymanot</persName>` so that if someone looks for a church dedicated to TH one may come across and if possible at some point say more)

PR #605, opened 2020-06-16
Update BLorient758.xml

DariaElagina:


LondonBritishLibrary/orient/BLorient758.xml
```diff
@@ -55,57 +54,81 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+                        </q>
+                              </item>
+                              <item xml:id="a2">
+                                <desc type="Genealogy"><title type="complete" ref="LIT3143Ledata" xml:lang="gez">
+                                ልደተ፡ መነኮሳት፡</title>Genealogy of monks, beginning with
+                                <persName ref="PRS1962Anthony"></persName> and ending with <persName ref="PRS9162TaklaHa"></persName>.
+                              </desc>
+                              </item>
+                              <item xml:id="a3">
+                                <desc type="GuestText">The names of the doctors of <placeName ref="INS0341DL"></placeName>
```
eu-genia, line 109: "doctors"? better maybe mamher or rather abbots (look like abbots to me)? or should we leave Wright's version?

LondonBritishLibrary/orient/BLorient758.xml
```diff
@@ -31,22 +31,21 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
                     <msContents>
                         <summary/>
-                        
-                        
-                        
+                        <msItem xml:id="ms_i1">
+                          <title type="complete" ref="LIT1948Mashaf"></title>
+                          <textLang mainLang="gez"></textLang>
+                          <colophon xml:id="coloph1" xml:lang="gez"><locus from="143a"></locus>
+                          <seg part="I">ዛቲ፡ መጽሐፍ፡ <add>ዘአባ፡ ወልደ፡ ስምዖን፡</add> ዘተዓገላ፡ ወዘሠረቃ፡</seg>
```
thea-m, line 44: This does not seem to be what we understand by colophon...of course, now it is difficult to say if it is an addition or not...I would maybe tentatively go with explicit here.
And could you create a person ID for Walda Səmʿon?

LondonBritishLibrary/orient/BLorient758.xml
```diff
@@ -31,22 +31,21 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+                          <note>The colophon has been tampered with.</note>
```
thea-m, line 45: "tampered with" sounds a bit colloquial to me.. Do you think he refers to the addition of the new owner's name? Maybe "has been modified later" or something similar?

LondonBritishLibrary/orient/BLorient758.xml
thea-m, line 109: is doctors here standard English in this context? Maybe scholars?

LondonBritishLibrary/orient/BLorient758.xml
DariaElagina: I think it is not an addition, it was classified as a colophon by Wright, and it is just the very beginning, no idea what follows...

thea-m: I think that is terminology is just different than ours. In my experience, it would not be typical for a condemnation to be followed by something on the production of the manuscript (the other way round would be more usual), but of course we cannot know.

DariaElagina: I see, I went for excplicit for clarity

thea-m: It is always difficult :)

PR #613, opened 2020-06-19
Bd aeth29

SusanneHummel:
Dear all, 
finally, I am releasing the famous Ms with some remarks. I would like to ask your help concerning the following points:
1. As I showed you, there are two notes written in pencil at the beginning of the Ms that I cannot clearly read (encoded as a3 and a4). I would be grateful if you could have a look at them.
2. I found a reference to this very Ms in Sean’s Phd thesis (related to the beautiful blind-tooling). At the same pages (p. 235, 239) Sean also refers to BDLaethb3. 
3. Ms_i6 is a short prayer. Therefore, I typed the whole text. I tried to follow what Pietro suggested ("using div[@type='edition']" https://github.com/BetaMasaheft/Documentation/issues/1309), but the div-element is not allowed in Ms_item; so he probably meant to include this information somewhere else, or most probably I did something wrong. 
4. There are a few pending issues related to some phenomena in the mss. I commented all those cases so that I won’t forget them later; you might simply ignore those for the review now. I write it down for documentation:
a)	one issue is related to the following scribal phenomenon: the scribe uses original characters of a given word to transform them to another character (he does not erases the (whole) letter but simply changes it (adding a stroke or deleteing a tiny part) in order to get a another letter and reach a new meaning) (ms_i2)
b)	another pending issue is how to encode homoioteleutons (ms_i4)
c)	ms_i5: If there are work authority files that do not provide the fidal, translations and transcription of the Ethiopic title of the work, Dorothea and me decided for now just to comment it and to correct and improve the BM authority files later; at first we would need to decide how we would like to have it displayed in our catalogue.


thea-m:
Dear @SusanneHummel , great, thank you!
You accidentally made a Pull Request into the master branch, not Juel-Jensen (You can see this, among other places, in the fact that it says you changed 67 files). I will close this pull request now. Could you re-pull request your BdAeth29 branch into Juel-Jensen? 
You should see this when you make a Pull Request (with BdAeth29  of course, not PersonsBLorient535):
image

SusanneHummel:
Oje. Thanks a lot!

PR #619, opened 2020-06-22
updated occurrences of @n in locus

PietroLiuzzo:
this was previously used to mark a line, with the new referencing system this can be done inside `@from`, `@to`, `@target`. In this commit there are also some fixes of instances of `@from` `@n` , a mistake for `@from` `@to` and a few other corrections. especially intriguing are cases of locus possibly being marked up out of context, see  BAVet36 . I also added a couple of instances of `rec[@cRef]`


PietroLiuzzo:
https://github.com/BetaMasaheft/Documentation/issues/1162

PietroLiuzzo:
in EMIP02235 is simply a welformedness error.

PietroLiuzzo:
some files are not valid as they already were not and I was only making this one change.

PR #635, opened 2020-07-02
LIT4270 fixed, facs fixed

eu-genia:


eu-genia:
sorry there is now a bit of overlapping with https://github.com/BetaMasaheft/Manuscripts/pull/636 but there the only real corrections are the last two commits

DenisNosnitsin1970:
ok

PR #638, opened 2020-07-03
BLorient537

thea-m:
Slightly instransparent catalogue description occasionally.
And is there another term for "royal umbrella"?


DenisNosnitsin1970:
I would say, it may be also just "umbrella". Unless umbrella is a specific attribute of the Ethiopian kingship (it is possible).

PR #644, opened 2020-07-06
BDLaethb5 example

thea-m:
Here is another preliminary PR, as previously announced, to check in with you before I continue with the entire manuscript. 
Please just have a look at the month of Maskaram. I have now described all possible textual units, and given incipits at the beginning and explicits at the end of the month (also note the last salām and subscription in div, since they are fully quoted).
My questions:
- Do we stick with this depth of description? I would tend towards yes. Another question would be how to print this, and here I could imagine for example to only list the eventual differences from the edition.
- What do you think of the amount of text given?
- Right now, I only referred to the textparts in the work record, as usual. The issue with the Senkessar is that there is some variance in the names of the commemorated persons, and those in the work record were based on the edition and then developed depending on the manuscripts in which the new commemorations were attested. If we want to print the exact names as given in the manuscript (which would make sense), we would need to have:
 `<title type="complete" ref="LIT4035SenkessarDL#Mask30Salusi">Commemoration of Śālusi</title>`.

Let me know what you think!


thea-m:
Aah PR to master, not Juel-Jensen, ignore this, I'll open a new one, sorry!

PR #645, opened 2020-07-06
BDLaethb5

thea-m:
Here is another preliminary PR, as previously announced, to check in with you before I continue with the entire manuscript. 
Please just have a look at the month of Maskaram. I have now described all possible textual units, and given incipits at the beginning and explicits at the end of the month (also note the last salām and subscription in div, since they are fully quoted).
My questions:
- Do we stick with this depth of description? I would tend towards yes. Another question would be how to print this, and here I could imagine for example to only list the eventual differences from the edition.
- What do you think of the amount of text given?
- Right now, I only referred to the textparts in the work record, as usual. The issue with the Senkessar is that there is some variance in the names of the commemorated persons, and those in the work record were based on the edition and then developed depending on the manuscripts in which the new commemorations were attested. If we want to print the exact names as given in the manuscript (which would make sense), we would need to have:
 `<title type="complete" ref="LIT4035SenkessarDL#Mask30Salusi">Commemoration of Śālusi</title>`.

Let me know what you think!

### DIFF
--- a/OxfordBodleian/Juel-Jensen/BDLaethb5.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethb5.xml
@@ -39,43 +39,192 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         
                         <msItem xml:id="ms_i1">
                             <locus from="4ra" to="212rc"/>
-                            <title type="complete" ref="LIT4968SenkessarSec"/>
+                            <title type="complete" ref="LIT4035SenkessarDL#FirstHalf"/>
                             <textLang mainLang="gez"/>
                             
                             <msItem xml:id="ms_i1.1">
-                                <locus from="4ra" to="46va"/>
-                                <title type="complete" ref="LIT4968SenkessarSec"/>
-                                <textLang mainLang="gez"/>
+                                <locus from="4ra" to="4rb"/>
+                                <title type="complete" ref="LIT4035SenkessarDL#MaskIntroduction"/>
                             </msItem>
                             
                             <msItem xml:id="ms_i1.2">
-                                <locus from="46va" to="78vc"/>
-                                <title type="complete" ref="LIT4968SenkessarSec"/>
-                                <textLang mainLang="gez"/>
+                                <locus from="4rb" to="46va"/>
+                                <title type="complete" ref="LIT4035SenkessarDL#Maskaram"/>
+                                
+                             <!--   <msItem xml:id="ms_i1.2.1">
+                                    <locus from="4rb" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask1"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.2.2">
+                                    <locus from="" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask2"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.2.3">
+                                    <locus from="" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask3"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.2.4">
+                                    <locus from="" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask4"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.2.5">
+                                    <locus from="" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask5"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.2.6">
+                                    <locus from="" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask6"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.2.7">
+                                    <locus from="" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask7"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.2.8">
+                                    <locus from="" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask8"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.2.9">
+                                    <locus from="" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask9"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.2.10">
+                                    <locus from="" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask10"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.2.11">
+                                    <locus from="" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask11"/>
+                                </msItem>
+                                
+                                <msItem xml:id="ms_i1.2.12">
+                                    <locus from="" to=""/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Mask12"/>
+                                </msItem>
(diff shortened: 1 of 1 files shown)


OxfordBodleian/Juel-Jensen/BDLaethb5.xml
```diff
@@ -37,9 +37,2085 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                <msItem xml:id="ms_i1.7.31">
+                                    <locus from="211rc" to="212rc"/>
+                                    <title type="complete" ref="LIT4035SenkessarDL#Yak30"/>
+                                </msItem>
+                            </msItem>
+                            
+                            <colophon xml:id="coloph1" xml:lang="gez">
+                                <locus from="212rc" to="212rb"/>
+                                <note xml:lang="en">The colophon begins in column c and continues in columns a and b.
+                                    It states that the manuscript was completed in the time of <persName ref="PRS10473Yostos"/> (r. 1711-1716)
```
SusanneHummel, line 2106: maybe to be added "King" Yostos and "Patriarch" Marqos

OxfordBodleian/Juel-Jensen/BDLaethb5.xml
```diff
@@ -37,9 +37,2085 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                        <locus target="#46ra"/>
+                                        <title type="complete" ref="LIT4035SenkessarDL#Mask30John"/>
+                                    </msItem>
+                                    
+                                    <msItem xml:id="ms_i1.2.30.4">
+                                        <locus from="46rb" to="46rc"/>
+                                        <title type="complete" ref="LIT4035SenkessarDL#Mask30Salusi"/>
+                                        <explicit xml:lang="gez">
+                                            <locus target="#46rc"/>
+                                            ወሶበ፡ ሰምዓ፡ ሊአ፡ ጳጳሳት፡ ፈቀደ፡ ይንሥአ፡ ለይእቲ፡ ቆብዕ፡ ይትባረክ፡ እምኔሃ<hi rend="rubric" rendition="#partialRubric">፨</hi>
```
SusanneHummel, line 1270: This is the explicit of the commemoration of the last day of Maskaram. I wonder whether it makes sense to provide the explicit of the last commemoration notice. it is not the explicit of the month (unlike the beginning of the month, which refers to the whole month and not only to the first commemoration notice.  
As you provide the explicit of the month under "div @ edtion", I twould suggest that toghether with the incipit of the month (quoted under ms item) the text is sufficiently described and made identifiable.

OxfordBodleian/Juel-Jensen/BDLaethb5.xml
```diff
@@ -37,9 +37,2085 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                </incipit>
+                                
+                                <explicit xml:lang="gez">
+                                    <locus from="4ra" to="4rb"/>
+                                    ዘጻድቃን፡ ወዘሰማዕታት፡ ወዘሊቃነ፡ ጳ<hi rend="rubric">ጳሳት፡ ቅዱሳን፡ ወዘገዳማውያን፡ መነኮሳት፡</hi> ከመ፡ ይግበሩ፡
+                                    ተዝካሮሙ<hi rend="rubric" rendition="#partialRubric">።</hi>
+                                     ወያንብቡ፡ ገድሎሙ፡ ኵሎ፡ አሚረ፡ እምሠርቀ፡ አሚሩ፡
+                                    ለ<cb n="b"/><hi rend="rubric">መስከረም፡ እስከ፡ ተ</hi>ፍጻሜተ፡ ዓመት<hi rend="rubric" rendition="#partialRubric">፨</hi>
+                                    ወ<hi rend="rubric">ኮነ፡ አስተጋብአቶሙ፡</hi> ለዝንቱ፡ መጽሐፍ፡ በ<hi rend="rubric">፱፻፡ ፷፡ ወ፫፡ ዓመተ፡</hi> ሰማዕታት፡ ንጹሓን<hi rend="rubric" rendition="#partialRubric">፨</hi>
+                                    <seg type="supplication">ጸሎቶሙ፡ ወበረከቶሙ፡ የሀሉ፡ ምስለ፡ ፍቁራኒሆሙ፡ <hi rend="rubric">ተክለ፡ ማርያም፡</hi>
```
MassimoVilla, line 68: New Person records for Takla Māryām and Nǝwāya Māryām should be created, right?

thea-m:
Hi everyone, thank you for the comments! This is really only an example for the structure of the content description, everything else is not finished!

MassimoVilla:
Thank you Thea,

I think we have to be aware, as usual of the two-layer structure of our initiative: the BM online database, which can virtually accommodate an endless amount of information, and our printed output. I am far from setting the agenda for the BM database, but in theory I'd be in favour of being as much (reasonably) detailed as possible in order to provide BM with the highest amount of data. On the other hand, the printed catalogue imposes some serious constraints, i.e. mainly the number of pages we can devote to one single manuscript (or two, since ms b.6 should be the second part of the Sǝnkessār, right?). So, provided that in my view the depth of description suits the BM database expectations, we might make some transformation tests in order to reduce the needed space without losing too much data. For example, in the book we can give the name of all the commemorated saints in one single line per day, without providing the exact number of page for each commemoration, so, e.g.: `Maskaram 2: fols 7rc-8ra, commemoration entries for John the Baptist, Dasya and Didemos; salām-hymn to John the Baptist, and Dasya. Maskaram 3: etc..`, or in another synthetic way. If this requires too much space as well (or is too complicated to be generated via a xml->pdf transformation), then we can simply list any difference from the available printed edition.

Gnisci:
Thanks Tea. Like Massimo, I think you should verge on the side of economy, as long as there is an identifiable degree of consistency, not only in terms of your personal time (which is very important), but also to follow the catalogue > bm principle, to which we have agreed to adhere, since I think that most publishers will not want something that is excessively bulky and detailed - that said, it will probably be worth talking about this at our meeting!

PietroLiuzzo:
please note that the detailedness of the output in PDF can be "flattened", i.e. more detailed records do not constitute an issue for that procedure at all. records lacking information which is required for the printed output are an issue, too detailed records are not, and only what is needed for the specific output can be selected on the basis of clearly stated rules.

OxfordBodleian/Juel-Jensen/BDLaethb5.xml
thea-m: Just to ensure that I understood you well: You suggest only giving the beginning of the month and the subscription, not the incipits and explicits of the first and last commemoration?
The other months don't have such an elaborated beginning, so I would go on giving an incipit of the first commemoration there. Also regarding the subscription - I don't know the Second Recensions well enough to attest to their standardisation. But in general, there is more flexibility in these than in the commemorations, so I think it would still be better to give at least one commemoration incipit and explicit per month, to have a basis for a first philological assessment.

thea-m:
Thank you! Massimo's space-saving suggestions for the printed output sound good to me.

OxfordBodleian/Juel-Jensen/BDLaethb5.xml
SusanneHummel: I thought that every month has a kind of prologue, but if not, hm.
However, isn't it odd to provide the beginning of one text and the end of another one? We consider each commemoration notice as an independend text unit. That was my reasoning behind it. But it was just a suggestion. You are the Senkessar expert. I am fine with whatever you think is right. :)

SusanneHummel: If you provide the incipits and explicits of the first and last commemoration notice this would make more sense to me (but as I said, it is entitely up to you as trusted expert)

thea-m: I agree that it is odd! Incipit and explicit of the first and last commemoration seem reasonable to me.

thea-m:
I will close this PR and continue working on the branch, and then open a new PR when it's ready.
Thank you!

PR #648, opened 2020-07-07
added forgotten binding

thea-m:


LondonBritishLibrary/add/BLadd19659.xml
```diff
@@ -87,7 +87,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                     <material key="parchment"/>
                                 </support>
                                 <extent>
-                                    <measure unit="leaf">i + 164 + ii</measure>
+                                    <measure unit="leaf">1+164+2</measure>
                                     <measure unit="leaf" type="blank">2</measure>
```
DenisNosnitsin1970, line 91: I think that i + 164 + ii works out better than 1 + 164 + 2. This second indication fails to show that 1 and 2 refer to the units of another level (to such an extend that they were not foliated). The only problem of  i + 164 + ii is that it does not let the reader understand that numbers i and ii are unwritten. If we assign these numbers in our descriptions only, but they do not exist in physical manuscript, this should be somehow indicated.

thea-m: I am indicating it in the note (especially as this is a record created purely from the catalogue description).
I am a bit confused now, so let me try to paraphrase: You would suggest to use Arabic numerals only for guard leaves that have been foliated, while in the case of unfoliated guard leaves, to use Latin numerals?
In a previous discussion that let to the current guidelines, it was agreed to only use Arabic numerals (which are visualised on the app as Latin numerals).

DenisNosnitsin1970: I have just dealt with a related problem you commented upon in Abb 202. The main thing is as follows. If  Arabic numbres are used for the foliation of the textblock only,  I think we should preferably use a different system to foliate fly leaves. I think it is the common practice.

DenisNosnitsin1970: In addition, if these are numbers added by us (they are not in the catalogue and not in the physical object), it should also be made visible.

thea-m:
I think this should be fine.

PR #652, opened 2020-07-08
Abb202

DenisNosnitsin1970:
Add 202

### DIFF
--- a/ParisBNF/abb/BNFabb202.xml
+++ b/ParisBNF/abb/BNFabb202.xml
@@ -43,7 +43,6 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         </support>
                                 <extent>
                                     <measure unit="leaf">92</measure>
-                                    <measure unit="leaf" type="blank"></measure>
                                     <dimensions type="outer" unit="mm">
                                       <height>160</height>
                                       <width>140</width>
@@ -52,7 +51,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                 <foliation>
                                   <note>
                                   Foliation of the Bibliothèque nationale de France in the upper
-                                  right corner of the recto sides of the folia. There is one unfoliated paper and one parchment leaves
+                                  right corner of the recto sides of the folia. There are one unfoliated paper leaf and one parchment leaf
                                   at the beginning, and one unfoliated paper leaf at the end of the volume.
                                   </note>
                                 </foliation>
@@ -89,11 +88,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                               <persName ref="PRS1071dAbbadi" role="owner">Antoine d'Abbadie</persName>,
                               <ref target="#p2">Unit 2</ref> was added after <ref target="#p2">Unit 1</ref> was bound.
 
-                            Both were united in one binding probably in <placeName ref="INS0303BNF">Bibliothèque nationale de France</placeName>.</origDate>
+                            Both were united in one binding probably in the <placeName ref="INS0303BNF">Bibliothèque nationale de France</placeName>.</origDate>
                                </origin>
                                <provenance>The manuscript was bought by
                                  <persName ref="PRS1071dAbbadi" role="owner">Antoine d'Abbadie</persName>
-                                 sometime between <date>1840</date> <date>1848</date>.
+                                 sometime between <date>1840</date> and <date>1848</date>.
                                </provenance>
                                <acquisition>The manuscript entered the Bibliothèque nationale in <date>1902</date>.
                               </acquisition>
@@ -138,7 +137,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                     <title type="complete" ref="LIT2516Wisdom"/>
                                     <incipit xml:lang="gez"><hi rend="rubric">ጥበበ፡ ሰሎሞን፡ ወልደ፡ ዳዊት፡ ዘነግሠ፡ ለእስራኤል፡ አፈቅርዋ፡ ለጽድቅ፡
                                       መኳን</hi>ንተ᎓ ምድር፨ ወሐልዩ᎓ በእንተ᎓ <hi rend="rubric">እግዚአብሔር᎓</hi></incipit>
-                                       <note>Lacuna occupies half of <locus target="#43vb #43vc"/>.</note>
+                                       <note>A lacuna occupies half of <locus target="#43vb #43vc"/>.</note>
                                     </msItem>
                                     </msItem>
                                   </msContents>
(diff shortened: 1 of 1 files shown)


ParisBNF/abb/BNFabb202.xml
```diff
@@ -0,0 +1,483 @@
...
+
+                        <additions>
+                          <list>
+                            <item xml:id="a1">
+                              <locus target="#1ra"/>
+                               <desc type="Unclear">A short prayer defying Satan.</desc>
+                                 <q xml:lang="gez">በስመ፡ አብ፡ ወወልድ፡ ወመንፈስ፡ ቅዱስ፡ ፩አምላክ። በቅደስት፡ ሥላሴ፡
+                                 እንዘ፡ ኣምን፡ ወእት፡ መኃፀን፡ እከህደከ፡ ሰይጣን፡ በቅድመ፡ ዛቲ፡ እምየ፡ ቅድስት፡ ቤተ፡ ክርስቲያን፡ እንተ፡
+                               ይእቲ፡ ስማዕየ፡ ማርያም። ጽዮን፡ ለዓለመ፡ ዓለም።</q>
+                                   <note>Written in a good hand.</note>
```
thea-m, line 244: Ok, I see that you like `<note>` in additions :) In my understanding, it is really not necessary, because the information given in `<note>` can be give in `<desc>` . But maybe my understading is limited. Is there a specific reason why you prefer to have `<note>` ?

ParisBNF/abb/BNFabb202.xml
```diff
@@ -0,0 +1,483 @@
...
+                                  </item>
+
+                                  <item xml:id="a5">
+                                    <locus target="#2r"/>
+                                     <desc type="PoemQene"/>
+                                       <q xml:lang="gez"></q>
+                                       <q xml:lang="en">For which generation among the number of generations,
+                                       </q>
+                                       <note>Qǝne-poem of the type <foreign xml:lang="gez">mawaddǝs</foreign>,
+                                     of Ṭǝbaba Krǝstos.</note>
```
thea-m, line 276: Is that the name of the author? Then it would be better to create an ID.

ParisBNF/abb/BNFabb202.xml
```diff
@@ -0,0 +1,483 @@
...
+               </xi:include>
+            </encodingDesc>
+            <profileDesc>
+               <textClass>
+                  <keywords>
+                     <term key="ChristianLiterature"/>
+                     <term key="OldTestament"/>
+                     <term key="Apocrypha"/>
+                     <term key="Prayers"/>
+
```
thea-m, line 466: Maybe add Poetry?

ParisBNF/abb/BNFabb202.xml
```diff
@@ -0,0 +1,483 @@
...
+                  </keywords>
+               </textClass>
+               <langUsage><language ident="en">English</language><language ident="ar">Arabic</language><language ident="gez">Gǝʿǝz</language><language ident="am">Amharic</language></langUsage>
+            </profileDesc>
+            <revisionDesc>
+               <change who="DR" when="2018-02-26">Recorded manuscript</change>
```
thea-m, line 472: I think the changes are copy pasted from another record.

ParisBNF/abb/BNFabb202.xml
```diff
@@ -0,0 +1,483 @@
...
+                                    <dimensions type="outer" unit="mm">
+                                      <height>160</height>
+                                      <width>140</width>
+                                    </dimensions>
+                                </extent>
+                                <foliation>
+                                  <note>
+                                  Foliation of the Bibliothèque nationale de France in the upper
+                                  right corner of the recto sides of the folia. There is one unfoliated paper and one parchment leaves
+                                  at the beginning, and one unfoliated paper leaf at the end of the volume.
```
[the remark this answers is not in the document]
DenisNosnitsin1970: I implicate that they are not foliated and therefore not included into the number 92. I see that these additional leaves amount to a kind of small problem of cataloguing. I would tend to take over the number provided in the catalogue, and only add  details in the "note". I would not add numbers of folios that are not in the catalogue and not in the images (in this specific case, I do not really understand well the status of those added paper leaves). Or we need to indicate clearly that those numbers are unwritten and exist in the imagination of the cataloguer only, smth like (ii) - 92 - (i)

ParisBNF/abb/BNFabb202.xml
DenisNosnitsin1970: In "note" I used to place the information that pertain on physical aspect of the note including hands and text arrangement; the information above in <desc> in for me an equivalent of the "title" and explains the content. Not good? :)

ParisBNF/abb/BNFabb202.xml
DenisNosnitsin1970: Yes.

ParisBNF/abb/BNFabb202.xml
DenisNosnitsin1970: Sorry. Thank you for corrections! :)

ParisBNF/abb/BNFabb202.xml
thea-m: Aha! This is a bit idiosyncratic, as it was never discussed, I think. For me, the `@type` of `<desc>` or its first words would serve as the title, while the description would follow continuously. But if you want a more precise structure, we can find a solution for this (I know you are not the only one who wants to use note in additions). For me, there is nothing intuitive for stating that note serves for the physical description. Maybe @PietroLiuzzo has an idea? Do you think that such a structure could be introduced (with note or otherwise)?

PietroLiuzzo: you can do both, I do not think we are introducing anything. the use of note is an additional layer, you can imagine you are adding a footnote, while without it you are continuing the description. I personally see no need to dictate one way or another, of course a consistent use would be better, but we do have already quite a lot of additional rules and this does not seem to me terribly decisive in any context of use for the application.

DenisNosnitsin1970: I do not know. Should it be fixed at all? I now remember that I I use <description> here because it cannot be assigned to any of the fixed additional note typologies, so its quasi-title + content needs some exaplanation. For the rest, I really use "note".

thea-m: It is not fixed now. For me, it would seem easier to read everything in desc (I also think that you don't intend the note to have a footnote like character), than to introduce more elements, but do it as you like.

ParisBNF/abb/BNFabb202.xml
```diff
@@ -268,12 +274,31 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+                                                                         </q>
+                                       <q xml:lang="en">For what generation among the generations,
+                                         For what time among the times,
+                                         Mary, tabot of the only Son of the Father,
+                                         The hand of the prophets were studying you, the salvation,
+                                         Some were exiled from the land, reaching a land they were not accustomed to,
+                                         Examination of the burning fire through its flame and fire,
+                                         Untill it burned them and consumed them like collected wood,
+                                         Them, those who denied you, the tabot of their salvation,
+                                        Illness they caused for themselves though their heads.
```
thea-m, line 297: I would understand it rather as "by themselves"

DenisNosnitsin1970: Ok, I modified. I think "ay" and "maeze" should be understood in this special way (possible for qene).

PR #664, opened 2020-07-14
Create BLorient784.xml

DariaElagina:


LondonBritishLibrary/orient/BLorient784.xml
```diff
@@ -0,0 +1,633 @@
...
+
+                            </handNote>
+
+                        </handDesc>
+
+                        <additions>
+                          <list>
+                            <item xml:id="a1">
+                              <locus target="#12 #209"></locus>
+                              <desc>Amharic deeds of the reign of <persName ref="PRS2377Bakaffa"></persName>,
```
DenisNosnitsin1970, line 522: <desc type="LandGrant"?

DariaElagina: Hmm, I am not sure, there is no specification. I have therefore added type=Record

PR #698, opened 2020-08-07
Corrections to Dabra Maso MY 016 018

eu-genia:
It was quicker to correct directly - there were nesting and numbering problems in MY-018 and some smaller issues (e.g. you forgot to add a change element, please check other files, too)

I personally would split MY-018 in 2 msParts - there is a clear break, there is a colophon after the first part, formerly blank pages with additions and a change of hand and change of layout - to me this is clearly a composite manuscript, even if the second part possibly was never a circulation unit of its own it is still an msPart

If you want I can edit the description accordingly

### DIFF
--- a/ES/ESmy016.xml
+++ b/ES/ESmy016.xml
@@ -28,7 +28,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                <msIdentifier>
                   <repository ref="INS0039MY"/>
                   <collection>Ethio-SPaRe</collection>
-                  <idno>MY-016</idno>
+                  <idno facs="MY/016/MY-016" n="54">MY-016</idno>
+                  
                </msIdentifier>
                <msContents>
                 <summary/>
@@ -141,12 +142,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                  </item>
 
                                  <item xml:id="q6" n="6">
+                                   <note> (the structure of the quire
+                                    cannot be ascertained;
+                                    leaf 1 (<locus target="#38"/>) is possibly single.)</note>
                                    <dim unit="leaf">7</dim>
                                  <locus from="38r" to="44v"/>
-                                 s.l.: 1, no stub
-                                 <note>The structure of the quire
-                                   cannot be ascertained;
-                                   possible leaf 1 (<locus target="#38"/>) is possibly single.</note>
+                                    s.l.: 1, no stub 
                                   </item>
                                  <note><locus from ="38" to="40"/> have been misplaced, they should be
                                  before <locus target="#30"/>.</note>
@@ -331,6 +332,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <change who="EG" when="2014-06-05">last edited</change>
          <change who="PL" when="2016-05-10">transformed from mycore to TEI P5</change>
          <change who="PL" when="2019-04-25">added missing extras from domlib</change>
+         <change when="2020-08-07" who="DN">updated description, adjusted to TEI schema, introduced IDs</change>
+         <change when="2020-08-07" who="ES">minor correction in collation, facs added</change>
       </revisionDesc>
    </teiHeader>
    <text xml:base="https://betamasaheft.eu/">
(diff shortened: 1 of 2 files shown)


ES/ESmy016.xml
```diff
@@ -141,12 +142,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                  </item>
 
                                  <item xml:id="q6" n="6">
+                                   <note> (the structure of the quire
+                                    cannot be ascertained;
+                                    leaf 1 (<locus target="#38"/>) is possibly single.)</note>
```
eu-genia, line 147: I had to move the note up, otherwise the app did not recognize the syntax and did not produce the diagram

ES/ESmy018.xml
```diff
@@ -260,19 +261,23 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       <persName ref="PRS13125ZabanaKrestos">Zabāna Krǝstos</persName>.</note>
     </msItem>
 </msItem>
+                    </msItem>
```
eu-genia, line 264: You forgot to close msItem number 1 here

ES/ESmy018.xml
```diff
@@ -260,19 +261,23 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
-decollation of John the Baptist</title>
-<incipit xml:lang="gez">
-<locus target="#44ra"/><hi rend="rubric">በስመ፡ አብ፡ ወወልድ፡ ወ</hi>መንፈስ፡ ቅዱስ፡ አሐዱ፡ <hi rend="rubric">አምላክ፡
-ዝንቱ፡ ነገር፡ በከ</hi>መ፡ ኮነ፡ ሕይወቱ፡ ለቅዱስ፡ <hi rend="rubric">ዮሐንስ፡ መጥምቅ፡ ወከ</hi>መ፡ እፎ፡
-ኮነ፡ ምትረተ፡ ርእሱ፡ ወተርጐሞ፡ ቅዱስ፡ ዮሐንስ፡ አፈ፡ ወርቅ፡ በትርያርክ፡ ሊቀ፡ ጳጳሳት፡ ብሂል፡ ዘሀገረ፡ ቈስጠንጥንያ፤ ወሶበ፡ ኮነ፡ ስደቱ፡ ውስተ፡ አብራክይስ፡ አመ፡ ተስእሎ፡ አይሁዳዊ፡
-ኬልቄዶናዊ፡ ዘአምነ፡ በእግዚእነ፡ ኢየሱስ፡ ክርስቶስ።
-</incipit>
-
-                     <msItem xml:id="ms_i3">
+<title>Works dedicated to John the Baptist</title>
```
eu-genia, line 267: They can be nested as a collection, or be separate items, but in any case after the msItem 1 was closed. The collection solution fits better as the colophon at the end belongs to the whole set of texts and the colophon must be within an msItem

ES/ESmy016.xml
DenisNosnitsin1970: ok

ES/ESmy018.xml
DenisNosnitsin1970: ok

ES/ESmy018.xml
DenisNosnitsin1970: Ok, we can call it a collection, even if to do this we would need to find at least one more manuscript with the same set of texts.

eu-genia: we don't have to, we are not creating any work ID, we simply nest the texts together  - you must have a wrapping msItem anyway to assign the colophon to the unit

eu-genia: Please approve and merge after reviewing

ES/ESmy016.xml
DenisNosnitsin1970: I was thinking of this. It is a border case because the biggest part of the texts in part 2 is in Hand 2 and the colophon 2 is also in Hand 2 (unless I am wrong assuming this), and the commissioner is the same. The leaf 43 indeed looks like a flyleave. A slight difference in layout is not so decisive if the size of the leaves is the same (cp MY-019). Perhaps the manuscript was a project that dragged on in time, a little bit, and was completed largerly by the same person. Of course you can say that 30 miracles of Mary + John the Baptist collection is an unlikely constellation, and to disprove this I would need to make a research. Well, sure, if you want, you can rearrange it into a composite one.

PR #709, opened 2020-08-12
Create BLadd26553.xml

DariaElagina:
Do you have an idea who this Rich from a1 is?

### DIFF
--- a/LondonBritishLibrary/add/BLadd26553.xml
+++ b/LondonBritishLibrary/add/BLadd26553.xml
@@ -60,7 +60,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         </msItem>
                         <msItem xml:id="ms_i3">
                           <locus from="9v"></locus>
-                          <title type="complete" xml:lang="en">Lord's Prayer</title>
+                          <title type="complete" xml:lang="en" ref="LIT2168PaterN">Lord's Prayer</title>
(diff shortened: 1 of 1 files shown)


LondonBritishLibrary/add/BLadd26553.xml
```diff
@@ -0,0 +1,172 @@
...
+                                <desc>Neatly written</desc>
+                            </handNote>
+
+                        </handDesc>
+
+                        <additions>
+                            <list>
+                                <item xml:id="a1">
+                                    <locus target="#1v"/>
+                                    <desc type="ScribalSignature"></desc>
```
thea-m, line 103: I am not sure from the description whether it is a scribal signature, we do not seem to know whether this note was written by Rich or someone else.
I don't know who he is either.

DariaElagina:
@thea-m , thank you!

LondonBritishLibrary/add/BLadd26553.xml
DenisNosnitsin1970: I would assume that the entire text of this small manuscript might have transcribed by Mr Rich in his early years, 
and he left a note about this.

DariaElagina: I will leave the category Unclear untill we could say more about the character of this note.

thea-m (approved):

PR #722, opened 2020-08-14
Bn fabb152 updated

eu-genia:
restored the ref as suggested


PietroLiuzzo:
I think this change could be simply done on the previous branch before merging it. Now we cannot do much about it. Please let us try in the future to keep things a bit tidy, it does not take much effort, it saves us time for sure. allowing me to merge instead of rushing ahead on your own PR, would have been another solution, as I was writing a comment on this while you proceeded.

PR #730, opened 2020-08-17
desc around titles

eu-genia:
@DenisNosnitsin1970 
I also updated RNBefns5 and IVorlov11 where some instances remained

note also in RNBefns5 : the English translation of a quotation should go inside of a `<foreign>` element nested inside of `<q>` (we have no English texts in the mss, so no `<q xml:lang="en">` should be there


DenisNosnitsin1970:
Thanks!

PR #754, opened 2020-09-07
Lessana Sab' MSS

smaugustine:
Add BAVet 208

### DIFF
--- a/VaticanBAV/et/BAVet116.xml
+++ b/VaticanBAV/et/BAVet116.xml
@@ -1,0 +1,301 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="BAVet116" xml:lang="en" type="mss">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>Missal</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="AD"/>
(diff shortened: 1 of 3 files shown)


smaugustine:
I just noticed that some of the commits got mixed up so that it includes 116 and 117 (this pull request is only to add 208). I will see what I can do about this.

PietroLiuzzo:
in theory if you create your new branch from master, you should not have those changes which are part of vatican-manuscript branch, or is that not so?

smaugustine:
@PietroLiuzzo That's what I expected, but that doesn't seem to be the case in practice. I even tried to make a new branch from master using VSCode instead of Atom -- specifying clearly that it was new from master -- but the commits from the open pull request on the other branch persist.

I've now tried three times and each branch ends up with at least three files changed (116, 117, and 208).

PietroLiuzzo:
what I do usually is to commit what is uncommitted to the branch where I am working, switch back to master, then make a branch from there.  I use Github Desktop, which asks you and warns you if for example you have uncommitted changes which will then have to be brought to the new branch not to get lost. I do not know VSCode, sorry!

eu-genia:
Augustine - when you have multiple branches and you do not want to move changes around (indeed, what you describe is the normal behaviour) you should always go via master branch.
that is - you have the 116-117 branch, you have committed all changes there. even if they are not yet merged yet - if you are not now continuing your work on them you have to go to master, and create the 208 branch from there, then it would have only 208 in it

eu-genia:
(when you switch to master you should of course pull from master, then your local changes in 116 117 would be overwritten, so they will "disappear")

smaugustine:
The issue, it turned out, was that there were some commits that had crept in due to Atom messing up pull requests from master (sounds odd, I know, but something strange happened), and so the commits were actually from master, not from another branch, which is why they were copied over. In the end I simply deleted my local copy and did a fresh clone to remove the commits and make a clean branch from master.

smaugustine:
Suffice to say, Atom is not my Git app of choice (nor even my editor of choice, but I still haven't succeeded in adding TEI linting/validating to VS Code). Giving in and using GitHub Desktop has helped avoid these issues (though, of course, GitHub also makes Atom...).

PR #785, opened 2020-09-18
Update EMML4398.xml

PietroLiuzzo:
not urgent at all, for later on, after summer school and everything else.
in relation to the third item in [this list](https://github.com/BetaMasaheft/Documentation/issues/1472), with the newly trained model in transkribus I have followed the steps newly documented[ in the guidelines](https://betamasaheft.eu/Guidelines/?id=transcription-transkribus) and added a part of trans_k_ription  to test the impact of this additional edition as well as the encoded text, etc. image has been edited for segmentation, passed to HTR+ and then downloaded as TEI from Transkribus, then transformed with transkribus2BM.xsl and only the relevant part (facsimile 2 and 2verso and 3recto) have been copied into the record. Of course this numbering is entirely ignorant of the real location in the mss of the leaves in this image.


thea-m:
I think that this can be closed after the amazing https://github.com/BetaMasaheft/Manuscripts/pull/834/files, right?

PR #790, opened 2020-09-22
Update EMIP00480.xml

PietroLiuzzo:
fixed targets in locus with comma instead of #


PietroLiuzzo:
this error in the encoding is what prevent the manifest from generating correctly. thanks for spotting this.

smaugustine (approved):

PR #792, opened 2020-09-26
Update EMIP01859.xml

PietroLiuzzo:
@thea-m this is the transcription I run yesterday. hope it is helpful!

if you merge, it is quite possible it will not go through. but I should see and upload manually. thanks!


thea-m:
Thank you! It is very helpful. After f. 55, only the ra and vb of each leaf are present. Before f. 55, the missing parts are less regular. I added them all commented out in the file. In transkribus, everything is apparently present, so this loss might have happened during the export?

thea-m:
I think if this is merged, I can add the missing parts from transkribus and make a new PR

PietroLiuzzo:
This will now go if merged, but it is too old and I cannot resolve the conflict. I will close it and make a new pr

PR #803, opened 2020-10-07
Vat Borg et 9

smaugustine:


VaticanBAV/borget/BAVborget9.xml
```diff
@@ -0,0 +1,190 @@
...
+
+                    <msItem xml:id="ms_i3">
+                      <locus from="44v" to="53r"/>
+                      <title ref="LIT2444Temher"/>
+                      <textLang mainLang="gez"/>
+                    </msItem>
+
+                    <msItem xml:id="ms_i4">
+                      <locus from="53v" to="64v"/>
+                      <title ref="LIT1438GadlaE">Ten Miracles of St ʾEwosṭātewos with hymns.</title>
```
eu-genia, line 64: not LIT4315Miracles?

VaticanBAV/borget/BAVborget9.xml
```diff
@@ -72,8 +72,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
 
                       <msItem xml:id="ms_i4.2">
                         <locus from="54v" to="55r"/>
-                        <title>First Miracle</title>
-                        <!-- Cf. LIT3053RepCh333 -->
+                        <title>First Miracle</title><!-- Cf. LIT3053RepCh333 -->
```
eu-genia, line 75: it might be helpful to introduce `<div>`s with derived IDs and labels + incipits in the work record (which I assume is LIT4315Miracles), with incipits, and then here the `@ref` can be added to the `<title>`

VaticanBAV/borget/BAVborget9.xml
```diff
@@ -266,7 +266,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         <item xml:id="a3">
                           <locus target="#3v"/>
                           <desc type="OwnershipNote">
-                            A note, part of which is lost, indicating that the manuscript belonged to Gabra Giyorgis.
+                            A note, part of which is lost, indicating that the manuscript belonged to <persName role="scribe" ref="PRS13179Gabra"/>.
```
eu-genia, line 269: if the MS belonged to him then his role should be "owner"

eu-genia (changes requested):

VaticanBAV/borget/BAVborget9.xml
eu-genia: Check also LIT3200DossierEw as the ID for combination of various hagiographic works dedicated to Ewostatewos (Gadl, Miracles, Hymns, in any constellation)

smaugustine: It is a different set of miracles than those included in LIT4315Miracles. According to a comment in LIT4315Miracles, Turaiev's edition included 16, which also seems to be a different recension.

The catalogue here notes a similarity with the fragment BL Or 829*, but it is only the incipit, it otherwise seems to be a different text.

I haven't been able to get access to Lusini's article on the manuscript tradition, and I'm not sure that there are work records for all the various recensions. This, in any case, seems different than is typical.

eu-genia: We have at least three records for the various recensions 4319 alpha 4320 beta 4321 gamma
Lusini's article I think has been superseded by his PhD monograph

eu-genia: As to the number of miracles - this varies, I had MSS with 11 in EthioSPaRe etc, LIT4315Miracles is a general label, we have not gotten to identify single miracles

smaugustine: Looking more closely at the text in the images, it appears that it is a versified version of the Miracles, which explains why they are all so short.

smaugustine: It would seem best not to follow Grébaut's division of the text here and instead to use the divisions of Chaîne, listing these as 1.4.1 (መንክራቲከ፡ ሰማዕነ፡), 1.4.2 (RepCh 323), 1.4.3 (RepCh 333), and 1.4.4 (RepCh 353).

eu-genia: Sounds good. Can you check the refs in BAVet46? Don't look right to me.
For completeness I have pushed to the same branch an E-S ms with the miracles (facs added, to see the images)

VaticanBAV/borget/BAVborget9.xml
```diff
@@ -226,7 +226,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         <seg type="rubrication">Titles, nomina sacra, and the first two letters of each stanza</seg>
```
eu-genia, line 226: just noticed - we do not use "nomina sacra" - which has a different meaning (abbreviated names) - we write "holy names" instead

eu-genia (changes requested):

VaticanBAV/borget/BAVborget9.xml
smaugustine: An amazing coincidence, @Ralph-Lee-UK just sent me images for EMIP 764 and it also contains (84r-86r) this same additional stanza at the beginning ("መንክራቲከ፡ ሰማዕነ፡ እንዘ፡ ትነብር፡ አርማንያ፡ መካነ።") followed by Chaîne 323, 333, and 353 and then some additional stanzas.

eu-genia: Since this seems to be a frequent constellation we might mention it in the work record for miracles, or consider creating a new overarching work record for the collection miracles+hymns

smaugustine: I was just thinking about this, a work record for "Versified Miracles of ʾEwosṭātewos" or the like, with textparts for each of the constituent parts.

BL Or 641 titles it "ኡደተ፡ ባሕር፡", but the MS only contains RepCh 353.

smaugustine: Or maybe better, መንክራቲከ፡ + 323 + 333 and then separately 353 (ዑደተ፡ ባሕር፡). In most manuscripts, while these are all copied together, there is a clear separation between 353 and what precedes it (no clear separation with the other parts).

eu-genia (approved):

PR #808, opened 2020-10-11
Update BNFet116.xml

DenisNosnitsin1970:
Correction, introduction of new ID

### DIFF
--- a/ParisBNF/et/BNFet116.xml
+++ b/ParisBNF/et/BNFet116.xml
@@ -253,7 +253,11 @@ schematypens="http://relaxng.org/ns/structure/1.0"
 
                   <handDesc>
                      <handNote xml:id="h1" script="Ethiopic">
-                        <date notBefore="1700" notAfter="1799"/>
+                        <date notBefore="1700" notAfter="1799" resp="DN"/>
+                        <desc>Contrary to the initial estimation (16th century) in <bibl>
+                           <ptr target="bm:Zotenberg1877Catalogue"/>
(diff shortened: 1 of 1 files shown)


eu-genia:
Since @thea-m is the one who created the original description I think she should be involved as the reviewer. For me all looks OK, but I think this should be the correct procedure.

eu-genia (approved):

ParisBNF/et/BNFet116.xml
```diff
@@ -193,61 +244,61 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+
                      </list>
                   </additions>
                </physDesc>
-            
+
                <history>
                   <origin>
-                     <origDate notBefore="1500" notAfter="1599" evidence="lettering"/>
+                     <origDate notBefore="1700" notAfter="1799" evidence="lettering"/>
```
thea-m, line 298: Maybe, in the case of such stark change from the catalogue estimation, it would be good to add a `@resp` to the date, or possible even a short sentence? I think users would not be able to tell whether it's the catalogue's or your updated date now.

thea-m (changes requested):
Thank you, just a general remark on the date

ParisBNF/et/BNFet116.xml
eu-genia: @DenisNosnitsin1970 I think @thea-m is waiting for this small update before she approves and we can close this one

eu-genia (approved):

thea-m (approved):

PR #815, opened 2020-10-13
Max07

MassimoVilla:
I added some remarks concerning this composite manuscript, which I also give here:
- Data concerning the text area of p1 is missing. This has been added to the REQUEST FOR FURTHER ACTION list.
- Information on leafdim2 (leaf dimensions of p2) is missing. The leaf size of p2 looks by eye extremely similar to that of p1, so those data might be extended to p2 too, yet I am inclined to add that task to the REQUEST FOR FURTHER ACTION list.
- Information on the weight of the manuscript is currently encoded within p1. Yet, this is misleading since that piece of information obviously refers to the entire manuscript. So, it should be encoded elsewhere, preferably at the beginning of the description. I think this is a general problem concerning the BM encoding practice and needs a consistent solution.
-  The manuscript is kept in a box, in an embroidered rectangular piece of textile, measuring 620 x 920 mm. How to encode this information (currently given under `<condition/>` of p1)? A quick survey of the guidelines has not been helpful.

Thank you for any suggestion!

### DIFF
--- a/OxfordBodleian/Juel-Jensen/BDLaethc14.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethc14.xml
@@ -38,203 +38,495 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         </altIdentifier>
                     </msIdentifier>
                     
-                    <msContents>
-                        <summary/>
+                    <physDesc>
                         
-                        <msItem xml:id="ms_i1">
-                            <locus from="1ra" to="130va"/>
-                            <title type="complete" ref="LIT1560Gospel"/>
-                            <note>The introductory materials to the Four Gospels, i.e. the Synopis of the class, the Letter of Eusebius to Carpianus and the Eusebian Canons, are missing.</note>
-                            
-                            <msItem xml:id="ms_i1.1">
-                                <locus from="1ra" to="35ra"/>
-                                <title type="complete" ref="LIT1558Matthew"></title>
-                                <msItem xml:id="ms_i1.1.1">
-                                    <locus from="1ra" to="1va"/>
-                                    <title type="complete" ref="LIT1558Matthew#TituliMatthew"/>
-                                    <textLang mainLang="gez"/>
-                                    <incipit xml:lang="gez">
-                                        <locus target="#1ra"/>
-                                        <lb/><hi rend="rubric">፩</hi> በእንተ፡ ቅዳሜ፡ ትምህርት፡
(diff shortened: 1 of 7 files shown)


thea-m:
Just very quickly without having really read the record, regarding the question on the weight: All information pertaining to the entire manuscript (especially physical description and history) can be added before the two msParts. In fact, there is already a `<physDesc>` at the beginning of the description, which contains the binding description. Here, any relevant elements can be included. These can then be either deleted from the `<physDesc>`s of each msPart, or be modified there as to contain only information on the respective msPart. I hope that helps!
Maybe @ElianaDalSasso has suggestions on the encoding of the box and textile? (The same is true here as well: as this pertains to the entire manuscript, it has to be moved from p1 to the beginning of the description, I assume in the physDesc there.)

thea-m:
I'll close this PR, then you can continue working on the same branch and open a new PR when you're ready.

Gnisci:
Just to get back to two of your questions:

a) box this is a modern box in which all the manuscripts in the Bodleian are stored. I don't think it's worth describing them. We can discuss about whether we want to mention in the introduction that all the mss are in modern boxes (does not seem necessary to me).

b) The textile is a souvenir that JJ purchased in Ethiopia to use to wrap the manuscript, he also does this with another one where he uses a flag. If we describe it, I would not say much more than the "the manuscript is wrapped in a modern textile with two lions that was probably purchased by Juel-Jensen."

OxfordBodleian/Juel-Jensen/BDLaethc14.xml
```diff
@@ -34,6 +38,590 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                                stub between <locus target="#113"/> and <locus target="#114"/>
+                                            </item>
+                                            <item xml:id="q18" n="18">
+                                                <dim unit="leaf">9</dim>
+                                                <locus from="115r" to="123v"/>
+                                            </item>
+                                        </list>
+                                    </collation>
+                                    <condition key="deficient">
+                                        The manuscript is kept in a box, in an embroidered rectangular piece of textile, measuring 620 x 920 mm. <!-- how to encode this information? -->
```
ElianaDalSasso, line 404: I think that the embroidered textile could be considered as an enclosure according to Ligatus definition https://www.ligatus.org.uk/lob/concept/3311, and added to binding description. However, this has never been encoded in BM. 
As regards the box, I think it is useful to mention it, in order to give information on the preservation state of the manuscript.

PR #821, opened 2020-10-16
Bav borg et 22

smaugustine:
(One of the MSS used by Guidi for his edition of ገድለ፡ አረጋዊ፡ and by Lifchitz for ሐፁረ፡ መስቀል፡)


VaticanBAV/borget/BAVborget22.xml
```diff
@@ -310,14 +313,31 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                      <seg type="rubrication">
+                        Titles, the first one or two lines at the beginning of a section, holy names
+                      </seg>
+                    </handNote>
+                    <handNote xml:id="h2" script="Ethiopic" corresp="#ms_i5 #ms_i6 #ms_i7 #ms_i8 #ms_i9">
+                      <seg type="rubrication">
+                        Titles, the first one or two lines at the beginning of a section, holy names
+                      </seg>
+                    </handNote>
+                    <handNote xml:id="h3" script="Ethiopic" corresp="#ms_i10 #ms_i11">
```
eu-genia, line 326: we have a new text, new layout, and new hand beginning with ms_i10 - sure no unit break? (probably not, Grebaut usually comments on that, but just to be sure...)

eu-genia (commented):
I now notice that there is a listBibl type="editions" (line 129 + 150) - this type is reserved for work records, if the bibliography must be quoted here use "secondary"

No title at all for msItem 10 and 11?

VaticanBAV/borget/BAVborget22.xml
smaugustine: That's a very good point, especially considering the difference in topic. Grébaut doesn't discuss it, and the quire structure is perhaps an issue. The thing would be this item spanning ff. 87v-88v, which is the end of one quire and the beginning of a next. In this case, it is "three fragments" of a magical prayer, which doesn't create so much of an issue, but without images or a clearer description it's hard to say for certain.

smaugustine:
@eu-genia Thank you for catching that, I meant to go back and sort out the issue of 10 and 11. While Grébaut considered these two works, the incipit for 10 is that of LIT1637Homily and the incipit for 11 is rather the explicit for LIT1637Homily. What follows after the "incipit" for 11 up to the explicit is presumably a miracle of Michael or the like, as is usually found in ድርሳነ፡ ሚካኤል፡, but it's hard to say for certain without seeing the manuscript. I've just looked through the printed book, but I don't see anything corresponding. Grébaut was evidently confused by the various subdivisions of these homilies, as he notes for 11 "In festum s. Michaelis fragmentum sine titulo".

eu-genia (approved):

PR #826, opened 2020-10-20
BAV et 173, 190, 196

smaugustine:
Grébaut is fairly light on the description about these MSS, sometimes giving no info about hands or the number of columns &c.


VaticanBAV/et/BAVet190.xml
```diff
@@ -27,6 +27,26 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
                     </msIdentifier>
+
+                    <msContents>
+
+                      <msItem xml:id="ms_i1">
+                        <locus from="2r" to="7v" />
+                        <title type="incomplete" ref="LIT2915RepCh196" />
+                        <textLang mainLang="gez"/>
+                        <explicit xml:lang="gez">ወሰላም፡ ለመቃብሩ፨ ገብረ፡ መንፈስ፡ ቅዱስ፡ አቡየ፡ ለእግዚአብሔር፡ ፍቁሩ፨ ደመናት፡ አንኰርኰሩ፨ አዕዋፈ፡ ሰማይ፡ ዘመሩ፨ በጊዜ፡ ሞትከ፡ መላእክት፡ አንከሩ፨ ኦአምላከ፡ ገብረ፡ መንፈስ፡ <gap reason="ellipsis" extent="unknown" resp="PRS4805Grebaut PRS9530Tisseran"/></explicit>
+                        <note>Incomplete at the end.</note>
```
eu-genia, line 38: is it incomplete because it was copied thus, or because of material loss? we never finished the discussion https://github.com/BetaMasaheft/Documentation/issues/1161 as to whether the two types of incompleteness should be kept apart in the encoding, for the moment we do not, but it might be helpful if the information is at least provided as free text/note when possible

VaticanBAV/et/BAVet196.xml
```diff
@@ -163,6 +164,20 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                           </layoutDesc>
                         </objectDesc>
 
+                        <additions>
+                          <list>
+                            <item xml:id="a1">
+                              <locus target="#1v" />
+                              <desc type="Supplication">
+                                <title>ʾAsmāt Prayer</title>
```
eu-genia, line 181: there is a generic ID for ʾAsmāt-texts that could be added in a `@ref`

VaticanBAV/et/BAVet196.xml
```diff
@@ -39,49 +68,49 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
-                          <textLang mainLang="gez"/>
+                          <textLang mainLang="gez" />
                         </msItem>
 
                         <msItem xml:id="p1_i4">
                           <locus from="52v" to="54r" />
-                          <title>Prayer for</title>
-                          <textLang mainLang="gez"/>
-                          <incipit xml:lang="gez">በስመ፡ አብ፡ ... ግልል፡ ዝልል፡ ምልል፡ ምዝርር፡ ዘልዑለ፡ ይሰርር፡ ወልዑለ፡ ይትናገር፡ አስናብር፡ ውስተ፡ ልብየ፡ ቃለ፡ ትአብከ፡ ለአንብቦ፡ መጻሕፍት፡ ወለፈክሮ፡ ቃላት፡</incipit>
+                          <title>Prayer for Reading and Interpreting Books</title>
```
eu-genia, line 89: no IDs here and below?

eu-genia (changes requested):

VaticanBAV/et/BAVet190.xml
smaugustine: The catalogue is quite lacking in this regard. It doesn't say anything explicit regarding lost text or give any information that would clarify it, such as differences in hands. That the quires and units don't correspond complicates the matter, unless the quires have folios stitched in, but the catalogue never includes such information.

VaticanBAV/et/BAVet196.xml
thea-m: How has this been resolved? It would also be possible to use `@type="Asmat"` and then mark up the quotated text with `<seg type="supplication">`

VaticanBAV/et/BAVet196.xml
```diff
@@ -0,0 +1,308 @@
...
+
+                        <msItem xml:id="p2_i1">
+                          <locus from="55r" to="77v" />
+                          <title type="complete" ref="LIT2866RepCh147" />
+                          <textLang mainLang="gez" />
+                        </msItem>
+
+                        <msItem xml:id="p2_i2">
+                          <locus from="78r" to="80r" />
+                          <title>ʾAsmāt Prayer</title>
```
thea-m, line 214: it would be possible to either use the general ID LIT5888AsmatPrayer or create a new one

VaticanBAV/et/BAVet196.xml
```diff
@@ -0,0 +1,308 @@
...
+                          <textLang mainLang="gez" />
+                          <incipit xml:lang="gez"><gap reason="ellipsis" extent="unknown" resp="PRS4805Grebaut PRS9530Tisseran" /> እስር፡ ጕርዔሆሙ፡ ለፀርየ፡ <gap reason="ellipsis" extent="unknown" resp="PRS4805Grebaut PRS9530Tisseran" /> በስመ፡ አብ፡ በል፡ ጸለየ፡ ምስለ፡ ዳዊት፡ ሳዶር፡ አላዶር፡ ዳናት፡ አዴራ፡ ሮዳስ፡</incipit>
+                          <note>Written in pencil.</note>
+                        </msItem>
+
+                        <msItem xml:id="p2_i3">
+                          <locus target="#80v" />
+                          <title/>
```
thea-m, line 222: I take that no more information is given in the catalogue? Maybe you could open a text identification issue.

VaticanBAV/et/BAVet196.xml
```diff
@@ -0,0 +1,308 @@
...
+                      </msContents>
+
+                      <physDesc>
+                        <objectDesc form="Codex">
+                          <supportDesc>
+                            <support>
+                              <material key="parchment" />
+                            </support>
+                            <extent>
+                              <measure unit="leaf">26</measure>
```
thea-m, line 235: Could you add here and in the first msPart (if it is known) a locus with the extent of the msPart in the manuscript?

VaticanBAV/et/BAVet196.xml
```diff
@@ -0,0 +1,308 @@
...
+                            <material key="parchment" />
+                          </support>
+                          <extent>
+                            <measure unit="leaf">80</measure>
+                          </extent>
+                        </supportDesc>
+                      </objectDesc>
+
+                      <handDesc>
+                        <handNote xml:id="h1" script="Ethiopic">
```
thea-m, line 46: I assume that the catalogue does not state why the manuscript is composed of two units, but written by one hand?

VaticanBAV/et/BAVet196.xml
```diff
@@ -0,0 +1,308 @@
...
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <particDesc>
+               <listPerson>
+                 <person>
+                    <persName ref="PRS13202Gabra" role="owner"/><!-- ff. 13v 24v -->
```
thea-m, line 278: The information on the persons and the folia on which they are attested would be best encoded as an extra in the respective msPart

thea-m (changes requested):
Thank you!

VaticanBAV/et/BAVet173.xml
```diff
@@ -0,0 +1,291 @@
...
+                        <repository ref="INS0003BAV"/>
+                        <collection>Aethiopici</collection>
+                        <idno>Aeth. 173</idno>
+                    </msIdentifier>
+
+                    <msContents>
+
+                      <msItem xml:id="ms_i1">
+                        <locus from="3r" to="12r" />
+                        <title type="complete" ref="LIT2835RepCh119" />
```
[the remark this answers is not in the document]
smaugustine: It that a problem? It's a habit from web development.

I could be wrong, but I believe whitespace (or at least this sort of whitespace) has no effect on the processing of XML (for sure it has no effect in HTML, which is based on the specifications for XML).

VaticanBAV/et/BAVet196.xml
smaugustine: Ahh, sorry, I meant to back to it. Grébaut notes "A historical note, it seems, regarding the time of some calamity"

VaticanBAV/et/BAVet196.xml
smaugustine: Actually, there is no notes at all regarding how many hands there are, only about the rubrication. Is there a better way to include information about the rubrication?

eu-genia: if this is a true composite MS with 2 distinct parts it is more probable that there are also 2 hands, even if the cataloguers don't specifically mention it... if we take it for granted that we have two units I would have the handDesc with the parts (one each), and possibly add a note with the second handDesc that we do not know if h2 is the same as h1

eu-genia (approved):

VaticanBAV/et/BAVet173.xml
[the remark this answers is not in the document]
thea-m: Good, I remember that this used to be a problem, but it might not be so anymore. @PietroLiuzzo ?

eu-genia:
I see that this old PR is still open I guess waiting for @thea-m ?

smaugustine:
I guess because the discussion about whitespace in self-closing tags was left a bit open.

VaticanBAV/et/BAVet173.xml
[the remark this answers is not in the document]
PietroLiuzzo: these spaces are irrelevant, they can stay where they are and they do no harm at all as already stated by @smaugustine . I do not recall these spaces causing any issue before, although many spaces did and still do, for example inside attributes values etc.

PietroLiuzzo:
Thanks for pointing to the open discussion, I missed that ping completely!

thea-m (approved):

PR #831, opened 2020-10-21
Update BLorient791.xml

DariaElagina:


thea-m (approved):
I think the date of the previous manuscript was more precise, because a ruler was mentioned? If they were really produced in one act of production, it could be useful to indicate this narrower date as a possibility in history.
Similarily, this might be a worthwile example to try out the Syntaxe du codex relations to connect both manuscripts.

DariaElagina:
Concerning the Syntaxe du codex, it is a good point, but I would need sometime to find the best solution.

thea-m (approved):

PR #836, opened 2020-10-27
Multiple mss

DenisNosnitsin1970:
Corrections,  LIT5069Prayer and LIT6205Saalnaka


DayrSuryan/DSEthiop9.xml
```diff
@@ -58,9 +58,9 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         <textLang mainLang="gez"/>
 
                      <incipit xml:lang="gez" type="inscriptio">
-                         <hi rend="rubric">በስመ፡ አብ ወወልድ፡ ወመንፈ<sic resp="DR">ን</sic>ስ፡ ቅዱስ፡</hi>
+                         <hi rend="rubric">በስመ፡ አብ ወወልድ፡ <sic resp="DR">ወመንፈንስ፡</sic> ቅዱስ፡</hi>
```
thea-m, line 61: I think that this change has been carried out to change the visualisation. @PietroLiuzzo the fact that `<sic>` is not yet rendered as (!) after the word (not the letter) does not mean that it is possible, right?

thea-m (changes requested):

DenisNosnitsin1970:
Can this be merged?

DayrSuryan/DSEthiop9.xml
thea-m: @PietroLiuzzo could you confirm (or correct me)?

thea-m:
I still think that changing the exact markup of sic is not necessary, let's wait for @PietroLiuzzo 's answer

DenisNosnitsin1970:
Dorothea, let us merge it?

thea-m (approved):

PR #838, opened 2020-10-30
added facsimile element with link

eu-genia:


thea-m (approved):
How did you find them? I always filtered http://www.bl.uk/manuscripts/Default.aspx with "ethiopic", but these are not among the results. Is the only way to be sure to try out each shelfmark? :(

eu-genia:
I got the info from Ilana, I will ask her for a complete list :)

PR #848, opened 2020-11-06
Bl or769 mk

MarcinKrawczuk:


LondonBritishLibrary/orient/BLorient769.xml
```diff
@@ -30,20 +30,34 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
                         </altIdentifier>
-                        
+
                     </msIdentifier>
                     <msContents>
-                        <summary/>
-                        
+<msItem xml:id="ms_i1">
+<locus from="3r"/>
+<title xml:lang="gez" type="complete" ref="LIT4336Ledata">Genealogy of the monks from Antony down to Philip and Jacob in
```
thea-m, line 40: As I noted elsewhere (of course, this is the older PR), I believe it is better to add the discoursive information to the work files than the titles in the manuscript records. This will contribute to a better reading experience (my opinion, it is not wrong to have them here)

thea-m (approved):
Thank you!

PR #849, opened 2020-11-06
D sminorcorrections

DenisNosnitsin1970:
Small corrections MS DS 5-10

### DIFF
--- a/DayrSuryan/DSEthiop7.xml
+++ b/DayrSuryan/DSEthiop7.xml
@@ -294,7 +294,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                      <supportDesc>
                         <support>
                            <material key="paper"/>
-                           <p>The manuscript is a re-utilized industrially produced Italian notebook.
+                           <p>The manuscript is a reutilised industrially produced Italian notebook.
                               For writing, the notebook has been turned upside down.
                               The title <foreign xml:lang="it">“Quaderno”</foreign> now appears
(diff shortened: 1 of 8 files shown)


DayrSuryan/DSEthiop7.xml
```diff
@@ -286,12 +294,10 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                      <supportDesc>
                         <support>
                            <material key="paper"/>
-                           <p>The manuscript is made of an industrially produced Italian notebook
-                              (the title <foreign xml:lang="it">“Quaderno”</foreign>
-                              appears at the end, on <locus target="#iiv"/>).
-                              The notebook has been turned around for writing,
-                              the title <foreign xml:lang="it">“Quaderno”</foreign> now appears
-                              upside down. </p>
+                           <p>The manuscript is a re-utilized industrially produced Italian notebook.
```
thea-m, line 297: reutilised?

thea-m (changes requested):
orthographic question

DayrSuryan/DSEthiop7.xml
DenisNosnitsin1970: ok. strange, New Oxford Style has re-embark, re-emerge, re-equip, re-establish. Perhaps we consider reutilise not just "utilise the second time", but "reconsider as smth new".

thea-m: it might be right. but let's keep the s, re-utilised

thea-m (approved):

PR #850, opened 2020-11-06
DS additions

thea-m:
added Arabic
@PietroLiuzzo I'm still not sure about the encoding of the translated decorations descriptions (in DSEthiop1, 4 and 8). I left them commented out for now because the schema complains about a missing xml:id.


thea-m:
@DenisNosnitsin1970, I'm not sure how to solve the conflict in DSEthiop3. The change is having "the" `<origDate>the seventeenth century</origDate>`, which indeed I approved of in the last PR but wonder why now. Do we want "the seventeenth century" in the title line?

DayrSuryan/DSEthiop1.xml
```diff
@@ -561,6 +572,12 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         <note>The legend is written in a crude hand, the vowel marker of <foreign xml:lang="gez">ዝ</foreign> is placed wrongly
                         (at the right leg of the letter, and turned to the right).</note>
                      </decoNote>
+                        
+                  <!--   <decoNote xml:id="drawing" corresp="#d3" xml:lang="ar">
```
PietroLiuzzo, line 576: the decoration is the same, you need 2 desc one for each xml:lang, but the decoNote should be one, I think

DenisNosnitsin1970 (approved):

DayrSuryan/DSEthiop1.xml
thea-m: Same for the locus? For example:
```
    <locus target="#118v"/>
    <locus target="#118v" xml:lang="ar">ورقة ١١٨ ظ</locus>
```

DenisNosnitsin1970: Is it for the Arabic inventory?

thea-m: exactly

PR #867, opened 2020-11-18
Emip metadata 11 november 2020

Ralph-Lee-UK:
Standard EMIP edits


PietroLiuzzo:
it looks like files have been committed with conflicts highlighted by github, simply ignoring them. I am fixing and will commit the necessary changes. please note that Atom points these out with an easy to use button-like interface, so that it should be easy to avoid this sort of things.

PietroLiuzzo:
also, please, note that "Minor Changes" does not mean anything, is only a self qualitative/quantitative evaluation of the work done. inside a change element if you are to write anything, please write what sort of minor changes you did, e.g. moved commas, added white space, indented, etc. if you consider your changes so "minor" you can also avoid adding a change element. same applies for commits descriptions to github, try to say what you did, so that the record we keep in that way helps. please refer to the guidelines https://betamasaheft.eu/Guidelines/?id=revisions

PietroLiuzzo (approved):
validation only

Ralph-Lee-UK:
@Pietro noted

PR #874, opened 2020-11-21
B lorient690

MarcinKrawczuk:


thea-m (changes requested):
Thank you! some remarks on the encoding

thea-m (approved):

PR #878, opened 2020-11-23
Corrections after PR and to old files

thea-m:


OxfordBodleian/Juel-Jensen/BDLaethf29.xml
```diff
@@ -182,8 +182,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                         <add place="overstrike">ገብረ፡</add>
                                     </subst>
                                     <hi rend="rubric">ማርያም፡</hi></persName>ወለጸሐፊሁ፡ <hi rend="rubric"
-                                    >ገብረ፡ <del rend="erasure" unit="chars" extent="4"><certainty
-                                            locus="value" match="../@extent" assertedValue="5"
+                                    >ገብረ፡ <del rend="erasure" unit="chars" quantity="4"><certainty
```
SusanneHummel, line 185: In this case, I did not change the attribute extent to quantity since here I am not sure about the exact number of erased characters. Here we have a range of 4-5 characters. 
I understood that quantity is only valid in cases when we know the exact number of lost characters.
In any case, I  think I am also currently updating the file (I can check only the next time I will be in the office). Would it be okay that I myself would include those changes (if we change it) together with my other updates? I might be in the office on Wendesday, so I know wehther or not I committed already some changes to this file.
Thank you!!

SusanneHummel (changes requested):

OxfordBodleian/Juel-Jensen/BDLaethd20.xml
```diff
@@ -2781,12 +2782,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
                                     <locus from="88va" to="88vb"/>
-                                    ወሀለወት፡ አሐቲ፡ ብእሲት፡ በደወለ፡ ቃህራ፡ ወወለደት፡ ፯ውሉደ፡ ወሞቱ፡ ኵሎሙ<hi rend="rubric">፨</hi>
+                                    ወሀለወት፡ አሐቲ፡ ብእሲት፡ በደወለ፡ ቃህራ፡ ወወለደት፡ ፯ ውሉደ፡ ወሞቱ፡ ኵሎሙ<hi rend="rubric">፨</hi>
                                     ወእምዝ፡ ሰአለት፡ ወበከየት፡ ኀበ፡ እግዝእትነ፡ ቅድስት፡ <cb n="b"/> ድንግል፡ 
                                     በ<hi rend="rubric">፪</hi> <hi rend="rubric">ማርያም፡</hi> ወላዲተ፡ አምላክ፡
                                     ከመ<supplied resp="DR" reason="omitted">፡</supplied> ተሀባ፡ ውሉደ<hi rend="rubric">፨</hi>
-                                    ወአስተርአየታ፡ እግዝእትነ፡ ቅድስት፡ ድንግል፡ በሕልም፡ ወምስሌሃ፡ <hi rend="rubric">፫</hi>ደቂቅ፡
-                                    ወ<hi rend="rubric">፩</hi>በአምሳለ፡ ወሬዛ፡ ወ<hi rend="rubric">፪</hi>በአምላለ፡ ሕፃናት<hi rend="rubric">፨</hi>
+                                    ወአስተርአየታ፡ እግዝእትነ፡ ቅድስት፡ ድንግል፡ በሕልም፡ ወምስሌሃ፡ <hi rend="rubric">፫</hi> ደቂቅ፡
+                                    ወ<hi rend="rubric">፩</hi> በአምሳለ፡ ወሬዛ፡ ወ<hi rend="rubric">፪</hi> በአምላለ፡ ሕፃናት<hi rend="rubric">፨</hi>
```
MassimoVilla, line 2790: በአምሳለ፡ instead of በአምላለ፡?

OxfordBodleian/Juel-Jensen/BDLaethd20.xml
```diff
@@ -3566,13 +3567,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                    Incipits (one word to one or several lines, consecutive or alternating with black lines in different groupings); 
+                                    captions of miniatures; holy names (<persName ref="PRS6819Mary"/>, <persName ref="PRS4377Gabriel"/>, 
+                                    rarely <persName ref="PRS5684JesusCh"/>); the names of individuals related to the production of 
                                     the manuscript (<persName ref="PRS12911KidanaK"/>,
                                     <persName ref="PRS13031WaldaA"/>, <persName ref="PRS12964Gorgoryos"/>
-                                    <persName ref="PRS13026KenfaM"/>), and captions of miniatures.
+                                    <persName ref="PRS13026KenfaM"/>); numerals; text dividers; elements of punctuation signs.
                                 </seg>
                                 <list type="abbreviations">
                                     <item><abbr>አ</abbr>, <abbr>አሜ</abbr> for <expan>አሜን</expan> (e.g., on <locus target="#9rb #13ra #93ra #96vb"/>);</item>
```
MassimoVilla, line 3578: no comma after e.g. (@SusanneHummel ?)

MassimoVilla (changes requested):
Thank you!

OxfordBodleian/Juel-Jensen/BDLaethd20.xml
thea-m: poor susanne :( I've added this to the guidelines document in the hope that she does not have to confirm comma use with every new PR...

MassimoVilla (approved):

OxfordBodleian/Juel-Jensen/BDLaethg37.xml
```diff
@@ -401,8 +401,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                 </collation>
                                 
                                 <condition key="good">The first leaf and the outer margins of many pages are stained by dirt, e.g. 
-                                <locus target="#25 #30v #31r"/>. The upper edge might have been slightly damaged by fire.
-                                    All edges have been trimmed.</condition>
+                                <locus target="#25 #30v #31r"/>. The upper edge might have been slightly damaged by fire.</condition>
```
ElianaDalSasso, line 404: It is hard to tell from the pictures with any degree of certainty if this is the case, but fire damage usually causes the parchment to shrink and wrinkle ... here I would see "only" (heavy) insect droppings

ElianaDalSasso (commented):
Thank you :)

OxfordBodleian/Juel-Jensen/BDLaethg37.xml
thea-m: Thank you! I've added it, but of course it would be good if you revised the existing condition descriptions in your own workflow, clearly they will become much better!

ElianaDalSasso: Thank you Dorothea!I will do it!

OxfordBodleian/Juel-Jensen/BDLaethf29.xml
SusanneHummel: @thea-m : you are right, the encoding at this place also needs to be adjusted. I must have overseen it wghen I adjusted all other occurences. Thanks for spotting it. I wold need to do some more corrections anyway. Would it be ojay if I include all in my file, you you could delete it here?

thea-m: Yes, I had already deleted it in https://github.com/BetaMasaheft/Manuscripts/pull/878/commits/0af6bff8315b9e4da3fbdb91315c117082c8815d, it should be as before now

SusanneHummel (approved):

PR #883, opened 2020-11-25
B lorient656

MarcinKrawczuk:


LondonBritishLibrary/orient/BLorient656.xml
```diff
@@ -56,60 +90,67 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+
+                        </handDesc>
                         <additions>
-                            
-                            
+<list>
+<item xml:id="e1">
+<locus target="#72r #200r"/>
+<desc>On <locus target="#72r">f. 72r</locus> and <locus target="#200r">f. 200r</locus> a name partially erased, perhaps that of the scribe or owner,
+  <foreign xml:lang="gez"><gap reason="illegible" resp="bm:Wright1877BM"></gap>... ጊዮርጊስ፡</foreign></desc>
```
DariaElagina, line 126: three dots are not needed, if you have encoded the gap. I also do not think that the resp is right here...

DariaElagina (approved):

PR #887, opened 2020-11-26
Image links and Sənkəssār ID

thea-m:
updated Sənkəssār general ID to second recension ID
Added links to images in the British Library (not all manuscripts are listed under Ethiopic, searching for each shelfmark is apparently necessary :( @DariaElagina @MarcinKrawczuk)
Some minor random corrections if I saw something invalid and easily fixable in opened records, without system.
@PietroLiuzzo I'm not sure whether these might be too many records for a PR?

### DIFF
--- a/LondonBritishLibrary/orient/BLorient732.xml
+++ b/LondonBritishLibrary/orient/BLorient732.xml
@@ -127,17 +127,20 @@ schematypens="http://relaxng.org/ns/structure/1.0"
             <textClass>
                 <keywords>
                     <term key="Commentary"/>
-                    
+                    <term key="ChristianLiterature"/>
                 </keywords>
             </textClass>
-            <langUsage/>
+            <langUsage><language ident="gez">Gǝʿǝz</language></langUsage>
         </profileDesc>
         <revisionDesc>
             <change who="SG" when="2019-04-25">Created catalogue entry</change>
-            
+            <change when="2020-11-26" who="DR">Added link to images</change>
             
         </revisionDesc>
     </teiHeader>
+    <facsimile>
+        <graphic url="http://www.bl.uk/manuscripts/Viewer.aspx?ref=or_732_fs001r"/>
+    </facsimile>
     <text xml:base="https://betamasaheft.eu/">
         <body>
             <ab/>
--- a/LondonBritishLibrary/orient/BLorient739.xml
+++ b/LondonBritishLibrary/orient/BLorient739.xml
@@ -132,17 +132,20 @@ schematypens="http://relaxng.org/ns/structure/1.0"
             <textClass>
                 <keywords>
                     <term key="Commentary"/>
-                    
+                    <term key="ChristianLiterature"/>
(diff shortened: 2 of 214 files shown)


LondonBritishLibrary/orient/BLorient739.xml
```diff
@@ -132,17 +132,20 @@ schematypens="http://relaxng.org/ns/structure/1.0"
             <textClass>
                 <keywords>
                     <term key="Commentary"/>
-                    
+                    <term key="ChristianLiterature"/>
                 </keywords>
             </textClass>
-            <langUsage/>
+            <langUsage><language ident="gez">Gǝʿǝz</language></langUsage>
```
eu-genia, line 138: and "en"?

LondonBritishLibrary/orient/BLorient732.xml
```diff
@@ -127,17 +127,20 @@ schematypens="http://relaxng.org/ns/structure/1.0"
             <textClass>
                 <keywords>
                     <term key="Commentary"/>
-                    
+                    <term key="ChristianLiterature"/>
                 </keywords>
             </textClass>
-            <langUsage/>
+            <langUsage><language ident="gez">Gǝʿǝz</language></langUsage>
```
eu-genia, line 133: and "en"? here and elsewhere?

eu-genia (approved):
pre-approved with one small question as to langUsage

LondonBritishLibrary/orient/BLorient739.xml
thea-m: I had some time agp asked @PietroLiuzzo whether this was necessary and didn't receive an answer...I vaguely remember that it doesn't matter, but a confirmation would be nice, as the question comes up often!

PietroLiuzzo: Sorry, I do not recall that question, I apologise if I did not reply promptly enough or missed it entirely, or caused unwillingly reasons to assume other offending behaviour. [`<language>`](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-language.html)s and their `@ident` in `<langUsage>` should resolve the reference for any `@xml:lang` in the TEI file. [When these where automatically updated to remove the extra values all values attested in the file where added, those absent removed.](https://github.com/BetaMasaheft/Documentation/issues/1243)  In the [Guidelines](https://betamasaheft.eu/Guidelines/?id=languageIntro) it is pointed out how the presence of extra values is not a problem, like in this case, where there is no `xml:lang="gez"` in the file. This is especially true where we have no text at all, because it will be that little bit of information allowing to get the text searching by language. You do have a `@xml:lang='en'` but that is the standard value for all our TEI files. So, to be correct following the above guidance, you should have en as well, but it is not used by the app for example.

thea-m: Thank you! I'll merge this as it is and make a new PR to add English

PR #894, opened 2020-11-30
Update BLorient815.xml

smaugustine:
The link is to the images for BL Or 814, not 815, which, unfortunately, is not available online.


thea-m:
thank you for noticing! This is also corrected as part of https://github.com/BetaMasaheft/Manuscripts/pull/887/files, which will hopefully be merged soon, so I'll just close this one.

PR #896, opened 2020-12-02
BLorient80

thea-m:


LondonBritishLibrary/orient/BLorient80.xml
```diff
@@ -54,47 +124,69 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
                             </handNote>
-                            
                         </handDesc>   
+                        
                         <additions>
-                            
+                            <list>
+                                <item xml:id="e1">
+                                    <locus target="#1 #112 #113"/>
+                                    <desc>Verses of hymns are written in a very crude hand.</desc>
```
DariaElagina, line 151: Thank you! I wonder if it would rather belong to the hand description?

LondonBritishLibrary/orient/BLorient80.xml
```diff
@@ -54,47 +124,69 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
                         <origin>
-                            
-                            <origDate/>
+                            <origDate notBefore="1801" notAfter="1803" evidence="reign"/>
                         </origin>
                         <provenance>The manuscript had very probably belonged to a church in the <placeName ref="LOC3577Gondar"/> area, 
                             from where it was taken by
                             <persName ref="PRS9430Tewodros"/> and brought to <placeName ref="INS0101MadhaneAlam"/>. It was then looted 
                             by the British <persName ref="PRS7484Napier">Napier</persName> expedition in <date>1868</date>.</provenance>
-
```
DariaElagina, line 97: If I understand the information in the Preface correctly, this manuscript has been purchased by the British Museum and does not belong to those brought by the expedition.

LondonBritishLibrary/orient/BLorient80.xml
thea-m: My interpretation is that these are not part of the main text.

DariaElagina: Maybe than one could also encode those hymns as additions?

thea-m: That's what I had first, but then decided to switch to extras :D I don't think that they are complete hymns, just verses (maybe also contained in the main text?), possibly in the margin? Do you have a type for desc to be suggested, if encoded as additions?

DariaElagina (approved):

PR #905, opened 2020-12-08
Update BLorient594.xml

MarcinKrawczuk:
I only made a cosmetic change to the position of <msItem>s so that everything displays nicely in the app.


eu-genia (approved):
Nesting is surely not cosmetic ;) But all looks good here now

PR #912, opened 2020-12-09
Emml1840

MarcinKrawczuk:
So like I said, I encoded the contents and gave the basic info about the rest. It's a composite one, it turned out.


thea-m (commented):
Thank you, all seems fine to me. Sorry for the delay.
There is a BM-internal rule to use "St", not "St.". If that doesn't clash with any specifications you follow, maybe you could consider changing this?

thea-m (approved):

PR #913, opened 2020-12-10
BLorient453

thea-m:


LondonBritishLibrary/orient/BLorient453.xml
```diff
@@ -58,7 +66,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                         <width>159</width>
                                     </dimensions>
                                 </extent>
-                                
+                                <condition key="deficient">Slightly damaged.</condition>
```
DenisNosnitsin1970, line 69: This statement does not bring much information, but it is probably like this in the catalogue, isn't it?

DenisNosnitsin1970 (approved):

LondonBritishLibrary/orient/BLorient453.xml
thea-m: yes

thea-m: I chose "deficient", because Wright normally does not have any statement on the condition, so I assume that for him to remark this, it must actually be more than what we would understand as "slightly"...but there are no images, we can only guess

PR #915, opened 2020-12-11
Ds10minor

DenisNosnitsin1970:
minor change


thea-m:
I think that you continued to work on a branch that I deleted after the last PR :) Now it should be fine, but in the future, can you open a new branch after each PR?

thea-m (approved):

PR #916, opened 2020-12-13
DS 2 10 21

DenisNosnitsin1970:
Minor corrections related to calendar


thea-m (changes requested):
Thank you! one question

DenisNosnitsin1970:
No, we cannot, but I do not consider this now. We will have to change them in one special swap.

thea-m (approved):

PR #931, opened 2020-12-21
Genbot corrections

thea-m:
@PietroLiuzzo and @Hizkie : There is a schema error in ESqdq004 (which I should have noticed when approving your previous PR, but I hadn't opened the file in oxygen...): 

> The zone element should not be categorized in detail with @subtype unless also categorized in general with @type

I would be great if you could correct this, so that I can work with a valid file and have a better chance at seeing my own errors.


PietroLiuzzo:
I will look at this now.

PietroLiuzzo:
@thea-m in this PR you are pushing your work, so that I can make the corrections you mention before it is merged?

thea-m:
@PietroLiuzzo that was my thought, in case you have time for this now, yes. Thank you! If it takes longer, no trouble!

PietroLiuzzo:
the fix for this was trivial. I am trying to understand from where do those attribute come from.

thea-m:
Great, thank you!

PietroLiuzzo:
They come out of Transkribus. These column labels have been added to image 72, apparently manually, I have not yet understood how. It could be an interesting feature, but at the moment we have to make sure when we copy from there that these are not present. only these two occurrences cropped out so far.

PietroLiuzzo:
should I also squash and merge?

thea-m:
Thanks!

PR #938, opened 2021-01-05
Bl orient811

DariaElagina:


LondonBritishLibrary/orient/BLorient811.xml
```diff
@@ -52,48 +310,70 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
                         </objectDesc>
-                        
+
+
                         <handDesc>
-                            
+
                             <handNote script="Ethiopic" xml:id="h1">
-                                <desc>The manuscript is well written  apparently in <date>1755</date>
+                                <desc>The manuscript is well written apparently in <date>1755</date>
```
thea-m, line 332: Is this date given in the manuscript? Is it the intersection of the mentioned persons' lifespans?

thea-m (changes requested):
and same question on the Wonders of the World (and also maybe the patriarchs' list) as in the previous PR

LondonBritishLibrary/orient/BLorient811.xml
DariaElagina: This is indeed an intersection of king's reigns. The date has been proposed by Wright.

thea-m (approved):

PR #944, opened 2021-01-12
colophon try, once again

thea-m:
This will hopefully correct the script size


PietroLiuzzo:
mmm... do you need help?

thea-m:
If this makes the English script appear as that of main text, no, otherwise, yes!

PietroLiuzzo:
this remains highly cryptic to me, sorry. if this change in `@xml:lang` is made for the printed catalogue, changing from `gez` to `ti` will have no visible effect, because the settings are the same for both languages, i.e. `letter-spacing = '0.5pt'  font-size ='0.9em'`. Removing as you did one or the other (which, btw, looks to me to be more correct encoding in any case) will default on Ludolfus without further specifications. If this is what you wanted to achieve you are all set!

DenisNosnitsin1970 (approved):

thea-m:
That is what we wanted and what we have achieved.

PR #949, opened 2021-01-14
Emip metadata 23 december 2020

Ralph-Lee-UK:


PietroLiuzzo:
There are three files which contain validation error on this branch. they may not be some of those edited now, but the errors reported cannot be patched by me. In all three cases two `title`s are present in `msItem`.
in EMIP00223.xml you have
````xml
   <msItem xml:id="ms_i3">
<locus from="9r" to="18v" facs="010"/>
<title type="incomplete" ref="LIT4711Mastabqwe"/>
<title type="incomplete" ref="LIT3188Liton"/>
<note>arranged for the days of the week</note>
</msItem>               
````
please either split in two or remove one of the two or please let me know what was intended with this encoding, so that I can help sort it out.

in EMIP000106.xml

```xml
<msItem xml:id="ms_i1.2.13">
                              <locus from="113v" facs="117"/>
                              <title xml:lang="gez" xml:id="t32"> ጸሎተ፡ እግዝእትነ፡ ማርያም፡ ድንግል፡ </title>
                              <title type="complete" ref="LIT1827Magnif#Mary"/>
                              </msItem>
```
I think here simply the ref belongs in the previous title element. I am not sure why the title element has a xml:id of that sort. copied from the Work file? in which case it does not need to...

in EMIP000105.xml
```xml
<msItem xml:id="ms_i1.14.1">
                          <locus from="46r" facs="091"/>
                          <title>Matins</title>
                          <title>Miracle of Mary: The monk who had no virtue other than observing the fast of Mary</title>
                        </msItem>
```
https://betamasaheft.eu/facet.html?query=Miracle+AND+monk+AND+virtue this seems to be in two mss. perhaps it is worth creating a Work for it following the usual proposal procedure, and then use the ID here?

Ralph-Lee-UK:
@PietroLiuzzo I believe that I have corrected this code now each was actually a different problem. 

You correctly comment that one miracle should have a work file, and in fact the work file already exists. I am working through old code on the Miracles of Mary to include the work refs for each of the miracles.

PR #968, opened 2021-02-02
Sh corrafter meet

SusanneHummel:
Here come my updated records:
- mark up for all proper names
- no space after numerals
- distinction between protective leaves and blank leaves within the text block

### DIFF
--- a/Casamari/MNC008.xml
+++ b/Casamari/MNC008.xml
@@ -426,18 +426,18 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                            <locus from="1v" to="2r"/>
                            <desc>The list is crudely written in pencil and includes the following
                               names: <persName ref="PRS13340WaldaIya"><roleName type="title"
-                                    >ʾAbbā</roleName>Walda ʾIyasus</persName>, <persName
+                                    >ʾAbbā</roleName> Walda ʾIyasus</persName>, <persName
                                  ref="PRS13341GabraWald"><roleName type="title"
-                                 >ʾAbbā</roleName>Gabra Wald</persName>, <persName
-                                 ref="PRS13342GabraWald"><roleName type="title">Qes</roleName>Gabra
+                                 >ʾAbbā</roleName> Gabra Wald</persName>, <persName
+                                 ref="PRS13342GabraWald"><roleName type="title">Qes</roleName> Gabra
                                  Wald</persName>, <persName ref="PRS13348TasfaBeyy">Tasfā
                                  Bǝyy</persName>, <persName ref="PRS13349Dasta"><roleName
-                                    type="title">Qes</roleName>Dastā</persName>, <persName
+                                    type="title">Qes</roleName> Dastā</persName>, <persName
                                  ref="PRS13350Warqenah"><roleName type="title"
-                                 >Qes</roleName>Warqǝnah</persName>, <persName ref="PRS13351WaldYoh"
-                                    ><roleName type="title">Qes</roleName>Walda Yoḥannǝs</persName>,
+                                 >Qes</roleName> Warqǝnah</persName>, <persName ref="PRS13351WaldYoh"
+                                    ><roleName type="title">Qes</roleName> Walda Yoḥannǝs</persName>,
                                  <persName ref="PRS13352Sahle"><roleName type="title"
-                                 >Qes</roleName>Śāhle</persName>, <persName ref="PRS13353KidMar"
+                                 >Qes</roleName> Śāhle</persName>, <persName ref="PRS13353KidMar"
                                  >Kidāna Māryām</persName>, <persName ref="PRS13354GabYoh">Gabra
                                  Yoḥannǝs</persName>, <persName ref="PRS13355GabraWald">Gabra
                                  Wald</persName>. The list begins and ends with unclear scribbles.
--- a/Casamari/MNC008.xml
+++ b/Casamari/MNC008.xml
@@ -47,7 +47,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                   <msItem xml:id="ms_i1" corresp="#h1">
                      <locus from="3r" to="16v"/>
                      <title type="complete" ref="LIT2835RepCh119"><foreign xml:lang="gez">Malkǝʾa
-                           Mikāʾel</foreign></title>
+                              <persName ref="PRS7049Michael">Mikāʾel</persName></foreign></title>
                      <textLang mainLang="gez"/>
                      <note>The work comprises 48 stanzas.</note>
                      <incipit xml:lang="gez">
@@ -69,7 +69,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                   <msItem xml:id="ms_i2" corresp="#h2">
                      <locus from="16v" to="25r"/>
                      <title type="complete" ref="LIT2965RepCh246"><foreign xml:lang="gez">Malkǝʾa
-                           Gabrǝʾel</foreign></title>
+                              <persName ref="PRS4377Gabriel">Gabrǝʾel</persName></foreign></title>
                      <textLang mainLang="gez"/>
                      <note>The work comprises 27 stanzas.</note>
                      <incipit xml:lang="gez">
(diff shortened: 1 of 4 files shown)


Casamari/MNC008.xml
```diff
@@ -222,7 +224,8 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         <extent>
                            <measure unit="leaf">64</measure>
                            <measure unit="leaf" type="blank">9</measure>
```
MassimoVilla, line 226: Perhaps things don't add up here: even excluding f. 1r (a protective leaf), the range from f. 40v to 44v plus ff. 62v-64v gives a total amount of 14 leaves, isn't it?

Casamari/MNC008.xml
```diff
@@ -408,28 +411,39 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
-                                 <roleName type="title">ʾAbbā</roleName> Gabra Wald, <roleName
-                                 type="title">Qes</roleName> Gabra Wald, Tasfā Bǝyy, <roleName
-                                 type="title">Qes</roleName> Dastā, <roleName type="title"
-                                 >Qes</roleName> Warqǝnǝh, <roleName type="title">Qes</roleName>
-                              Walda Yoḥannǝs, <roleName type="title">Qes</roleName> Śāhle Kidāna
-                              Māryām, Gabra Yoḥannǝs, Gabra Wald. The list begins and ends with
-                              unclear scribbles. The name of Bayyanačč Zāwde is written with a blue
-                              ballpoint pen on <locus target="#2v"/>.</desc>
+                              names: <persName ref="PRS13340WaldaIya"><roleName type="title"
+                                    >ʾAbbā</roleName>Walda ʾIyasus</persName>, <persName
```
MassimoVilla, line 429: Here and in the following names, a space should be left after `ʾAbbā</roleName>` or `Qes</roleName>` and before the proper name.

VeroliBGV/BGV003.xml
```diff
@@ -763,7 +763,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                <p>Definitions of prefixes used.</p>
             </xi:fallback>
          </xi:include>
-      </encodingDesc>
+      </encodingDesc>cond
```
MassimoVilla, line 766: cond?

MassimoVilla (changes requested):
Thank you Sanne! I also noticed that "Italy" in Condition should be marked up in some files, but let's see first Marco's contribution to that section.

Casamari/MNC010.xml
```diff
@@ -141,6 +144,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                   <msItem xml:id="ms_i9">
                      <locus target="#39r"/>
                      <title ref="LIT2715John">Excerpt from the Gospel of John</title>
+                     <!-- NO BM-record for John the Evangelist?-->
```
MassimoVilla, line 147: It should be PRS5695John

Casamari/MNC008.xml
SusanneHummel: yes, I don't like this number here: if we count leaf, that 40v is half and 41rv is one leaf? Do we need to give a number? Meybe it would be enough to rpovide the range of folia.

Casamari/MNC008.xml
SusanneHummel: I hate those spaces :)

SusanneHummel (commented):
Thank you Massimo! I added the spaced and the name record for John. I haven't found it before though searched for it.

MassimoVilla (approved):

Casamari/MNC008.xml
antobrita: > Perhaps things don't add up here: even excluding f. 1r (a protective leaf), the range from f. 40v to 44v plus ff. 62v-64v gives a total amount of 14 leaves, isn't it?

Sorry but I'm confused with this calculation. The sum of folia from 40v to 44v + 62v to 64v is 6 (12 pages). Isn't it?

MassimoVilla: In reality 14 pages, not leaves (40v, 41r, 41v, 42r, 42v, 43r, 43v, 44r, 44v, 62v, 63r, 63v, 64r, 64v). 
I think there are two issues here:
1. in BM I think we (or perhaps only I?) have mistakenly given inside `<measure unit="leaf" type="blank"/>` the total amount of pages, not leaves. I am going to correct and give the exact number.
2. Like Sanne, I also feel unsatisfied with this information: is it of any use to the reader or aesthetically pleasing to write something like, e.g., 0.5. or 4.5 leaves? I am in favour of not giving this information, at least in the printed catalogue.

antobrita: > In reality 14 pages, not leaves (40v, 41r, 41v, 42r, 42v, 43r, 43v, 44r, 44v, 62v, 63r, 63v, 64r, 64v), so 7 leaves.
> I think there are two issues here:
> 
> 1. in BM I think we (or perhaps only I?) have mistakenly given inside `<measure unit="leaf" type="blank"/>` the total amount of pages, not leaves. I am going to correct and give the exact number.
> 2. Like Sanne, I also feel unsatisfied with this information: is it of any use to the reader or aesthetically pleasing to write something like, e.g., 0.5. or 4.5 leaves? I am in favour of not giving this information, at least in the printed catalogue.

Clear, now. I also agree that this information does not work as it is structured here. What shall we do, then? Do we provide a range as Sanne suggested? Or we do not give this information at all?

MassimoVilla: I would be in favour of providing the range. Let's assume the following case
```
<measure unit="leaf">82</measure>
<measure unit="page" type="blank">3</measure>
<locus from="1r" to="2r/>
```
We can decide to omit `<measure unit="page" type="blank">3</measure>` (here I intentionally changed "leaf" with "page") in the XML record. This is a general BM issue I think, because our files will be part of the general BM database.
But, most importantly, if we want that the system is able to generate a PDF output in which that range is associated with blank leaves, I think we have to specify it somehow, so we can write something like `<locus type="blank" from="1r" to="2r/>`. I noticed that to add type="blank" in that position does not generate any error in the XML file: the schema allows it.
I would like to write an issue and mention you, together with Pietro and other members of BM, to discuss the page/leaf thing and to see if this solution can fulfill our needs

antobrita: Perfect, Massimo! Thank you. Then we decide which is the best option for our catalogue.

SusanneHummel: Wonderful, Massimo! Thank you both for the discussion. Sorry to join you only now.
I am very much in favour of Massimo's suggestion to encode: `<locus type="blank" from="1r" to="2r/>`.Shall we also have "`<locus type="blank protective leaves" from="1r" to="2r/>` ?

The first would refer to the blank leaves in the textblock, the second to blank guirad/protective leaves. Or does comlicates things?

MassimoVilla: I agree with this further specification, which seemingly clarifies things. Just one thing: I tried and I found out that the schema does not allow a sequence like `type="blank protective"`, i.e. two words as a value. But `type="protective"` is allowed. So, we currently can have:

```
<locus type="protective" from="1r" to="2r/>
<locus type="blank" target="#13r #76r"/>
```

We can potentially ask the system to convert type="protective" in any output we desire, e.g.: Blank protective leaves and similar. I also understand that to type "protective" while we actually mean "protective AND blank" is ambiguous. If we want that `type="blank protective"` is allowed by the schema we need Pietro's help. 
.

SusanneHummel: Maybe we set up anothe rmeeting soon and discuss also this before your (or we) write an issue to Pietro? We could have the meeting just the three of us since we haven't finished all point from the list.

MassimoVilla: Good, let's do that!

antobrita: > <locus type="protective" from="1r" to="2r/>
> <locus type="blank" target="#13r #76r"/>

I like that! I think that the main problem with protective and blank leaves is that, once again, we rely on a consolidated tradition in manuscript cataloguing that has not been recently put into discussion. Maybe it is time to talk about it. I agree with having a meeting. Thank you.

PR #974, opened 2021-02-04
Update BLorient817.xml

DariaElagina:


DenisNosnitsin1970 (changes requested):
I would say: "Royal genealogy" (why "collections"?)

thea-m (approved):

PR #980, opened 2021-02-08
remake of PR with conflicts

PietroLiuzzo:
https://github.com/BetaMasaheft/Manuscripts/pull/792


PietroLiuzzo:
@thea-m this is a bit puzzling for me. I have saved copies of the file from the branch but this is quite different from master in any case. should I keep the div[@type='edition'] of master? i.e. was it edited elsewhere?

PietroLiuzzo:
I see that in the old PR there was nothing from you, so it is safe to assume that yes, this was modified elsewhere and I there is no reason for me to replace what is on master with anything else from the other branch. Sorry for the consufion

thea-m:
There is still some confusion from me, sorry. There had been text missing in the transcription, see my comments in the file. If this is how it is and the text is not there (i.e. it is not due to a purely technical transformation problem that can be easily resolved), that's fine, then I'll merge it and add the text to the file. I have since the first PR continued to edit this file, but this should not concern the transkribus edition part.

PietroLiuzzo:
ok, so the edition is fine as is with your later modifications, is there anything new in transkribus? then I will make a fresh export.

thea-m:
I haven't done anything new in Transkribus, I was simply wondering why everything that is in Transkribus is not here (when it worked very well for all other files)

PietroLiuzzo:
ok, so I will make a fresh export to see if it is simply a miss match of time

thea-m:
Thank you!

PietroLiuzzo:
ok, since there where edits in forms of comments in the old transcription file instead of replacing it I have added and included the new one.

thea-m:
Thank you, it seems to be the same text as before (I checked only the first folia, but the same colums/folia are missing there). Should we delete then Transkribus2, as it seems superfluous? I will correct and add the relevant text to me in the parts that I will work with.

PietroLiuzzo:
ok, I will leave the file and simple un-include it

PR #989, opened 2021-02-10
Tana stubs

PietroLiuzzo:


eu-genia (changes requested):
The second batch of stubs does not seem to have the title/summary/citedrange

eu-genia (approved):

PR #992, opened 2021-02-10
anonymised

thea-m:
I am aware that these are no proper shelfmarks, this is so that we can publish now this "Zahl der Woche", then later we can find a better solution


DenisNosnitsin1970 (commented):
Sorry, actually I assigned a kind of shelfmark to the manuscripts of this collection, which is CHJN, do not know if it helps of not.

DenisNosnitsin1970:
Ok, for the purpose indicated it is fine.

DenisNosnitsin1970 (approved):

thea-m:
Yes, actually it is a good idea to use that one. It had puzzled me because I did not understand, but now I have :)

PR #995, opened 2021-02-10
Update ESmr045.xml

PietroLiuzzo:
https://github.com/BetaMasaheft/Documentation/issues/1694#issuecomment-776734761


PietroLiuzzo:
also these are loading. over VPN it will take some hours before they are really avaialable

PR #1001, opened 2021-02-17
Emip metadata 11 feb 2021

Ralph-Lee-UK:


PietroLiuzzo (approved):
Please use an XML editor, like OxygenXML or any other of your preference. Some alternatives in use by our project are listed here https://betamasaheft.eu/Guidelines/?id=setup Another alternative which is not there is eXide, coming with the exist-db database, and there are others available which will perform simple validation while you type based on the schema. Thanks!

PR #1020, opened 2021-03-05
Create BNFabb2.xml

DenisNosnitsin1970:
new record


ParisBNF/abb/BNFabb2.xml
```diff
@@ -0,0 +1,308 @@
...
+                          <note><bibl>
+                             <ptr target="bm:Abbadie1859Cat"/>
+                             <citedRange unit="page">2</citedRange></bibl>
+                             indicates the presence of only 162 folia.</note>
+                             <measure unit="leaf" type="blank">4<locus target="#1 #2 #7 #57"></locus>
+                          </measure>
+                               <dimensions type="outer">
+                              <height unit="mm">276</height>
+                              <width unit="mm">241</width>
+                              <depth unit="mm"></depth>
```
thea-m, line 155: I think that all empty elements, here and in layout and binding, could be deleted. If the images will be available at some point, they can be added.

thea-m (changes requested):
Thank you!

ParisBNF/abb/BNFabb2.xml
DenisNosnitsin1970: done

thea-m (approved):

PR #1023, opened 2021-03-08
corrections mss 11-24

thea-m:


DayrSuryan/DSEthiop12.xml
```diff
@@ -60,7 +60,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                   </msItem>
 
                   <msItem xml:id="ms_i2">
-                     <locus from="24r" to="37v"/>
+                     <locus from="24r" to="37r"/>
```
DenisNosnitsin1970, line 63: was 37v wrong? it matches the following 38r of i3

DenisNosnitsin1970 (changes requested):

DayrSuryan/DSEthiop12.xml
thea-m: it was wrong, 37v is emtpy

PR #1025, opened 2021-03-09
Chjn007

DariaElagina:
Dating to be added after the discussion

### DIFF
--- a/PrivateCollections/Germany/CHJN007.xml
+++ b/PrivateCollections/Germany/CHJN007.xml
@@ -43,18 +43,19 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                            <dimensions type="outer" unit="mm">
                               <height>195</height>
                               <width>170</width>
-                              <depth></depth>
+                              <depth>75</depth>
                            </dimensions>
                         </extent>
-                        <foliation>Foliated by the cataloguer in the middle of the
+                        <foliation>Foliated in the middle of the
                         lower margin.</foliation>
-                        <condition key="good">The first folia a damaged by insects. Textile inlays of the binding are almost lost. The leather
-                      cover of the binding is rubbed off and damaged by insects.</condition>
+                        <condition key="good">Textile inlays of the binding are almost lost. The leather
+                      cover of the binding is rubbed off and damaged by insects. The spine is
+                    missing.</condition>
                          </supportDesc>
                   </objectDesc>
                   <bindingDesc>
                      <binding contemporary="true" xml:id="binding">
-                        <decoNote xml:id="b1">Two wooden boards covered with tooled leather.</decoNote>
+                        <decoNote xml:id="b1">Two wooden boards covered with brown blind-tooled leather.</decoNote>
                         <decoNote xml:id="b2" type="Boards">Rests of textile inlays.</decoNote>
                         <decoNote xml:id="b3" type="bindingMaterial">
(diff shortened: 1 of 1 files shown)


PrivateCollections/Germany/CHJN007.xml
```diff
@@ -0,0 +1,1175 @@
...
+               <physDesc>
+                  <objectDesc form="Codex">
+                     <supportDesc>
+                        <support>
+
+                           <material key="parchment"/>
+                        </support>
+                        <extent>
+                           <measure unit="leaf">164</measure>
+                        <measure unit="quire"></measure>
```
thea-m, line 42: quire number missing?

PrivateCollections/Germany/CHJN007.xml
```diff
@@ -0,0 +1,1175 @@
...
+                 <physDesc>
+                    <objectDesc form="Codex">
+                       <supportDesc>
+                          <support>
+
+                             <material key="parchment"/>
+                          </support>
+                          <extent>
+                             <measure unit="leaf">163</measure>
+                          <measure unit="quire"></measure>
```
thea-m, line 969: quire number missing

PrivateCollections/Germany/CHJN007.xml
```diff
@@ -0,0 +1,1175 @@
...
+                           <note>Numbered 150.</note>
+                         </msItem>
+                         <msItem xml:id="p2_i1.1.147">
+                           <locus from="130v"></locus>
+                           <title type="complete" ref="LIT2000Mazmur#Ps147"></title>
+                           <textLang mainLang="gez"></textLang>
+                           <note>Numbered 150.</note>
+                         </msItem>
+                         <msItem xml:id="p2_i1.1.148">
+                           <locus from="131r"></locus>
```
thea-m, line 918: Not now for this manuscript (I know it will take an eternity), but I think that as a general rule it would be good to avoid providing only the starting point of texts, but always to indicate the beginning and end.

PrivateCollections/Germany/CHJN007.xml
```diff
@@ -0,0 +1,1175 @@
...
+                 <locus from="163" to="164"></locus>
+               </item>
+                            </list>
+                          </collation>
+                          <condition key="good">Stains of dirt throughout the manuscript. The
+                            first and the last leaves are damaged by insects.
+                            <locus target="#86 #106 #112 #113 #114 #115 #116 #119 #121 #122
+                              #125 #126 #127 #132 #133"></locus> are torn or damaged and repaired.
+                              The right lower corner of <locus target="#158"></locus> is torn down.
+                              The last quire is incomplete.
```
thea-m, line 1070: I don't exactly understand what is meant by "the last quire is incomplete". Is it a bifolium, but you can see that folia are missing in the middle?

thea-m (changes requested):
Thank you and sorry for the delay! amazing detail in the text description :)

PrivateCollections/Germany/CHJN007.xml
DariaElagina: You are right, the text is incomplete, but the quire is complete.

thea-m (approved):

PrivateCollections/Germany/CHJN007.xml
thea-m: Aha, now I understand!

PR #1028, opened 2021-03-14
DS 1 to 23

DenisNosnitsin1970:
Minor corrections


DayrSuryan/DSEthiop4.xml
```diff
@@ -408,19 +408,22 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                      <layoutDesc>
 
                         <layout columns="1" writtenLines="19">
+                          <locus from="1r" to="172v"/>
```
thea-m, line 411: now there are two locus elements, I think one can be deleted?

thea-m (changes requested):
Thank you!

thea-m (approved):

PR #1055, opened 2021-04-01
Update BLorient686.xml

MarcinKrawczuk:
So I noticed that half of the ms_i does not display correctly in the app, my guess is that you have to change corresp to ref


thea-m (approved):
Yes, probably that's it...

PR #1058, opened 2021-04-01
Abbadie 47, ESmy006

DenisNosnitsin1970:
Abbadie 47 new

### DIFF
--- a/ParisBNF/abb/BNFabb47.xml
+++ b/ParisBNF/abb/BNFabb47.xml
@@ -210,8 +210,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
 
                         <decoNote xml:id="b2" type="SlipCase">Two part leather slip case (composed of
                           <foreign xml:lang="am">mahdar</foreign> and <foreign xml:lang="am"> defat</foreign>) was
-                          supplied by persName ref="PRS1071dAbbadi" role="owner"/>.</decoNote>
-                          <decoNote xml:id="b3" type="Other" cert="medium">Textile inlays (squares of purple silk with white strypes)</decoNote>
+                          supplied by <persName ref="PRS1071dAbbadi" role="owner"/>.</decoNote>
+                          <decoNote xml:id="b3" type="Other" cert="medium">Textile inlays (squares of purple silk with white stripes)</decoNote>
 
                     </binding>
                   </bindingDesc>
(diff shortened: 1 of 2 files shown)


ParisBNF/abb/BNFabb47.xml
```diff
@@ -0,0 +1,313 @@
...
+                              </msItem>
+
+                              <msItem xml:id="ms_i1.3.2">
+                                  <locus from="95r"/>
+                                 <title type="complete" ref="LIT1882MarkGo#BiographyMark"/>
+                                 <textLang mainLang="gez"/>
+                              </msItem>
+
+                              <msItem xml:id="ms_i1.3.3">
+                                <title type="complete" ref="LIT1882MarkGo#TituliMark" cert="low"/>
```
eu-genia, line 117: no folio given?

eu-genia (changes requested):

ParisBNF/abb/BNFabb47.xml
DenisNosnitsin1970: Not at all sure that the text is there. The description of Abbadie 47 in the catalogue fragmentary, the manuscript is not online.

ParisBNF/abb/BNFabb47.xml
```diff
@@ -0,0 +1,308 @@
...
+                       <locus from="1r"/>
+                        <title type="complete" ref="LIT1560Gospel"/>
+
+
+                        <msItem xml:id="ms_i1.1">
+                          <locus from="1r"/>
+                           <title type="incomplete" ref="LIT2705Introd">Prefatory part</title>
+
+                           <msItem xml:id="ms_i1.1.1">
+                             <locus from="1r" to="8v" cert="medium"/>
```
thea-m, line 47: I know that writing something like cert="medium" can seem satisfactory to us, but I don't think that it's actually very meaningful, because (at least to me) it is still unclear what is how uncertain. I also don't think that it is printed on the app, so to most readers it will stay hidden. Maybe in addition, an explicit sentence in a `<note>` stating what is uncertain could be helpful.

thea-m (changes requested):

ParisBNF/abb/BNFabb47.xml
DenisNosnitsin1970: done, enough like this? the catalogues leave some space for speculation, as to what the ms includes.

thea-m (approved):
Yes, I think that's an improvement for the readers!

PR #1069, opened 2021-04-12
Dr 21 01 genbot

thea-m:


thea-m:
@PietroLiuzzo , could you have a look at the last commit, where I split ESqdq004 and EMML4398? I get error messages (Error attempting to parse XML file (href='facsimile.xml'), but it seems to me that I did everything as explained in the guidelines. Both facsimile files also complain "The markup in the document following the root element must be well-formed.", maybe because there is no root element? I copied this without changing anything from the original fiiles. (When I added a facsimile root element to test this, there only were more error messages and the error in ESqdq004 and EMML4398 remained).
Thank you!

thea-m:
@PietroLiuzzo would you be able to have a look at this? It would be extremely helpful for me to be able to continue this work. 
Of course, if it is not possible in the nearer future, no problem, maybe you could let me know when it will be?
Thank you!

PietroLiuzzo:
I fixed the two. I simply used the latest version of the transkribus2BM.xsl script, where a change in the structure of the facsimile export was introduced. I hope that actually solves the issue.

thea-m:
Thank you, I think that solved it!

PR #1081, opened 2021-04-20
Update BLorient600.xml

MarcinKrawczuk:
So this ms. has already been checked by Daria, but after I saw it in the app I see it doesn't display correctly, starting Friday. Maybe you could have a look.


thea-m (approved):

thea-m:
Sorry for the delay! Yes, there were just the nesting issues you noticed, it seems to display as it should now.

PR #1146, opened 2021-06-16
Emip metadata 9 june 2021

Ralph-Lee-UK:
Standard EMIP record edits and creation of new ms files


PietroLiuzzo (approved):
validation only

PR #1156, opened 2021-06-22
Bl orient 735

DariaElagina:
Most of the msItems have not been identified and do not have refs. I have also not created any new textual units for them so far, since their status in the manuscript is completely unclear. They might turn out to be additions, at least some of them. Their relation with the text of the commentary is also unclear. So, I prefer to keep the description like that until the digital copy of the manuscript becomes available, or there will be some more work done on the study of the texts which accompany such commentaries. I could not find any relevant information neither in Cowley, nor in EAE.

P.S. The same commentary is stored in the XML file

### DIFF
--- a/LondonBritishLibrary/orient/BLorient735.xml
+++ b/LondonBritishLibrary/orient/BLorient735.xml
@@ -5,7 +5,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
     <teiHeader xml:base="https://betamasaheft.eu/">
         <fileDesc>
             <titleStmt>
-                <title>Commentary on the  Gospel of Matthew, theological texts</title>
+                <title>Tǝrgʷāme Mātewos, theological texts</title>
                 <editor key="SG"/>
                 <editor key="DE"></editor>
                 <editor key="AB" role="generalEditor"/>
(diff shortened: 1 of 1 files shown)


DenisNosnitsin1970 (approved):

LondonBritishLibrary/orient/BLorient735.xml
```diff
@@ -52,48 +183,47 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
-                        </handDesc>   
+                        </handDesc>
                         <additions>
-                            
+                          <list>
+                            <item xml:id="a1">
+                                    <locus target="#3r"/>
+                                    <desc type="OwnershipNote">The note states that the manuscript belongs to the church of
+                                        <placeName ref="INS0101MadhaneAlam"/>.</desc>
+                                <q xml:lang="gez"><title ref="LIT2616Zenaes">አፈ፡ ወርቅ፡</title> ዘ<placeName ref="INS0101MadhaneAlam">ቅዱስ፡ መድኃኔ፡ ዓለም።</placeName>
```
thea-m, line 209: I don't see a reason for the identification with LIT2616Zenaes in the manuscript

thea-m (changes requested):

LondonBritishLibrary/orient/BLorient735.xml
DariaElagina: You are right, I have overlooked that

thea-m (approved):

PR #1165, opened 2021-06-23
new records

thea-m:


Hamburg/HLC/RDW003.xml
```diff
@@ -0,0 +1,238 @@
...
+                     <list>
+                        <item xml:id="a1">
+                           <locus target="#1v"/>
+                           <desc type="GuestText">Short hymn.</desc>
+                           <q xml:lang="gez"/>
+                        </item>
+                        
+                        <item xml:id="a2">
+                           <locus target="#44v"/>
+                           <desc type="GuestText">Main text continued in a different hand.</desc>
```
DariaElagina, line 170: I am not sure about the category of an addition and Guest text here. It is not the guest text but the main one... Is to just switch of the hand?

Hamburg/HLC/RDW003.xml
```diff
@@ -0,0 +1,238 @@
...
+                  
+                  <msItem xml:id="ms_i2">
+                     <locus from="41r" to="43v"/>
+                     <title type="complete">Śǝrʿāta ʾaqqʷāqʷām</title>
+                     <textLang mainLang="gez"/>
+                  </msItem>
+                  
+                  <msItem xml:id="ms_i3">
+                     <locus from="44r" to="44v"/>
+                     <title type="complete">Mazmur ba-1 Yoḥannǝs</title>
```
DariaElagina, line 48: incomplete? taking into account a2?

Hamburg/HLC/RDW003.xml
thea-m: I actually do not understand the text (Mazmur ba-1 Yoḥannǝs) well enough right now. I will most probably anyway create an issue for its identification, maybe the identification of the end will also be helped then. For now, we could comment the addition and the type="complete" out.

DariaElagina:
Do you prefer to merge the branch as it is and make a new one, once the text is identified? Or would you prefer to keep it for possible changes?

thea-m:
I will merge it with the new commented out parts, so as to avoid many long-term unpublished branches, and continue to edit it

Issue #16, opened 2019-09-17
Asmat prayers as main text items and as additions


Ralph-Lee-UK:
I have had an initial discussion with Pietro. The problem is that when an asmat prayer is an additional item it can be identified as such with a @type entry, but it if is a main item of the text (which it occasionally is in the EMIP collection) then there is no special way of identifying it as an asmat prayer. The suggestion is that we create a narrative unit for asmat, and this narrative unit should be linked to the keyword Asmat, so that both the identified and unidentified texts related to either the keyword generically or to a more fitting narrative unit, either in the main content or as additions in a manuscript, would be picked up in a request for the keyword or the concept.

PR #1190, opened 2021-07-15
Sophia3

SophiaD-M:
I created the description of BerOrOct555 but there is the issue of part2 that I am not fully sure about. 
Happy to receive your comments.


Berlin/BerOrOct555.xml
```diff
@@ -0,0 +1,218 @@
...
+                      <citedRange unit="page">50-54</citedRange>
+                    </bibl>
+                    </listBibl>
+                    </source>
+                  </recordHist>
+                  </adminInfo>
+                </additional>
+
+            <msPart xml:id="p1">
+               <msIdentifier> <idno>SBP, Ms. orient. oct. 555, p1</idno>
```
thea-m, line 53: this idno is not necessary, I don't think that p1 will be understandable to those who don't encode

Berlin/BerOrOct555.xml
```diff
@@ -0,0 +1,218 @@
...
+             </msIdentifier>
+
+               <msContents>
+                 <msItem xml:id="p1_i1">
+                     <title ref="LIT1119DossierQMQ"/>
```
thea-m, line 58: I would also include here a locus with the extent of the entire hagiographic dossier

Berlin/BerOrOct555.xml
```diff
@@ -0,0 +1,218 @@
...
+                 <msItem xml:id="p1_i1.1">
+                     <locus from="6ra" to="94rb"/>
+                     <title type="complete" ref="LIT1451GadlaG"/>
+                     <textLang mainLang="gez"/>
+                     <incipit xml:lang="gez"> <hi rend="rubric">በስመ፡ አብ፡ ወወልድ፡ ወ</hi>መንፈስ፡ ቅዱስ፡ <hi rend="rubric">፩</hi>አም<hi rend="rubric">ላክ፡ በስመ፡ እግዚአ</hi>ብሔር፡ አብ፡
+                     ዘኢይትዌ<hi rend="rubric">ለጥ፡ እምህላዌሁ።</hi> ወበስሙ፡ ለወልድ፡ እግ<hi rend="rubric">ዚአብሔር፡ ዘኢይተፈ</hi>ለጥ፡ እምህላዌሁ፡ ለአ<hi rend="rubric">ቡሁ።</hi> </incipit>
+                     <note>The manuscript is ms M in Marrassini's 2003 edition of the text.</note>
```
thea-m, line 65: It would be good to encode the bibliotgraphic reference using `<bibl>`

Berlin/BerOrOct555.xml
```diff
@@ -0,0 +1,218 @@
...
+                        <dimensions type="outer" unit="mm">
+                           <height>150</height>
+                           <width>130</width>
+                           <depth>70</depth>
+                        </dimensions>
+                      </extent>
+
+                      <foliation>
+                         <list>
+                            <item xml:id="f.1r">
```
thea-m, line 95: I really don't think that the list element is necessary if there is only one foliation. I would simply write
<foliation> <locus from="1r" to="120r"/>Foliation in pencil in the upper right corner.</foliation>.

Berlin/BerOrOct555.xml
```diff
@@ -0,0 +1,218 @@
...
+
+                      <layoutDesc>
+                            <layout columns="2" writtenLines="17"/>
+                            </layoutDesc>
+                      </objectDesc>
+
+                      <handDesc>
+                         <handNote xml:id="h1" script="Ethiopic">
+                           <locus from="6ra" to="118ra"/>
+                           <seg type="script">Late fifteenth-century script.</seg>
```
thea-m, line 111: you could also add here a date element

Berlin/BerOrOct555.xml
```diff
@@ -0,0 +1,218 @@
...
+                             </decoNote>
+                            <decoNote xml:id="b2" type="bindingMaterial">
+                            <material key="wood"/>
+                            <material key="leather"/>
+                            <material key="textile"/>
+                            </decoNote>
+                            <decoNote xml:id="b3" type="SewingStations">4</decoNote>
+                            <decoNote xml:id= "b4" type="Endbands">Head and tail bands in dark brown leather.</decoNote>
+                            <decoNote xml:id="b5" type="Other">
+                              <term key="leafStringMark"></term> inserted in the upper corner of ff. 6, 16, 20, 35, 53, 65, 78, 82, 86, 94
```
thea-m, line 148: I would use here a locus element to avoid having to decide for ff. fols etc. `<locus target="#6 #16 #20...">`

Berlin/BerOrOct555.xml
```diff
@@ -0,0 +1,218 @@
...
+                              </decoNote>
+                          </binding>
+                        </bindingDesc>
+
+                </physDesc>
+                </msPart>
+
+                <msPart xml:id="p2">
+                  <msIdentifier> <idno>SBP, Ms. orient. oct. 555, p2</idno>
```
thea-m, line 157: see above

Berlin/BerOrOct555.xml
```diff
@@ -0,0 +1,218 @@
...
+                  </msIdentifier>
+
+                  <msContents>
+                     <msItem xml:id="p2_i1">
+                       <locus target="#119ra"/>
+                      <title>Possibly a commemorative or ownership note.</title>
+                      <note>Material loss, text incomplete, beginning is lost. No names inserted</note>
```
thea-m, line 164: you could use <title type="incomplete"> for both texts

thea-m, line 164: I would end with a period

thea-m (changes requested):
Thank you for this excellent work! 
Yes, the extra leaf is complicated. But if you're relatively sure that including its text it originated from another manuscript, I would stick to the current encoding.

Berlin/BerOrOct555.xml
SophiaD-M: ahhhhh :)

thea-m (approved):

PR #1197, opened 2021-07-20
Bl orinet602

DariaElagina:


DenisNosnitsin1970 (approved):
Is the triplet "full-page", "half-page", "small miniature" is in the guidelines, or it originates from the catalogue?

DariaElagina:
It is my addition since the miniatures are very variable in size in the manuscript.

PR #1200, opened 2021-07-26
rdw006

thea-m:


Hamburg/HLC/RDW006.xml
```diff
@@ -0,0 +1,410 @@
...
+                           <width>76</width>
+                        </dimensions>
+                     </decoNote>
+                  </decoDesc>
+
+               </physDesc>
+
+               <history>
+                  <origin>
+                     <origDate notAfter="1800"/>
```
DariaElagina, line 168: notBefore?

thea-m: You're right, I was hoping I could make it pass like this...I can just write 2019, hoping we'll get information on the acquisition.

DariaElagina (approved):

PR #1230, opened 2021-08-23
B lorient577

MarcinKrawczuk:
Please have a look at no. 22 and 24, they differ a bit from the rest, not sure whether this is a correct way to encode them, the ms. is not online sadly.


thea-m (approved):
Thank you! I think it's fine, without the manuscript it's impossible to estimate the intended structure

PR #1232, opened 2021-08-24
Update ESmy023.xml

DenisNosnitsin1970:
Finishing MY-023.

### DIFF
--- a/ES/ESmy023.xml
+++ b/ES/ESmy023.xml
@@ -32,10 +32,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                   <idno>MY-023</idno>
                </msIdentifier>
                <msContents>
-                  <summary>I) Gadla Yoḥannǝs Maṭmǝq “Vita of John the Baptist” (fols. 4ra-78ra);
-                     II) Taʾammǝra Yoḥannǝs Maṭmǝq “Miracles of John the Baptist”: 13 miracles (fols.
-                     78ra-88vb); III) Malkǝʾa Yoḥannǝs Maṭmǝq “Image of John the Baptist”: 28
-                     strophes (fols. 89rb-90va); Colophon.</summary>
+                  <summary></summary>
 
                     <msItem xml:id="ms_i1">
                        <locus from="4ra" to="78ra"/>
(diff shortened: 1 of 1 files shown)


ES/ESmy023.xml
```diff
@@ -32,10 +32,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                   <idno>MY-023</idno>
                </msIdentifier>
                <msContents>
-                  <summary>I) Gadla Yoḥannǝs Maṭmǝq “Vita of John the Baptist” (fols. 4ra-78ra);
-                     II) Taʾammǝra Yoḥannǝs Maṭmǝq “Miracles of John the Baptist”: 13 miracles (fols.
-                     78ra-88vb); III) Malkǝʾa Yoḥannǝs Maṭmǝq “Image of John the Baptist”: 28
-                     strophes (fols. 89rb-90va); Colophon.</summary>
+                  <summary></summary>
 
                     <msItem xml:id="ms_i1">
```
eu-genia, line 37: Should there not be a wrapping msItem for the Hagiographic Dossier on JB?

eu-genia (changes requested):

ES/ESmy023.xml
DenisNosnitsin1970: I explained my attitude to this - if you mean one more overarching msItem. I find that it is ok, more or less, for linking the information in BM, but of questionable utility for actual mss descriptions, esp. in printed form (adds one more artificial level that does not exist in reality, makes descriptions - that are already not very - even less transparent, etc.  I would do my catalogues without it).

eu-genia: Yet you just did it for Mamas. I find the information essential once we will want to see the makeup of the hagiographic collections.
But sure if you do not want....

eu-genia (approved):

ES/ESmy023.xml
DenisNosnitsin1970: As I told, for a printed catalogue it is hardly applicable, at least in my modest opinion. Dossiers IDs can exist, but as knots where links to relevant works are gathered, not as levels in ms descriptions. And if texts are consequently interlinked with the records for protagonists, it is nearly the same.

PR #1240, opened 2021-08-25
deleted entries

DariaElagina:
Despite renaming of the files (according to the wish of the collectors), previous files were not deleted. This is the second try.


PietroLiuzzo:
I am very curious about what you are trying again. could you instead point me to the first attempt? thanks.

PietroLiuzzo:
I wonder how you can delete something you have already deleted.

DariaElagina:
It was not my branch, here it is: #992

PietroLiuzzo:
As far as I can quickly reconstruct, one branch #992 was merged where the files where renamed on 10.2. . After that  however, a folder /Neukirchen with the old files, was merged #1145 on 18.6. so that they are again there. Probably the now-deleted branch was created before 992 was merged and was not rebased. While the first operation triggered the deletion marking, the following one triggered the 'remove from deleted list' action, so they came back to live. So, it looks to me like the app has been doing its job quite well in terms of responding to the PR merge requests. But let me ask again then: are you 100% sure these should go? If yes I will merge this change, check to be sure that the deletion mark is triggered (you can check this [here](https://betamasaheft.eu/deleted.html)) and we should be fine. One last question from me: is https://github.com/BetaMasaheft/Documentation/issues/1800 related to this?

DariaElagina:
Dear Pietro, thank you! This should go because this has nothing to do with harmonization but with the wish of the owners to stay anonymous. We can not guarantee that if the file name still contains their name. All further changes were done on the "correct" anonymized files.

PietroLiuzzo (approved):

PietroLiuzzo:
looks to me like all went well, the records are deleted as wished and listed as deleted accordingly. do let me know if you encounter unexpected behaviours.

DariaElagina:
Thank you so much!

PR #1254, opened 2021-09-04
Update ESmy026a.xml

DenisNosnitsin1970:
New record. Later I will change the shelfmark to a normal number without the letter; the book was t was fotographed together with MY-026.


DenisNosnitsin1970:
Please also this one :)

eu-genia (approved):

PR #1259, opened 2021-09-14
specified filename structure in path to images on server

PietroLiuzzo:


PietroLiuzzo:
https://github.com/BetaMasaheft/Documentation/issues/1815

PR #1263, opened 2021-09-15
Update BLorient617.xml

MarcinKrawczuk:
Sorry to bother you again by this ms., but I saw a mistake in the structure, I only see them after they're merged :(


thea-m (approved):
No need to apologise, 1) I should have noticed and 2) thank you for actually checking after the merge!

PR #1282, opened 2021-09-30
Update DabSey001.xml

Hasnake:


PietroLiuzzo:
having no images or original to check and no knowledge of Geez, I am assigning another reviewer.

thea-m:
I also have no images. Maybe @nafisa-valieva can help?

nafisa-valieva (approved):
Thank you! Changes are definitely accepted! some other variants are indicated, if possible, they can be integrated in the revised version as well!

PR #1284, opened 2021-10-05
B lorient686 corr mk

MarcinKrawczuk:
The story is thus: Zhenia started the cataloguing (until no. 14), and I was supposed to finish. I did it so that the layout is consistent with what she had started. Some info given in the catalogue (e. g. no. 3, the martyrdom of Fasiladas) I left behind, I think it belongs to the work record and I will be happy to include it there.

### DIFF
--- a/LondonBritishLibrary/orient/BLorient686.xml
+++ b/LondonBritishLibrary/orient/BLorient686.xml
@@ -418,6 +418,8 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                             <title type="complete" ref="LIT6207CyriacusVictor"/>
                             <note>read on <date ref="ethiocal:Miyazya27">27 Miyāzyā</date>
                          </note>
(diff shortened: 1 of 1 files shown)


LondonBritishLibrary/orient/BLorient686.xml
```diff
@@ -186,22 +582,24 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+                          <item xml:id="a1">
+                          <locus target="#1r"/>
+                          <desc type="OwnershipNote">The note states that the manuscript belongs to the church of
+                          <placeName ref="INS0101MadhaneAlam"/>.</desc>
+                          <q xml:lang="gez">ገድለ፡ ሰማዕታት፡ ዘ<placeName ref="INS0101MadhaneAlam">ቅዱስ፡ መድኃኔ፡ ዓለም።</placeName></q>
+                          </item>
 
-                           </item>
 
+<item xml:id="a2">
```
thea-m, line 593: I would encode both these items with an "e" ID, because they are not later additions.

MarcinKrawczuk: I will keep that in mind!

PR #1297, opened 2021-10-25
BDLaethg35

thea-m:
Hello again,

Here's the next one. I'm aware that not everyone finished looking at b. 6, and I absolutely don't want to make any pressure!! I just thought it's better to put everything that's there out, so that all can arrange their time as is best for them.
Here, I have two issues:

- could you have a look at the pencil note on the inner front board, which I can't read completely?
- there is nothing specific about the history, my palaeographic dating is also very broad. We might discuss in a meeting how to deal with such cases and whether we'll include a history section there.

Thank you and please take your time!!
Dorothea

### DIFF
--- a/OxfordBodleian/Juel-Jensen/BDLaethg35.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethg35.xml
@@ -244,10 +244,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                 numerals; elements of punctuation signs.</seg>
                                 <list type="abbreviations">
                                     <item>
-                                        <abbr>ሰ፡</abbr> or <abbr>ሰላ</abbr> for <expan>ሰላም፡</expan> (e.g. <locus target="#14 #26r #45r"/>)
+                                        <abbr>ሰ፡</abbr> or <abbr>ሰላ</abbr> for <expan>ሰላም፡</expan> (e.g. <locus target="#14r #26r #45r"/>)
                                     </item>
                                     <item>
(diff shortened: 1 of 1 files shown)


OxfordBodleian/Juel-Jensen/BDLaethg35.xml
```diff
@@ -48,11 +153,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                             <supportDesc>
                                 <support>
                                     <material key="parchment"/>
+                                    <p><locus from="ir" to="iiv"/> and have been cut at the bottom.</p>
```
ElianaDalSasso, line 156: "and have": is something is missing before "and"?

ElianaDalSasso (commented):
Thank you again Thea! Just a doubt, perhaps it is a typo.

OxfordBodleian/Juel-Jensen/BDLaethg35.xml
thea-m: no, it was a mistake, thanks!

thea-m:
@gnizla This is waiting for your review - if you could have a look at it, that would be great (not urgent of course, just noting this now so that it doesn't get lost). Thank you :)

gnizla:
> @gnizla This is waiting for your review - if you could have a look at it, that would be great (not urgent of course, just noting this now so that it doesn't get lost). Thank you :)

I am sorry, I had looked at this a long while ago, I did not have any comments other than thank you so thought I would not add an extra email!

PR #1305, opened 2021-10-31
Dibq gasha

Getinet309:


DebreGolgotaDibqGasha/DGDG001
```diff
@@ -0,0 +1,154 @@
...
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <msDesc xml:id="ms">
+                    <msIdentifier>
+                        <repository ref="ETH0001DibqGasha"/>
```
DariaElagina, line 25: Dear @Getinet309 ,

thank you for the collaboration! I think there is a some confusion with the ids for the repository and works which you have used in the entry. I could not find the ID ETH0001DibqGasha in our database, and judging from the prefix (ETH) it should refer to an ethnic group, but not to an institution/repository.

DebreGolgotaDibqGasha/DGDG001
```diff
@@ -0,0 +1,154 @@
...
+                        <idno>DGDG001</idno>
+                    </msIdentifier>
+                    <msContents>
+                          <msItem xml:id="ms_i1">
+                              <locus from="4r" to="50v"/>
+                              <title type="incomplete" ref="ETH0001SeneGol"/>
```
DariaElagina, line 31: Here, and below, I am not sure that the ID ETH0001SeneGol is a correct/existing one. In the title, you can (or should :)) refer to an internal ID of a literary work, for which you can search here (https://betamasaheft.eu/works/list) using different forms of a title. Maybe you mean LIT2246salotz? (https://betamasaheft.eu/works/LIT2246salotz/main)

DariaElagina (changes requested):

DebreGolgotaDibqGasha/DGDG001
Getinet309: Dear Daria,
Thank you I will change it, but the ms is not in the Betametsahft repository. A new ID is required.

DariaElagina: Dear Getinet,

sure, and you have created a new ID for the manuscript, which is correct! Now, you might search in the database if we already have an entry for the monastery, in which the manuscript is kept. I have conducted a preliminary search, but I could not find it. You could cross-check again, and in case we do not have an entry for this monastery, you could create a new record for it as well. (https://betamasaheft.eu/Guidelines/?id=institution)

DebbeqGasha/DG001.xml
```diff
@@ -0,0 +1,189 @@
...
+                    <msContents>
+                        <summary/>
+                        
+                          <msItem xml:id="ms_i1">
+                              <locus from="4r" to="50v"/>
+                              <title type="complete" ref="LIT2246salotz"/>
+                          </msItem>
+                        
+                                <msItem xml:id="ms_i1.1">
+                                    <locus from="4r" to="4v"/>
```
thea-m, line 39: I think we are nearly finished :) I just have some questions about the division into subsections. You encode 4r-50v as LIT2246salotz, and then subdivide this into 4r-4v (with incipit) and 4v-50v (with explicit). Is there a new section beginning on 4v?

thea-m (changes requested):

DebbeqGasha/DG001.xml
Getinet309: it is an error there is no new section that begins in 4v. the new section is started at 51r. let me correct it.

thea-m:
I think that all information given in this Pull Request has in the meantime already been made present in DG001, so this can be closed. Thanks again!

PR #1311, opened 2021-11-05
fixed place names

thea-m:


PietroLiuzzo:
The changes are clearly necessary but the nature of the error being corrected is quite strange. I want to check the history of the files to see if it was there always

PietroLiuzzo (approved):
I have checked the history of the files and this is still inheritance from mycore. Thanks for the fix.

PR #1312, opened 2021-11-07
Emml7638

MarcinKrawczuk:
This is for my own research outside the contract, if you are busy at the moment I can send it to sb else.

### DIFF
--- a/EMML/7001-8000/EMML7638.xml
+++ b/EMML/7001-8000/EMML7638.xml
@@ -115,6 +115,30 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
 <textLang xml:lang="gez"></textLang>
 </msItem>
 
+<msItem xml:id="ms_i1.13">
+<locus from="71r" to="83v"/>
+<title type="complete" ref="LIT2149OntheW"/>
+<textLang xml:lang="gez"></textLang>
+</msItem>
+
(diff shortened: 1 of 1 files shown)


EMML/7001-8000/EMML7638.xml
```diff
@@ -35,19 +36,233 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
...
+  <extent>
+  <measure unit="leaf">166</measure>
+  <dimensions type="outer" unit="mm">
+    <height>360</height>
+    <width>240</width>
+  </dimensions>
+  </extent>
                         <foliation/>
-                        <condition key="good"/>
+                        <condition key="deficient">The edges of the folios at the beginning and at the end are damaged.</condition>
```
thea-m, line 216: This sounds rather too insignificant for "deficient" (considering the definition) - but you looked at it and can estimate it better

thea-m (approved):
Thank you, everything seems fine. Just one question on the condition.

EMML/7001-8000/EMML7638.xml
MarcinKrawczuk: After checking the guidelines, "good" seems appropriate, the loss of text is very insignificant (the rest we cannot judge).

thea-m (approved):

PR #1347, opened 2021-12-08
Ub taeth2

MarcinKrawczuk:
The ms. is online. I have two doubts:
1) It shifts from one to two columns, sth which doesn't happen often I guess so I wasn't sure how to encode it
2) The foliation - unlike EMML1194 where there was page numbering, here it is done in a "traditional" way


Tuebingen/UBTaeth2.xml
```diff
@@ -0,0 +1,273 @@
...
+
+<physDesc>
+
+  <objectDesc form="Book">
+      <supportDesc>
+          <support>
+              <material key="paper"/>
+          </support>
+          <extent>
+              <measure unit="page">197</measure>
```
thea-m, line 173: I would use "leaf" here - they are not 197 pages. Do you prefer to use page because it's made of paper?

Tuebingen/UBTaeth2.xml
```diff
@@ -0,0 +1,273 @@
...
+    </handNote>
+  </handDesc>
+
+<additions>
+
+<list>
+
+<item xml:id="a1">
+<locus target="#106v #125r"/>
+<desc type="ScribalSupplication">The scribal supplication at the end of some homilies mentions the scribe
```
thea-m, line 206: Since this is written within the main text, I would go for e instead of a

thea-m (changes requested):
Thank you! The column encoding is correct, this is what we also do for instance for Psalters.
I think that for this manuscript it would be particularly useful to include a `<history>` element with origDate and origPlace (since we know them, and this is useful information that many people also are interested in searching or filtering for) - but of course you can decide in how much depth you want to encode your manuscripts. I can also add this after we merge this PR.

Tuebingen/UBTaeth2.xml
MarcinKrawczuk: Corrected, I had no particular intention here.

Tuebingen/UBTaeth2.xml
MarcinKrawczuk: Done!

thea-m (approved):
thank you!

PR #1350, opened 2021-12-10
new record, updated reference

thea-m:
Do you have an idea to which church Gabra Manfas Qeddus could refer? I find it remarkable that there are additional commemorations of him in the manuscripts Orient. 676 and 661. One might speculate whether they might also originate from this church.


DenisNosnitsin1970:
If the script is a kind of "Gondarine" gwelh, we can speculate that this might have been a GMQ church in Gondar (there must have been one, for sure).

thea-m:
There are no images available online

thea-m:
Then I suggest to leave the description as it is

PR #1352, opened 2021-12-13
Munich2

vitagrazia81:
Dear Dorothea, as for ms Munich 1, I have done some changes in Munich 2, simply saying that the bibliography is "secondary".

I would like to review this change, I would be happy if you could, when you will be able (not urgent). 
Thank you,
Vita

### DIFF
--- a/Munich/Museum5Kont/Munich1.xml
+++ b/Munich/Museum5Kont/Munich1.xml
@@ -29,7 +29,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
             <msDesc xml:id="ms">
                   <msIdentifier>
                      <repository ref="INS0969Munich5kont"/>
-                      <idno>MfVK 88.310022</idno>
+                     <idno>MfVK 88.310022</idno>
                   </msIdentifier>
                   <history>
                      <origin>
@@ -43,7 +43,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                      <adminInfo>
                         <recordHist>
                            <source>
-                              <listBibl type="catalogue">
+                              <listBibl type="secondary">
                                  <bibl>
                                  <ptr target="bm:Six1994VOHD6"/>
                                  <citedRange unit="page">442-444</citedRange>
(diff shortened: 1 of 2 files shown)


thea-m (approved):
I am sure you know this and this change is deliberate, just to confirm: changing the bibliography to "secondary" means that the record is based upon your own description and not on the existing catalogue description. It will also have the consequence that the catalogue by Six will not be listed in this list: https://betamasaheft.eu/catalogues/list.

vitagrazia81:
Dear Dorothea, 
thank you for your time and work. 
Sorry for the question; I am still not familiar with this: do I need do anything now, after you merged commit into master?

About the Bibliography:  thank you for the information. In the beginning I did not consider it as secondary, but then It appeared, on the online description, the information that: "this description is based on the catalogue", but  I did not find it completely correct, because I also viewed the ms in person. On the other hand, I just took some pictures of the folios with miniatures and the ones with some incipit pages and colophon, and I took some codicological data. I only had a fast look at the texts etc.. In case I have to do for instance a full description I will have to rely especially on the Six catalogue. What shall I do? Do you suggest me to change it again?
Thanks Dorothea.
Vita

gnizla:
Thanks both. I think that this research is transformative and adds to the existing literature, so, I think Six should indeed be moved to secondary literature, but let's wait for Thea's opinion.

thea-m:
Good, I also support Vita's change to "secondary". It is somewhat of a false binary that doesn't reflect reality very well. I think the important thing is, once the records are edited with more detail, to really state in very clearly in the change element what is based on what.

vitagrazia81:
Thanks both for sharing your opinion.
Yes, I also think that for this purpose and for this kind of record, Six can be "secondary". But I also agree with you, Dorothea, when you say that, once the records become more detailed, we need to be more clear referring to the sources used. And definitively, in that case, Six has to be considered in a different way, still not the main source but  probably not secondary anymore, at the least for the texts description. Anyway, this would require time to debate, good to know that there is already an ongoing discussion on this issue. 
Thanks Dorothea again for your careful work of reviewing my pull requests.
Vita

PR #1363, opened 2021-12-25
Dab seyon

Getinet309:
at line 171 what is "49"


thea-m:
Hi Getinet, thank you for this PR! I am on vacation until 3rd January and will look at it then.

thea-m:
Dear Getinet, it seems that the work on DabSey001 was carried out on a branch which contained old changes in the other files (DG001). Please correct me if I'm wrong. 
Could you please go to the master branch, synchronize it, and then create a new branch from the master branch? 
Before you do that, you could copy DabSey001 from your current branch and paste it into the file in the new branch, so that you don't have to carry out the same changes twice. 
But when you edit DabSey001  in the new branch, please use the elements `<sic>` and `<corr>` to mark up your changes - while I'm sure that your observations are correct, without having access to the images, it is better to be on the safe side and use these elements.
If anything is unclear, please let me know!

Getinet309:
the branches are not the same but it shows conflict. how it can be.

thea-m:
The reason for this is that this branch (DabSey) was created from the branch DibqGasha, not from the master branch. When creating a new branch, it is always possible to chose the branch on which it should be based, and in this case, it was probably 
DibqGasha. 
This problem can be easily avoided in the future by making sure to create new branches from the freshly synchronized master branch: 
image

DenisNosnitsin1970 (approved):

nafisa-valieva (approved):
I think it is better to merge these changes and then I will correct the file based on the images I have.

PR #1367, opened 2021-12-28
Corr repositories mk

MarcinKrawczuk:
These are all EMML mss. I've done. In two cases (EMML 7638 and EMML 8897) I have deleted the line with {repository} because I cannot find it (the former) or there is no institution record (the latter).

### DIFF
--- a/EMML/8001-9000/EMML8897.xml
+++ b/EMML/8001-9000/EMML8897.xml
@@ -32,8 +32,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
          <sourceDesc>
             <msDesc xml:id="ms">
                <msIdentifier>
-                  <repository ref="INS0004HMML"/>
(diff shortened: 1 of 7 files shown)


EMML/8001-9000/EMML8897.xml
```diff
@@ -32,8 +32,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
          <sourceDesc>
             <msDesc xml:id="ms">
                <msIdentifier>
-                  <repository ref="INS0004HMML"/>
```
eu-genia, line 35: you can use LOC4432Mahdar for now

eu-genia (changes requested):

EMML/8001-9000/EMML8897.xml
MarcinKrawczuk: Done

eu-genia (approved):

PR #1402, opened 2022-02-02
Update EMIP01927.xml

thea-m:
wrong ID (no time to look for the right one, but this one at least should not be there)


DenisNosnitsin1970 (approved):

eu-genia:
I would still leave some title, e.g. `<title type="incomplete">A Hymn to Virgin Mary</title>`

PR #1431, opened 2022-02-12
Update BerOrOct555.xml

eu-genia:
Dear Sophia,
I have been comparing the online SBB descriptions with what we have; I noted that this MS lacked the list of decorations present in the online description so I took it over (in German as it was, it has to be adjusted/translated).
I have also noticed some encoding problems:
1) the binding and general aspects of physical description always belong to the entire MS, not to a specific production unit, so I created the physDesc for the manuscript
2) it is recommended to always have the msContent element, also for composite MSS, at top level, with a `<summary>` element, the `<summary/>` may be left empty, then it will be populated automatically
3) what you describe as a second codicological unit is rather a guest text on a protective leaf which we usually describe as an addition (and Gumbert would describe as part of binding); it is not a codicological unit I believe
I have not changed the msPart, please check the MS and eventually update the description further.
Thank you very much!
Zhenia

### DIFF
--- a/Berlin/BerOrOct555.xml
+++ b/Berlin/BerOrOct555.xml
@@ -33,7 +33,73 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
 
                   <idno facs="https://content.staatsbibliothek-berlin.de/dc/863181112/manifest">SBP, Ms. orient. oct. 555</idno>
                </msIdentifier>
-
+              <msContents><summary>I. Vita des Abbā Gabra Manfas Qeddus, f. 6a-94b;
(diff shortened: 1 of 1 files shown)


eu-genia:
I have added the bibliography too, some directly encoded, some as a comment to be considered if relevant.
If the changes may be merged please approve the pull request, after that you will be able to continue working with the description.
Thank you!

thea-m (approved):
Thank you!
No technical comments from me. I remember that the division into msParts was discussed when the manuscript was created, Sophia will be able to decide this, having probably worked most closely with it. I also imagine that this project might have a separate workflow for the decorations - in any case, Sophia will be able to decide what to keep in the record.

SophiaD-M:
Dear Zhenia,

thank you for your work. There a few issues I have to take care of. Which I will gradually do.

Sophia

PR #1453, opened 2022-02-17
BDLaethf30

thea-m:
Dear all, 
here is the next one. 
Again, a few issues: My hypothesis is that we have two units of production, but written by the same hand (but not intended for the same manuscript) - please let me know if you see evidence for other interpretations (one unit of production, different hands etc)! 
Regarding the person(s?) mentioned in the secondary supplications, see also  https://github.com/BetaMasaheft/Persons/pull/511
Eliana might have more remarks on the condition (or more!), I wrote down what I noticed, of course everything can be edited or expanded.
Thank you!
(Images of one opening and measurements for the second unit - if we agree on this hypothesis - are missing, I already noted that in the table for further action.)

### DIFF
--- a/OxfordBodleian/Juel-Jensen/BDLaethf30.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethf30.xml
@@ -685,7 +685,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                     <!--double pricking f. 66, 82, 84-->
                                     <ab type="punctuation" subtype="Dividers">One chain of red and black small semicircles after
                                         Ps. 100 and Ps. 150, at the end of the Psalms of David, and after
-                                        the Wǝddāse Māryām for Thursday (<locus target="#76v #117v #138r #146rb"/>).
+                                        the <ref type="work" corresp="LIT2509Weddas#Thursday"/> (<locus target="#76v #117v #138r #146rb"/>).
                              </ab>
                                 </layout>
                                 
@@ -875,7 +875,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                         <dim type="left"></dim>
                                     </dimensions>-->
                                     <note> The characters per line per column are 15-17.</note>
-                                    <ab type="pricking">Ruling is visible. Primary pricks are mostly visible, ruling pricks are not visible due to the bad state of condition
+                                    <ab type="pricking">Ruling is visible. Primary pricks are mostly visible, ruling pricks are not visible due to the bad condition
                                         of the leaves' margin.</ab>
                                     <ab type="ruling" subtype="pattern">Ruling pattern: 1A-1A-1A1A/0-0/0-0/C.</ab>
                                     <ab type="ruling">The upper line is written above the ruling. The bottom line is written above the ruling.
--- a/OxfordBodleian/Juel-Jensen/BDLaethf30.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethf30.xml
@@ -628,7 +628,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                     <ab type="ruling" subtype="pattern">Ruling pattern: 1A-1A/0-0/0-0/C.</ab>
                                     <ab type="ruling"> The upper line is written above the ruling. The bottom line is written above the ruling.
                                     </ab>
-                                    <!--double pricking f. 82, 84-->
+                                    <!--double pricking f. 66, 82, 84-->
                                     <ab type="punctuation" subtype="Dividers">One chain of red and black small semicircles after
                                         Ps. 100 and Ps. 150, at the end of the Psalms of David, and after
                                         the Wǝddāse Māryām for Thursday (<locus target="#76v #117v #138 #146rb"/>).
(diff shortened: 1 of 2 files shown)


OxfordBodleian/Juel-Jensen/BDLaethf30.xml
```diff
@@ -180,38 +702,228 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                        <width></width>
+                                    </dimensions>
+                                    <dimensions type="margin" unit="mm" xml:id="margin3">
+                                        <dim type="top"></dim>
+                                        <dim type="bottom"></dim>
+                                        <dim type="right"></dim>
+                                        <dim type="left"></dim>
+                                    </dimensions>-->
+                                    <note> The characters per line per column are 15-17.</note>
+                                    <ab type="pricking">Ruling is visible. Primary pricks are mostly visible, ruling pricks are not visible due to the bad state of condition
```
ElianaDalSasso, line 878: I usually encounter 'bad condition' or 'bad state of preservation' in papers related to conservation, what do you think?

ElianaDalSasso (approved):
Thank you Thea! Perfect! I have only a doubt regarding the phrase 'bad state of condition' because I have not found it in technical conservation jargon so far. But it does not mean it is wrong!

OxfordBodleian/Juel-Jensen/BDLaethf30.xml
```diff
@@ -149,10 +136,535 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                           <material key="wood"/>
+                        </decoNote>
+                        leaf string marker f.  39, 101, 104
+                            </binding>
+                        </bindingDesc>-->
+                        
+                    </physDesc>
+                    
+                    <history>
+                        <p>Both units seem to be written by the same hand. However, <ref target="#p2"/> features different rubrication patterns
```
MassimoVilla, line 209: I can't say for sure whether the manuscript has one or two production units. The hand looks the same to me as well, as does the nuance of the red ink, although the size of the characters is quite different between f. 147v and 148r. It is true that the rubrication pattern also is a bit different in the Anqāṣa bǝrhān, but this is a separate text, so this could be an intentional choice. The number of lines (22 at the end of the first unit, 21 in the second unit) is not that different and might depend on fluctuations between different quires. In general, the size of the leaves looks the same to me between the two proposed units. What do you think @ElianaDalSasso ?

OxfordBodleian/Juel-Jensen/BDLaethf30.xml
```diff
@@ -165,10 +677,20 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
                                     </dimensions>
+                                    <note> The characters per line are 10-35.</note>
+                                    <ab type="pricking">Ruling and pricking are visible. Both primary and ruling pricks are visible.</ab>
+                                    <ab type="ruling" subtype="pattern">Ruling pattern: 1A-1A/0-0/0-0/C.</ab>
+                                    <ab type="ruling"> The upper line is written above the ruling. The bottom line is written above the ruling.
+                                    </ab>
+                                    <!--double pricking f. 66, 82, 84-->
+                                    <ab type="punctuation" subtype="Dividers">One chain of red and black small semicircles after
+                                        Ps. 100 and Ps. 150, at the end of the Psalms of David, and after
+                                        the Wǝddāse Māryām for Thursday (<locus target="#76v #117v #138 #146rb"/>).
```
MassimoVilla, line 688: Can `<ref type="work" corresp="LIT2509Weddas#Thursday"/>` be used here (like `<ref type="work" corresp="LIT2509Weddas"/>` in the layout description of p2)?

OxfordBodleian/Juel-Jensen/BDLaethf30.xml
ElianaDalSasso: I can't say either, but I too think that the shade of red ink is the same. The leaves are terribly damaged but could be slightly smaller. I notice the same differences in the in the size of the characters (do they start on the last line of f. 147v already?). The sewing is damaged and repaired so it is difficult to judge if the leaves have been added at a later stage...now they seems to have been coarsely sewn together and added to the bookblock...but it is a common situation of the last leaves of a ms.
I've just noticed that the last leaves seems to be single leaves (see inner margin of f. 148 photo 149). Does this influence the decision? Maybe not.

ElianaDalSasso:
> Eliana might have more remarks on the condition (or more!), I wrote down what I noticed, of course everything can be edited or expanded.

I will add more information on the condition as soon as I describe the binding. This manuscript is really in a terrible state!

OxfordBodleian/Juel-Jensen/BDLaethf30.xml
thea-m: I would absolutely follow your judgement, I thought nothing in particular when writing this.

OxfordBodleian/Juel-Jensen/BDLaethf30.xml
MassimoVilla: Good, then we can stick to the current structure of the MS description and postpone further discussion once the manuscript is consulted again.

MassimoVilla (approved):

OxfordBodleian/Juel-Jensen/BDLaethf30.xml
ElianaDalSasso: I think that checking again the manuscript would be the best solution!

PR #1454, opened 2022-02-17
BLorient672

thea-m:
note my question on the ruling, what do you think? Thank you!

### DIFF
--- a/LondonBritishLibrary/orient/BLorient672.xml
+++ b/LondonBritishLibrary/orient/BLorient672.xml
@@ -229,8 +229,35 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                         </item>
                                         
                                         <item xml:id="q25" n="24">
+                                            <dim unit="leaf">6</dim>
+                                            <locus from="203r" to="208v"/>
+                                        </item>
+                                        
+                                        <item xml:id="q26" n="25">
                                             <dim unit="leaf">10</dim>
-                                            <locus from="203r" to=""/>
+                                            <locus from="209r" to="218v"/>
+                                        </item>
+                                        
+                                        <item xml:id="q27" n="26">
+                                            <dim unit="leaf">10</dim>
+                                            <locus from="219r" to="228v"/>
+                                        </item>
+                                        
+                                        <item xml:id="q28" n="27">
+                                            <dim unit="leaf">10</dim>
+                                            <locus from="229r" to="238v"/>
+                                        </item>
+                                        
+                                        <item xml:id="q29" n="28">
+                                            <dim unit="leaf">10</dim>
+                                            <locus from="239r" to="248v"/>
+                                        </item>
+                                        
+                                        <item xml:id="q30" n="29">
+                                            <dim unit="leaf">6</dim>
+                                            <locus from="249r" to="iv"/>
+                                            2, stub before 6
+                                            3, stub after 5
                                         </item>
                                     </list>
                                 </collation>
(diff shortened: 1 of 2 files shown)


LondonBritishLibrary/orient/BLorient672.xml
```diff
@@ -0,0 +1,370 @@
...
+                                <desc>Written in a good hand.</desc>
+                            </handNote>
+                        </handDesc>   
+                        
+                        <additions>
+                            <list>
+                                <item xml:id="a1">
+                                    <locus target="#2r"/>
+                                    <desc type="OwnershipNote">The note states that the manuscript belongs to the church of 
+                                        <placeName ref="INS0101MadhaneAlam"/>.</desc>
```
DenisNosnitsin1970, line 292: Please note that if we say about the provenance etc. like this, we obscure the situation, as the book, as many other books were looted from other churches by Tewodros, and only after that they belonged for a while to Medhane Alam (before they were looted again).  Otherwise one might think Fasikawit, Mikael and Elyas donated the book to the Medhane Alam church.

DenisNosnitsin1970 (changes requested):

LondonBritishLibrary/orient/BLorient672.xml
eu-genia: This is true for the entire BL collection; when we know where the MS came from before MA we put this too, but provenance is about the entire history, we write what we know... What is your suggestion?

LondonBritishLibrary/orient/BLorient672.xml
```diff
@@ -241,7 +268,8 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                 <layout columns="3" writtenLines="31 33">
                                     <ab type="pricking">Pricking and ruling are visible.</ab>
                                     <ab type="pricking">Primary pricks and horizontal pricks are visible.</ab>
-                                    <ab type="ruling" subtype="pattern">1A-1A-1A1A1A/0-0/0-0/C</ab>
+                                    <ab type="ruling" subtype="pattern">1A-1A-1A1A-1A1A/0-0/0-0/C</ab>
+                                    <!--vertical lines stop at the pricks, is it still A?-->
```
eu-genia, line 273: this is a question for @DenisNosnitsin1970 I guess )

eu-genia (approved):

LondonBritishLibrary/orient/BLorient672.xml
eu-genia: Since Fasikawit and Mikael have also been identified here https://github.com/BetaMasaheft/Manuscripts/pull/1456/commits/762d12382f4a720abdbba03e7282405d991ac06e can we assume that the MS came to MA from Qalāqǝl Māryām Ṣǝyon?

DenisNosnitsin1970: Well, the story of MA collection is well-known, and the OwnershopNote for MA is also well recognizable. As it is now, it indicates that the book belongs (that means: still belong?) to MA. Previously a slightly different formula was used, somehow hinting to the fact that such a manuscript is one of the books brought by Tewodros II and donated to MA.

DenisNosnitsin1970: Not from Qalaqel Maryam, but both books have smth in common, and the presence of the same persons should be clearly noted and made visible. QS-005 is itself a book  that could be brought to Qalaqel from elswere, and be one of more than 500 mss that the British took from Magdala and left in Tigray.

eu-genia: > 

The different formula is used - here too - under provenance, the Ownership note as addition has simply the rendering of what the note says. I think this is correct as it is.

LondonBritishLibrary/orient/BLorient672.xml
thea-m: discussed with him orally

PR #1460, opened 2022-02-22
updated provenance

thea-m:


thea-m:
Actually, this had not been updated in more than 150 records of the orient collection as well. I will wait with a PR until this one here is approved or discussed. 
In case I'll update all the other ones - @eu-genia , do you know what the maximum reasonable number of files per PR is to avoid syncing problems?

eu-genia:
(Otherwise we can follow the guidelines with switching off the webhook and doing manually, I know how to manage files (adding / deleting), but I don't yet know how to "run a script" in Oxygen (or Exide) - do you?)

eu-genia (approved):

PR #1461, opened 2022-02-22
fixing folio and column numeration in EMML4398

eu-genia:
for some reason Transcribus export confused numeration (it probably has issues when images have sometimes two and sometimes one page), fixed manually


thea-m (approved):
Thank you! Sorry, I forgot about this

PR #75, opened 2019-11-20
Gabra Hemamat file revision E-S to BM

eu-genia:

PR #1462, opened 2022-02-22
Binding desc BDLaethe37

ElianaDalSasso:
Hi all! Here is the bindingDesc of Aeth e. 37. 
Something about the collation is not clear to me, let me know what you think but this might need to be checked on the manuscript. 

I have a doubt: do we add the information about the holes from the parchment making process in the description of the condition or of the support?

Thank you all!

### DIFF
--- a/OxfordBodleian/Juel-Jensen/BDLaethe37.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethe37.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng" 
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng"
 schematypens="http://relaxng.org/ns/structure/1.0"?>
-<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng" 
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng"
 type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
 <TEI xmlns="http://www.tei-c.org/ns/1.0"
     xml:id="BDLaethe37"
@@ -13,7 +13,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 <title>Ṣoma dǝggwa, salām to Mary</title>
                 <editor key="DR"/>
                 <editor key="JG"/>
-                <editor key="MV"/><editor key="SH"/><funder>Akademie der Wissenschaften in Hamburg</funder>
+                <editor key="MV"/>
+                <editor key="SH"/>
+                <editor key="EDS"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
             </titleStmt>
             <publicationStmt>
                 <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
@@ -33,28 +36,30 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                             <idno>Juel-Jensen 11</idno>
                         </altIdentifier>
                     </msIdentifier>
-                    
+
                     <msContents>
                         <summary/>
-                        
+
                         <!--<msItem xml:id="ms_i1">
-                      
+
                         </msItem>-->
-                        
+
                     </msContents>
-                    
+
                     <physDesc>
                         <objectDesc form="Codex">
                             <supportDesc>
                                 <support>
                                     <material key="parchment"/>
+                                    <p>Holes, from the parchment making process, are present in the leaves, e.g. <locus target="#77 #83 #84 #92"/> others have been repaired by sewing, e.g.
+                                    <locus target="#2 #9 #58 #60 #61 #72 #95"/>.</p>
                                 </support>
                                 <extent>
                                     <measure unit="leaf">96</measure>
                                     <measure unit="leaf" type="blank"></measure>
                                     <!--<locus target=""/>-->
-                                    <measure type="weight" unit="g">326</measure><!-- without māḫdar and modern notes placed inside taken out-->
-                                    <measure type="weight" unit="g">544</measure><!--with māḫdar, without modern notes-->
+                                    <measure type="weight" unit="g">326>without slip case</measure><!-- without māḫdar and modern notes placed inside taken out-->
+                                    <measure type="weight" unit="g">544>with slip case</measure><!--with māḫdar, without modern notes-->
                                     <dimensions type="outer" unit="mm">
                                         <height>147</height>
                                         <width>101</width>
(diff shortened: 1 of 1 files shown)


MassimoVilla:
> I have a doubt: do we add the information about the holes from the parchment making process in the description of the condition or of the support?
> 
We add that information within `support`, next to the `material`. For instance:
` <support>
                                    <material key="parchment"/>
                                    <p>Holes, presumably from the parchment making process, are found on <locus target="#3 #75"/>.</p>
                                </support>`

OxfordBodleian/Juel-Jensen/BDLaethe37.xml
```diff
@@ -66,13 +71,17 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
                                 <collation>
                                     <list>
                                         <item xml:id="q1">
                                             <dim unit="leaf">2</dim>
                                             <locus from="1r" to="96v"/>
-                                            <!--these should probably belong to the binding-->
+                                            <!--these should probably belong to the binding.
+                                            EDS (10.02.2022): I am not sure wether these leaves can be considered as part of a quire
+                                            to me they seems to be single leaves and I would not put them in the bindingDesc....in the case they do belong to a quire,
+                                            they should be considered as the outer bifolium, with all the other quires nested inside. Does the present encoding rapresent this structure?
```
MassimoVilla, line 83: Picture JJ-11.0104.jpg is a bit blurred but it seems to me that fols 1 and 96 belong to an outer bifolium which covers the spine.

OxfordBodleian/Juel-Jensen/BDLaethe37.xml
```diff
@@ -158,37 +174,55 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                             <decoNote xml:id="b4">
+                               The sewing is a chainstitch on four sewing stations, broken and repaired with a different kind of thread. Both threads are Z-plied.
+                              </decoNote>
+                             <decoNote xml:id="b5" type="Spine">
+                               The small holes close to the head and tail of the quires probably indicate that quire tackets were originally present.
+                               The spine has been protected by a sort of spine lining made by joining together a brown and a white piece of leather.
+                             </decoNote>
+                             <decoNote xml:id="b6" type="SlipCase">
+                               The original leather case is preserved. It is a one-piece, double-walled satchel where the back panel overlaps over the front.
+                               The satchel is equipped with long flaps extending from the side panels and a main rounded flap extending from the back panel to the front
```
MassimoVilla, line 209: "extending from the back to the front panel" (@gnizla ?)

OxfordBodleian/Juel-Jensen/BDLaethe37.xml
ElianaDalSasso: Ah, you are right! Now I see it. I thought it was a white piece of leather but it's parchment. So, shall we keep them in the collation? They could be considered as extensions of a sort of spine lining...so be included in the bindingDesc

OxfordBodleian/Juel-Jensen/BDLaethe37.xml
gnizla: overalps = 2 panels one covers the other
extend = 1 panel extends from front to back but does not cover the other

MassimoVilla: Sorry, my comment was unclear (and probably unnecessary). I was just proposing to move "panel" at the end of the sentence for simplicity in the word order

OxfordBodleian/Juel-Jensen/BDLaethe37.xml
```diff
@@ -33,28 +36,30 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                    <p>Holes, from the parchment making process, are present in the leaves, e.g. <locus target="#77 #83 #84 #92"/> others have been repaired by sewing, e.g.
+                                    <locus target="#2 #9 #58 #60 #61 #72 #95"/>.</p>
                                 </support>
                                 <extent>
                                     <measure unit="leaf">96</measure>
                                     <measure unit="leaf" type="blank"></measure>
                                     <!--<locus target=""/>-->
-                                    <measure type="weight" unit="g">326</measure><!-- without māḫdar and modern notes placed inside taken out-->
-                                    <measure type="weight" unit="g">544</measure><!--with māḫdar, without modern notes-->
+                                    <measure type="weight" unit="g">326>without slip case</measure><!-- without māḫdar and modern notes placed inside taken out-->
```
thea-m, line 61: encoding problem (one > too many). I think we might settle now whether we want these double weight measures to be included in the catalogue? In any case, the information on the presence or absence of the slip case might be better in a `<note>` directly after the respective measurement.

OxfordBodleian/Juel-Jensen/BDLaethe37.xml
thea-m: I also think (as I did two years ago, apparently) that they should be described in the bindingDesc, they are not part of the quire structure as far as I understand it.

OxfordBodleian/Juel-Jensen/BDLaethe37.xml
```diff
@@ -158,37 +174,55 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                             <decoNote xml:id="b3" type="SewingStations">4</decoNote>
+                             <decoNote xml:id="b4">
+                               The sewing is a chainstitch on four sewing stations, broken and repaired with a different kind of thread. Both threads are Z-plied.
+                              </decoNote>
+                             <decoNote xml:id="b5" type="Spine">
+                               The small holes close to the head and tail of the quires probably indicate that quire tackets were originally present.
+                               The spine has been protected by a sort of spine lining made by joining together a brown and a white piece of leather.
+                             </decoNote>
+                             <decoNote xml:id="b6" type="SlipCase">
+                               The original leather case is preserved. It is a one-piece, double-walled satchel where the back panel overlaps over the front.
```
thea-m, line 208: Here, "leather case" and "satchel" are used, above "slip case". To us, it is clear, but I wonder whether this might be confusing to those who don't know these cases well. Maybe here "leather slip case" could be used in the first sentence?

thea-m (changes requested):
Thanks a lot! Some minor remarks.

OxfordBodleian/Juel-Jensen/BDLaethe37.xml
ElianaDalSasso: I think that for the catalogue the most important information is the presence of the slip case while its weight might not be fundamental.

thea-m: It seems reasonable to me to include the weight of the manuscripts itself, without possible accompanying documents/slip cases (this second measurement could stay commented out in the records). That is already more information on weight than there is anywhere else ;)

OxfordBodleian/Juel-Jensen/BDLaethe37.xml
ElianaDalSasso: So, q2 becomes q1 (and all the other quires change accordingly) and f. 1 and f. 96 go in the bindingDesc, right?

MassimoVilla: I think so, as a consequence of the previous interpretation of this outer bifolium as part of the binding. But the best think would be to consult a handbook of codicology and see how such cases have been treated and if they have been considered in the collation.

ElianaDalSasso: @MassimoVilla  I think it is a good idea, just I don't know where to look...there is not such an handbook ,unfortunately!...the only other similar manuscript I have seen is the ms 119 of the Casamari abbey in Veroli. When I encoded it for Manus online, I have not considered the parchment extensions from the spine in the collation. But the should be in the forthcoming catalogue...how did you encode it?
Do you know of other similar cases?

MassimoVilla: Unfortunately MS Casamari MNC-018 (= 119) did not belong to my share of manuscripts and has not been described yet. As far as can be useful, this is what we wrote down: the textblock is wrapped in a leaf sewn to the guardleaves by means of parchment strips.

ElianaDalSasso: I see, I see...no problem. I propose then a tentative solution that exclude f.1 and f. 96 from the collation. We can then evaluate if it's ok

thea-m (approved):

PR #1463, opened 2022-02-23
IDs changed

vitagrazia81:
I have changed the Ids of  the Munich manuscripts. Could you please check them :-)?


thea-m (changes requested):
Thank you! Is it possible that for Munich 2, 5, and 6 you forgot to change the xml:id inside the file? Could you check that again?

vitagrazia81:
Dorothea, thank you. Could you check again now?

thea-m (approved):

vitagrazia81:
Thank you Dorothea. 
Just a thing: I have seen them online (e.g. https://betamasaheft.eu/manuscripts/MfVK86307647/main) but you don't see the shelfmark anymore. This is for all manuscripts. They are not displayed.  On the xml files there are, I have encoded them (under  <idno>.,,,). Do you think is a bug?

PR #1474, opened 2022-02-28
Create BNFabb115.xml

DenisNosnitsin1970:
New record.

### DIFF
--- a/ParisBNF/abb/BNFabb115.xml
+++ b/ParisBNF/abb/BNFabb115.xml
@@ -1,0 +1,283 @@
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="BNFabb115" xml:lang="en" type="mss">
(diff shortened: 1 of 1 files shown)


eu-genia (approved):

ParisBNF/abb/BNFabb115.xml
```diff
@@ -0,0 +1,283 @@
...
+                         <title type="complete" ref="LIT4002Salotz"></title>
+                         <textLang mainLang="gez"/>
+                         <incipit xml:lang="gez">
+                           ንነግረክሙ፡ አኃውየ፡ መግረሬ፡ ፀር፡ ወመድኅነ፡ ሙሴ፡ እምእደ፡ ፈርዖን፡ ዘወሀቦ፡ ሙሴ፨
+                           </incipit>
+                           <note>The prayer is divided into four parts that
+                           begin on <locus target="#2r"/>, <locus target="#28v"/>,
+                         <locus target="#31r"/>, <locus target="#44v"/>, resp. The name of the scribe, <persName ref="PRS13744Giyorgis">Giyorgis</persName>,
+                         appears in the concluding supplication on <locus target="#49r"/>. C. Conti Rossini suggests
+                       that this specific version of the text may be of <persName ref="ETH1274Betaes"/> origin.</note>
```
thea-m, line 48: I assume that since the text of the manuscript is not available to us, it's not possible to confirm and hasn't been studied yet? Otherwise it would make sense to add this information to the work record (even if it remains a general record)

thea-m (changes requested):
minor question

ParisBNF/abb/BNFabb115.xml
DenisNosnitsin1970: I agree, though this is an assumption only (A. d'Abbadie used to indicate Beta Esrael origin of the manuscripts, in this case he did nothing). There is an issue of the miniatures of in this manuscripts, which makes the Beta Esrael origin not very likely.

PR #1480, opened 2022-03-01
Update BNFabb115.xml

eu-genia:
removed facs and changed shelfmark, please check


DenisNosnitsin1970:
This is fine, but what I intended with the link in "idno" was a part-set of images at http://mandragore.bnf.fr/ark:/12148/cgfbt4443 ("BnF Ethiopien 60" was a mistake). Actually Dorothea recommended to open an issue as to if any other links but those of the "main repository" can be used and how, in "idno" or elsewhere, but I think that "mandragora" is on the same level as "gallica".

eu-genia:
it is not on the same level as it does not have a iiif json manifest.
it is a single image. you can add a link to decodesc through `ref target="http://mandragore.bnf.fr/ark:/12148/cgfbt4443"`

DenisNosnitsin1970:
OK.

PR #1481, opened 2022-03-02
ESqdq004

thea-m:
(since the name of the file turned out to be irrelevant for the problem - it makes life for me much easier if the text file is named with the manuscript ID, when I have multiple files open in oxygen :) )


eu-genia (approved):

eu-genia:
(and still I again had to manually delete the expanded doublet, will have to do it every time a file with an included transcription is updated until the bug https://github.com/BetaMasaheft/Documentation/issues/1955 is resolved)

PR #1490, opened 2022-03-04
Update BLorient576.xml

thea-m:


LondonBritishLibrary/orient/BLorient576.xml
```diff
@@ -38,14 +38,121 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                     <msContents>
                         <summary/>
                         
-                        
+                        <msItem xml:id="ms_i1">
+                            <locus from="11r"/>
+                            <title type="complete" ref="LIT1850Malkea"/>  
+                            <note>The codicological structure of the manuscript cannot be discerned by the catalogue description. Several
```
DenisNosnitsin1970, line 44: I find this sentence a little clumsy, frankly speaking, perhaps smth like: "The catalgue description does not indicate the structure of the manuscript that may be composed of several codicological units" (are there indications for that or it is an assumption?). Also, I would place this information elsewhere, I would not expect in a note ms_i1.

DenisNosnitsin1970 (changes requested):

LondonBritishLibrary/orient/BLorient576.xml
thea-m: I agree with your rewording and that the place is not ideal, do you have a better idea where to place this information? My priority would be maybe not the absolute ideal place from the point of view of the encoding, but one where under the current visualisation users have a good chance at spotting it. How about a `<note>` in support or history?

DenisNosnitsin1970:
History would be fine (I used to place such data in History), Support probably also ok.

eu-genia (approved):

PR #1494, opened 2022-03-07
BLorient578

thea-m:
@MarcinKrawczuk - I noticed too late that this was on your list. I hope you had not started working on it on some unpublished branch.

### DIFF
--- a/LondonBritishLibrary/orient/BLorient578.xml
+++ b/LondonBritishLibrary/orient/BLorient578.xml
@@ -38,7 +38,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         <summary/>
                         
                         <msItem xml:id="ms_i1">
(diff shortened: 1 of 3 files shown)


DenisNosnitsin1970 (approved):

MarcinKrawczuk:
I have actually, there is a branch called BLorient578. Will it be useful to
you in some way?

pon., 7 mar 2022 o 20:21 DenisNosnitsin1970 ***@***.***>
napisał(a):

thea-m:
It will be a lesson for the future. I'll go though both versions and compare them before merging.

thea-m:
@MarcinKrawczuk It was quite useful actually! There where three hymns we identified differently, so that allowed me to merge the concerned work records, which is always good. I hope this did not frustrate you.

PR #1505, opened 2022-03-14
Binding desc bd laethd18

ElianaDalSasso:


ElianaDalSasso:
Hi all! Here is another bindingDesc :)

MassimoVilla:
Unfortunately there will be a conflict with d.18, because in branch Max12 (still open and unmerged because of the "cardboard" issue in MS g.31) I moved the history of p1 in the general description, as agreed concerning the history of composite MSS. 
The first thing to do in my view is to merge Max12, if no one has other observations, merge this branch, solve the conflict keeping in mind that "history" has to be moved from p1 to the general description, and then we can discuss about the position of the physical part.

MassimoVilla (approved):

MassimoVilla:
In general, I think that the binding description should be kept in the general description, because it affects the present unit of circulation in its entirety, that is including both p1 and p2. I consider as more methodologically correct to split the physical part into two parts, but we have to agree to what goes here and what goes there, and possibly apply this to all composite mss of the collection.

MassimoVilla:
No need to apologise @ElianaDalSasso , none knows what is being modified in our private branches.
I propose then that you move the binding part into p1 as suggested by Thea, then - after new approval - Thea or I can merge this branch and, as soon as Max12 is merged too, I will take care of the conflict in the history of d.18 according to the agreed solution.

MassimoVilla (approved):

PR #1513, opened 2022-03-17
nesting fixed

eu-genia:
after this merge the record shall be correct, the nesting was indeed wrong

note that i also corrected the folio numbers and the numbering of msitems (already pushed so not visible, see here https://github.com/BetaMasaheft/Manuscripts/commit/ac78bfffadd1ec6f776ad949aecc746818e4310c)


DenisNosnitsin1970:
I see, thanks a lot!

PR #1525, opened 2022-03-23
Dr 22 03 golden gospel

thea-m:
Reverting Golden Gospel of Dabra Libānos commits
Sorry, @eu-genia , this file is very special and has been encoded with a lot particularities (also described in the new article). For now, the visualisation problem is secondary, @nafisa-valieva will let you know when she will have time to discuss it, but for now, it's better to leave it like this and preserve the encoding as it is. Thank you! It must have been a lot of work already figuring the problems out.

### DIFF
--- a/Ham/DabraLibanosHamGG1/DabraLibanosHamGG1.xml
+++ b/Ham/DabraLibanosHamGG1/DabraLibanosHamGG1.xml
@@ -7,26 +7,27 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
   <teiHeader >
     <fileDesc>
       <titleStmt>
-        <title xml:id="t1">Golden Gospel of Beta Mata za-Ham</title>
+        <title xml:id="t1" type="académique">Golden Gospel of Beta Mata za-Ham </title>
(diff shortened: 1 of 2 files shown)


thea-m:
For sure, it is important and the analysis is correct, but it cannot be dealt with right now, it simply has to wait a bit. I understand that this might be frustrating.

eu-genia:
the string in the code (viewItem)
`                string-join(distinct-values(($TEI//t:editor[@key = $resp]/text() | $TEI//t:respStmt[@xml:id = $resp]/t:name/text())), ', ')
`
says that the resp inside of respStmt is taken as `xml:id`. This is why there may not be multiple `resp `elements. I suggest that at least this is corrected in the data and not in the code.

nafisa-valieva (approved):
Dear Eugenia, this seems to be fine! thank you!

nafisa-valieva:
Dear @eu-genia, dear @thea-m,
thank you very much for the corrections! Because we had discussed a lot with Pietro, I fear to loose any bit of our success. To be able to decide upon bigger changes, I need to immerse in this work and I can't do it for the moment.

PR #1548, opened 2022-04-05
Update EMML51.xml

Dirbwork:


eu-genia (changes requested):
dimensions attributes must be inside

eu-genia (approved):

PR #1557, opened 2022-04-05
Update EMML4.xml

Addisie:


EMML/1-1000/EMML4.xml
```diff
@@ -102,6 +103,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
       <revisionDesc>
          <change who="PL" when="2017-11-30">Created minimal record with link from data sent from Daniel Gullo, HMML
                             project.</change>
+                            <change when="2022-04-05" who="AY"></change>
```
eu-genia, line 106: please add some content to the change element
eg
`<change when="2022-04-05" who="AY">updated dimensions</change>`
otherwise all perfect!
thank you!

eu-genia (changes requested):

EMML/1-1000/EMML4.xml
eu-genia: please add content

eu-genia (approved):

PR #1598, opened 2022-04-07
Dsl 004

MershaMengistie:
I see that the editionStmt is missed.


eu-genia (approved):
it is ok to first create stubs, get them merged, and then you can continue editing, you can see my hnk-001 as the model and copy from there

PR #1600, opened 2022-04-07
Create DLS-006.xml

MershaMengistie:


eu-genia (approved):
thank you, i think before you create more stubs you can 
(1) re-create dls-001 and dls-002 - the stubs are not yet in the github
(2) edit your stubs with the editionstmt, add element about the digitizing (see HNK-001 for the example), correcting transcription where needed (there were some cases), introduce msContents and physDesc.
for the six new records you created. thank you!

PR #1612, opened 2022-04-12
renaming ds-xvi

eu-genia:
Dear Jonas, I suggest to rename this record too, to comply with our naming guidelines
https://betamasaheft.eu/Guidelines/?id=entities-id-structure
If this is OK you can approve and merge.
Thank you!

### DIFF
--- a/new/DS16.xml
+++ b/new/DS16.xml
@@ -1,7 +1,7 @@
 <?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
 schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
 type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
-<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="ds-xvi" xml:lang="en" type="mss">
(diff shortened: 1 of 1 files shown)


eu-genia:
(the ID does not have to be the same as shelfmark)

karljonaskarlsson:
Dear Zhenia,

I think it is not optimal. The fragments from Dabra Sahl were given Roman numbers, while the more complete (and more recent) manuscripts were named using Arabic numbers, so DS-16 is the shelfmark of another manuscript.

eu-genia:
Ok let us leave roman but at least take the dash out )

karljonaskarlsson (approved):

thea-m:
(It is really a pity, because these IDs - which are not shelfmarks - are now against the stated rules of BM IDs, which should consists of letters and numerals. I see the problem, I'm not aware of a better solution, but just to be aware of this.)

karljonaskarlsson:
Would it be better to invent something? DSfragm16?

eu-genia:
I guess we have to - I see the app has problems with files that has no arabic numbers in the filename

thea-m:
I'm really sorry for all this confusion.  Could we then wait a moment and possibly change the ID even more? DS is unfortunately already "taken" by Dayr as-Suryān (now unfortunately difficult to rename due to the published catalogue). Is there per chance any accepted and understandable other abbreviation for Dabra Sahl?

eu-genia:
let us discuss this here https://github.com/BetaMasaheft/Manuscripts/pull/1613 as i have created a new branch

PR #1613, opened 2022-04-12
DSfragm16

eu-genia:
https://github.com/BetaMasaheft/Manuscripts/pull/1612


karljonaskarlsson:
As for using another abbreviation: DS is already used officially for Dabra Śāhl in publications.

karljonaskarlsson:
Otherwise, ADS (for ʾAgʷazā Dabra Śāhl) would be an understandable alternative… but as I said, the shelfmark is already used in publications.

PR #1615, opened 2022-04-12
Create DLS-012.xml

MershaMengistie:


eu-genia:
Dear Mersha, you probably did not read the email I sent today.
We should avoid dashes in filenames.
Please rename the file as DLS012 and change the `xml:id ` (line 4) to `DLS012`
Thank you!

eu-genia:
OK now on this branch you can continue editing the files as above

eu-genia:
Make sure that you do not confuse branches (you now have two active branches)
Only work with the files you just created on these branches
Note that you now have committed DLS-011 twice, in this branch and in the DLS-013 branch
This should not happen in the future
Please do not hurry, I think it is too early for you to work with several branches at a time.
Work slowly. In one branch only. Once it is merged, then you can go to the next one, or you will create a lot of confusion for everyone.

Now files must be renamed and corrected.

MershaMengistie:
Sorry...I was not attentive at all to my yahoo. I read your email warning that we should avoid using dashes. But I think I noticed somewhere (in the process of work) a kind of automatic note that says, "it would be saved as DSL-011. Okay, I will avoid it.

    Am Dienstag, 12. April 2022, 23:03:36 MESZ hat Eugenia ***@***.***> Folgendes geschrieben:  
 
 


Make sure that you do not confuse branches (you now have two active branches)
Only work with the files you just created on these branches
Note that you now have committed DLS-011 twice, in this branch and in the DLS-013 branch
This should not happen in the future
Please do not hurry, I think it is too early for you to work with several branches at a time.
Work slowly. In one branch only. Once it is merged, then you can go to the next one, or you will create a lot of confusion for everyone.

Now files must be renamed and corrected.

eu-genia (approved):

PR #1616, opened 2022-04-12
Dls 013

MershaMengistie:
why don't I see what I inserted as the <editionStmt> in the GitHub after I worked and saved on Oxygen?


eu-genia (changes requested):
1) please rename both files DLS011 and DLS013, both as filename and in xml:id
2) please correct the transcription
3) in order to work with this files please work in the same branch (which you called DLS-013) - commit new versions of these two files (only these two files) to this same branch
you can only see these files and the changes in this branch

eu-genia (approved):

PR #1646, opened 2022-04-24
Update EMML2584.xml

MarcinKrawczuk:
I managed to buy the manuscript, previously I had only transribed the cataloge data, so I've added a few details


thea-m (approved):
Cool! I'm sure we all need this prayer.

PR #1653, opened 2022-04-25
Update EMIP00601.xml

thea-m:


thea-m:
@eu-genia according to the table it had already been inserted?

eu-genia (approved):
the table is probably inprecise, i thought the non-inserted started only at a later point.
now it is, thank you
pietro suggests this may be a task for a hiwi

PR #1663, opened 2022-04-28
Delete ESmy027.xml

DenisNosnitsin1970:
Entry deleted (a few changes of sigratures in MY collection).


eu-genia:
if you change signatures of remaining manuscripts this must be clearly documented as there are also images to rename then

eu-genia:
it would be better not to change signatures, since these may be also documented in publications (including yours)

DenisNosnitsin1970:
Please merge.

eu-genia:
please make an issue with a list of changes/deletions in E-S shelfmarks, referring to this PR, this must be duly documented

DenisNosnitsin1970:
Sorry I confused, this deletion is not related to signature change, it is a mistaken record that should be removed (there will be a change elsewhere in MY, and I will open the issue, but it is a small-scale one, so no worries indeed).

eu-genia (approved):

PR #1676, opened 2022-05-04
BLorient572

thea-m:
@DenisNosnitsin1970 could you check whether the ruling pattern I gave for BLorient572 is correct?
(If it is correct: patterns ending in E are apparently not parsed by the site, see https://betamasaheft.eu/manuscripts/ESakm011/main / http://palaeographia.org/muzerelle/regGraph2.php?F=1A-1A-1A1A/0-0/0-0/E)


eu-genia (approved):

thea-m:
@DenisNosnitsin1970 is the ruling description now fine? Can it be merged?

PR #1679, opened 2022-05-05
Update BNFabb29.xml

DenisNosnitsin1970:
Minor corrections

### DIFF
--- a/ParisBNF/abb/BNFabb29.xml
+++ b/ParisBNF/abb/BNFabb29.xml
@@ -327,6 +327,14 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                          </list>
                                       </additions>
                                       </physDesc>
+                                      <history>
+                                         <origin>
+                                           <origDate notBefore="1700" notAfter="1799">
+                                           18th century.</origDate>
+                                           The spaces for the commissioner's name has been left blank on
+                                           <locus target="#13rb #16va #18va #20va"/>.
+                                            </origin>
+                                      </history>
                                     </msPart>
 
                             <msPart xml:id="p2">


ParisBNF/abb/BNFabb29.xml
```diff
@@ -335,13 +331,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                             <msPart xml:id="p2">
                         <msIdentifier/>
                         <msContents>
-                           <summary/>
```
eu-genia, line 338: I am always in favour of empty `summary `elements, this will allow the app to print the summary

ParisBNF/abb/BNFabb29.xml
```diff
@@ -214,7 +211,6 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                       <msIdentifier>
                 </msIdentifier>
                       <msContents>
-                         <summary/>
```
eu-genia, line 217: I am always in favour of empty `summary `elements, this will allow the app to print the summary

ParisBNF/abb/BNFabb29.xml
```diff
@@ -419,15 +413,17 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
   <msPart xml:id="p3">
         <msIdentifier/>
           <msContents>
-         <summary/>
```
eu-genia, line 422: I am always in favour of empty `summary `elements, this will allow the app to print the summary

eu-genia (changes requested):

ParisBNF/abb/BNFabb29.xml
DenisNosnitsin1970: If you look at the description now, what are the indications "extracted" and "given", under "Codicological Unit p1"
"Summary of codicological unit"? I dound them distracting. My intention was to remove them.

eu-genia: it is displayed correctly in the right column when there is history after physDesc (I have to explore why, but you can add a history element maybe)

DenisNosnitsin1970: mss?

eu-genia: I think it is very helpful when a summary is printed. So yes, you should always have a summary element under msContent. If there is text, then the given summary will be printed, if the element is empty, extracted summary will be produced. It is the TOC of the record.

eu-genia (approved):

PR #1680, opened 2022-05-06
Update EMML50.xml

thea-m:
Should the relation formerlyAlsoListedAs also be used here, even though the record BLorient819 never existed?

### DIFF
--- a/EMML/1-1000/EMML50.xml
+++ b/EMML/1-1000/EMML50.xml
@@ -322,7 +322,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
                            <listBibl type="catalogue">
                               <bibl>
                                  <ptr target="bm:EMML1"/>
-
+                                      <citedRange unit="page">54</citedRange>
                               </bibl>
                               <bibl><ptr target="bm:Wright1870Magdala"/><citedRange unit="page">614-615</citedRange></bibl>
                            </listBibl>
--- a/EMML/1-1000/EMML50.xml
+++ b/EMML/1-1000/EMML50.xml
@@ -39,7 +39,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
 
                   <altIdentifier>
 
-                    <idno from="1868" to="1872">BL Or 819</idno>
+                    <idno from="1868" to="1872">BL Oriental 819</idno>
 
                      </altIdentifier>
                </msIdentifier>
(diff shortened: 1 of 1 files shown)


DenisNosnitsin1970:
Unusual situation. Actually, also EMML 50 does not represent the true ms, it is a duplicate, the manuscript is held by an Entoto church. Orient 819 is not there any longer,  we can say that this shelfmark is only a part of the Orient 819 librarian history. I would not make a record, only explain in words.  Or, can we make a record for a manusript that is not there? On the other hand, making a stub-similar record with reference to another record would also be an alternative. This is an important volume, in itself, the one that was returned when Yohannes IV wrote to England saying smth like "my people will not obey me without this book".

eu-genia:
We have just decided, talking about EMIP, that we create no duplicates, or we need 2 records for all EMML, EMIP, Tanasee etc if we treat digital/mf separately. 
So let us stay with one record.
For EMML EMIP records where we have the real repository, the repository should be entered, just as it is done for EthioSPAre, EMML listed as collection.
If the ms is ever described as BLor then the relation may be added, otherwise it is simply to pit in the provenance, Entotto to Magdala to London and back.
In Chester Beatty there is Ladx Meux 3, it has no formerlyListed relation.

thea-m:
As far as I know, the manuscript is only listed as BL Or. 819 in Wright's article in the ZDMG, in the catalogue, it is only mentioned in a footnote as it had already been returned. I would then avoid adding the relation. I think that the provenance has already been well described by Anais.

EMML/1-1000/EMML50.xml
```diff
@@ -324,6 +324,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
                                  <ptr target="bm:EMML1"/>
```
eu-genia, line 325: EMML I - page (54)?

EMML/1-1000/EMML50.xml
```diff
@@ -324,6 +324,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
...
```
eu-genia, line 324: to line 294 above - I think the `placeName ref` to `LOC4506Manbar `should be replaced with `INS0101MadhaneAlam`

eu-genia (changes requested):

EMML/1-1000/EMML50.xml
thea-m: I don't know. Anais, who knows the manuscript better, specifically wrote "in Gondar", which would be LOC4506Manbar. On the other hand, the marginal note does indeed look like the typical one from Maqdala.

eu-genia (approved):

eu-genia:
https://github.com/BetaMasaheft/Documentation/issues/1280

PR #1687, opened 2022-05-10
deleted old date

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/2085


eu-genia (approved):
thank you, this is ok, once we are at it, then one could also clean out the `<p>` of the `availability`

thea-m:
This I don't manage to do in batch. If I delete `<p>` everywhere in `<licence>`, then I can't delete` </p> `afterwards, because all records are listed as errors (and the other way round, no difference). I remember that we tried the same thing once and for you it worked? I must be missing something.
image

PR #1688, opened 2022-05-11
Got D 782 - 852

DenisNosnitsin1970:
Stabs for further Dettenberg manuscripts.


SchleswigGottorf/GotD782.xml
```diff
@@ -0,0 +1,84 @@
...
+                    <term key="Prayers"/>
+                </keywords>
+            </textClass>
+            <langUsage>
+                <language ident="en">English</language>
+                <language ident="gez">Gǝʿǝz</language>
+            </langUsage>
+        </profileDesc>
+        <revisionDesc>
+            <change who="SD" when="2022-04-06">Created entity with rudimentary
```
eu-genia, line 74: i assume you are moving this ms from new? i do not see the deletion in the PR. you cannot create the same ms in a new directory without removing it from its old place

eu-genia (changes requested):

SchleswigGottorf/GotD782.xml
```diff
@@ -0,0 +1,84 @@
...
+                   <physDesc>
+                     <objectDesc form="Codex">
+                        <supportDesc>
+                           <support>
+                              <material key="parchment"/>
+                           </support>
+                            <foliation>The manuscript is not foliated.</foliation>
+                         </supportDesc>
+                     </objectDesc>
+                      </physDesc>
```
eu-genia, line 49: I was also wondering if Dettenberg should be mentioned in provenance

SchleswigGottorf/GotD782.xml
DenisNosnitsin1970: Indeed. Removed.

SchleswigGottorf/GotD782.xml
DenisNosnitsin1970: If it is acceptable, could you please merge?

eu-genia (approved):

PR #1695, opened 2022-05-13
Update BLorient751.xml

DariaElagina:
added a reference, since the contribution provides with a facsimile of the manuscript


thea-m (approved):
Thank you!

PR #1697, opened 2022-05-16
Dettenberg 91 - 100

DenisNosnitsin1970:
Dettenberg, further items.


DenisNosnitsin1970:
There is a change in shelfmarks from "D + number" to "SL Slg Dettenberg o.Nr. + number", valid for a part of the collection. I discussed the matter with the one responsibel for that matter.

eu-genia (approved):

PR #1714, opened 2022-05-27
Dr 22 05 wright codicology1

thea-m:


thea-m:
For BLorient480, I think the sewing stations are 4, correct?

eu-genia (approved):

thea-m:
I think I will not describe the sewing stations of the manuscripts that have been resewn, it's difficult for me to recognise whether they correspond to the original ones.

PR #1720, opened 2022-05-30
corrections

thea-m:


LondonBritishLibrary/add/BLadd19341.xml
```diff
@@ -69,7 +69,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         <objectDesc form="Codex">
                             <supportDesc>
                                 <support>
-                                    <material key="paper"/> wich bears <watermark>the date
+                                    <material key="paper"/> which bears <watermark>the date
```
eu-genia, line 72: the visualization at the moment does not support this sort of encoding, 
"paper" is printed as a button, not part of text;
"watermark" is printed as yes/no; date is not parsed

If this needs to be fixed a separate issue would be needed

LondonBritishLibrary/add/BLadd19341.xml
```diff
@@ -97,7 +97,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                 <desc>Neatly written, around <date>1810</date>.</desc>
                             </handNote>
                             <handNote xml:id="h2" script="Latin" corresp="#a1">
-                              <desc>Apparentely the handwritig of the traveller.</desc>
+                              <desc>Apparently the handwriting of the traveller.</desc>
                             </handNote>
                         </handDesc>
```
eu-genia, line 103: to line 109 below: qhat is encoded as `q `I guess should be `note` (Extracts...)

eu-genia (changes requested):

LondonBritishLibrary/add/BLadd19341.xml
thea-m: No, I think it is fine like this: 
image

thea-m:
Can this be merged?

eu-genia (approved):

PR #1721, opened 2022-05-30
Dr 22 05 wright codicology2

thea-m:


LondonBritishLibrary/orient/BLorient482.xml
```diff
@@ -156,9 +231,22 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         </list>
                      </additions>
                      
-                     
+                     <bindingDesc>
+                        <binding contemporary="false" xml:id="binding">
+                           <decoNote xml:id="b1">The manuscript has been resewn, and the original cover been replaced by 
+                              a wooden cover with blind-tooled leather quarter cover, very probably by
```
DenisNosnitsin1970, line 237: This is unclear. Boards and cover are different elements in terms of binding structure. I would summaries it like this:
"The manuscript has been rebound in the Library. Both original wooden boards have been replaced, and a blind-tooled leather quarter cover added (with the gilded title and shelfmark on the spine). Sewing stations are used in an irregular way."  I personally used to record the information on shelfmarks and stamps under "extras"

DenisNosnitsin1970 (approved):

LondonBritishLibrary/orient/BLorient486.xml
```diff
@@ -206,9 +213,11 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         </additions>
                         
                         <bindingDesc>
-                            <binding contemporary="true" xml:id="binding">
+                            <binding contemporary="partly" xml:id="binding">
                                 
-                                <decoNote xml:id="b0">In the original boards. </decoNote>
+                                <decoNote xml:id="b0">In the original boards according to Wright. Wooden boards with blind-tooled leather cover.
+                                The manuscript has apparently been resewn, it is also possible that the boards were replaced. The leather cover appears 
+                                to be original, but the spine has been imprinted in gold by the library with title, 'Brit. Mus,' and shelfmark.</decoNote>
```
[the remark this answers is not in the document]
thea-m: Thank you, yes, sorry for not updating the conversation. Denis said that he was going to open an issue on this.

LondonBritishLibrary/orient/BLorient485.xml
```diff
@@ -121,7 +246,12 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                             <origDate notBefore="1500" notAfter="1525" evidence="lettering">
                             </origDate>
                         </origin>
-                        <provenance>The manuscript had very probably belonged to a church in the <placeName ref="LOC3577Gondar"/> area, 
+                        <provenance>
+                            Several features indicate a possible Beta ʾƎsrāʾel origin of this manuscript (see  <bibl><ptr target="bm:DegeMueller2020BE"/><citedRange unit="page">18</citedRange></bibl>
+                            on the doxology on <locus target="#1r"/>, and note also the absence of trinitarian formulas integrated into the main text, and 
+                            the absence of a cross or other decoration on the quarter cover). It is probable that the manuscript changed owners repeatedly, before
+                            it came into the possession of <placeName ref="INS0101MadhaneAlam"/>.
```
DenisNosnitsin1970, line 253: A bit shaky. I assume, Old Testament books do not receive a trinitarian formula in CHristian manuscripts generally, unless the contrary is proven. :) Quarter cover must not be decorated with cross also at Christians. The only thing is the formula ይትባረክ፡, which is an added one, even though by the hand not too different from the main one. In my opinion, we could think only that the ms might have been in possession of Beta Esrael for some time.

LondonBritishLibrary/orient/BLorient487.xml
```diff
@@ -158,28 +165,24 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
-                        
-                        
                         <bindingDesc>
-                            <binding contemporary="true" xml:id="binding">
-                                
-                                <decoNote xml:id="b0">In the original boards. </decoNote>
+                            <binding contemporary="partly" xml:id="binding">
+                                <decoNote xml:id="b0">Wooden boards with leather quarter cover. According to Wright, the boards are original.
+                                    While this cannot be excluded, the back board looks very different from the front board and might be a
+                                substitution by the library. The leather quarter cover also does not seem to be original, but to have been added by the library,
```
DenisNosnitsin1970, line 175: This is very assumptive and at the end the use is lost. 
(After Wright a lot of time passed, at his time both boards could be original. We describe what we see. I assume many mss changed their appearance after having been catalogued, but we normally do not compare ).

"Wooden boards with leather quarter cover. The back board and the leather cover are unlikely original, 
probably added in the Library."

LondonBritishLibrary/orient/BLorient488.xml
```diff
@@ -230,34 +238,26 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
                             </list>
                         </additions>
                         
                         
                         
                         <bindingDesc>
                             <binding contemporary="true" xml:id="binding">
-                                
-                                <decoNote xml:id="b0">In the original boards lined with striped silk and covered with blind-tooled leather. </decoNote>
+                                <decoNote xml:id="b0">In the original boards, with textile inlays of striped silk and covered with blind-tooled leather.
```
DenisNosnitsin1970, line 249: Here is a style deviation. I usually would use, as above and below: Two wooden boards, or just wooden boards, as above and below. Otherwise we need to add a not on originality/non originality to many elements.

LondonBritishLibrary/orient/BLorient489.xml
```diff
@@ -240,14 +248,15 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         
                         
                         <bindingDesc>
-                            <binding contemporary="true" xml:id="binding">
+                            <binding contemporary="partly" xml:id="binding">
                                 
-                                <decoNote xml:id="b0">In the original boards. </decoNote>
+                                <decoNote xml:id="b0">Wooden boards with leather quarter cover. The manuscript might have been rebound by the library.
```
DenisNosnitsin1970, line 253: Above, there is no formula "The manuscript might have been...." whereas some of them - if not all - definitely are. Should this fact be indicated better through "Restoration modern"?

DenisNosnitsin1970 (changes requested):

LondonBritishLibrary/orient/BLorient485.xml
thea-m: This is exactly what the article argues.

eu-genia: Denis argues against the BE origin; he says it may have been in BE possession for some time - after its origin.

thea-m: So does Sophia. Now I see that my use of the word "origin" here was not very exact, I had not meant the production. I will change the formulation. Thank you!

eu-genia: replace it with "connection" )

LondonBritishLibrary/orient/BLorient489.xml
thea-m: "The manuscript has been rebound in the Library" has been suggested by you for BLorient482 above, I used it also for other manuscripts in this PR, with different degrees of certainty. I can add  "Restoration modern" in the relatively certain cases, but I would also include a narrative description, because "Restoration modern" could mean many things.

LondonBritishLibrary/orient/BLorient488.xml
thea-m: Good, "original" is already covered by "contemporary="true"" (Wright's formulation, hence the style deviation).

LondonBritishLibrary/orient/BLorient487.xml
thea-m: I agree with the proposed formulation, but I do think that the difference from the state of the manuscript at the time of Wright's description is an important information, which would be lost on those who would not open the catalogue in parallel (to make that a necessity for reading our descriptions would undermine our project a bit, I think).
How about something like:
"Wooden boards with leather quarter cover. The back board and the leather cover are unlikely original, probably added in the Library. Wright states that the boards are original."

DenisNosnitsin1970: It is ok, buy Wright seems to have done this very irregularly, isnt it? And it is unclear why in this case he says this. 
 The difficulty that i I see is that, if you say "Wright states...", you need a couple of arguments as to why what he says is wrong. He had the book in his hands, you did not. Perhaps, you could style it like:  "...Library (Wright, on the contrary, states that both boards are original)."

DenisNosnitsin1970: All together, could be like this: 
Wooden boards with leather quarter cover. The back board looks very different from the front board and might have been furnished in the Library (Wright, on the contrary, states that both boards are original). The leather quarter cover also does not seem to be original, possibly supplied together with the back board.

LondonBritishLibrary/orient/BLorient489.xml
thea-m: Now is the moment to discuss it, as it has not yet been frequently used :) Do you find the formulation problematic in general or is its use here irregular?

LondonBritishLibrary/orient/BLorient487.xml
thea-m: My hypothesis is that he stated this for the original bindings and those he stays silent on were rebound already at his time, but it is really not very certain. I think your solution is good.

eu-genia (approved):

PR #1723, opened 2022-05-30
Update EMML7028.xml

MarcinKrawczuk:


EMML/7001-8000/EMML7028.xml
```diff
@@ -31,16 +31,15 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
          <sourceDesc>
             <msDesc xml:id="ms">
                <msIdentifier>
-                  <repository ref="INS0004HMML"/>
```
eu-genia, line 34: why no repository? we replace hmml with true repositories for mss when we have them, but otherwise i don't think it should be removed

eu-genia (changes requested):

MarcinKrawczuk:
There's no INS... record, I would have to create it

pon., 30 maj 2022 o 22:18 Eugenia ***@***.***> napisał(a):

eu-genia:
which should be the repository?

eu-genia:
( I meant just to leave HMML if there is nothing to replace it with)

MarcinKrawczuk:
LOC2997Esatan

pon., 30 maj 2022 o 22:54 Eugenia ***@***.***> napisał(a):

PR #1725, opened 2022-06-01
updates

thea-m:


LondonBritishLibrary/orient/BLorient500.xml
```diff
@@ -69,17 +69,78 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                     
                                 </support>
                                 <extent>
-                                    <measure unit="leaf" quantity="68">68</measure>
+                                    <measure unit="leaf" quantity="68">68+2</measure>
```
eu-genia, line 72: should it then become 70? just asking

LondonBritishLibrary/orient/BLorient501.xml
```diff
@@ -175,11 +175,126 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+                                <condition key="good">Stained by water and otherwise slightly damaged.</condition>
                             </supportDesc>
                             <layoutDesc>
-                                <layout columns="2" writtenLines="28 40"/>
+                                <layout columns="2" writtenLines="28 40">
+                                    <ab type="pricking">Pricking and ruling are visible.</ab>
+                                    <ab type="pricking">Primary pricks are visible, horizontal pricks are visible.</ab>
+                                    <ab type="ruling" subtype="pattern">1A-1A-1A1A/0-0/0-0/C</ab>
+                                    <ab type="ruling"> The upper line is written above the ruling,
+                                        the lower line is written above the ruling.</ab>
```
eu-genia, line 296: I guess crux ansata is usually part of layout (?)
`<ab type="CruxAnsata"></ab>
<ab type="coronis"></ab>`

LondonBritishLibrary/orient/BLorient506.xml
```diff
@@ -79,17 +79,25 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                     
                                 </support>
                                 <extent>
-                                    <measure unit="leaf" quantity="174">174</measure>
+                                    <measure unit="leaf" quantity="174">174+1</measure>
```
eu-genia, line 82: 175?

LondonBritishLibrary/orient/BLorient509.xml
```diff
@@ -132,17 +132,27 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                     <material key="parchment"/>
                                 </support>
                                 <extent>
-                                    <measure unit="leaf">160+IV</measure>
+                                    <measure unit="leaf">2+160+2</measure>
```
eu-genia, line 135: quantity="164"?

eu-genia (approved):

LondonBritishLibrary/orient/BLorient500.xml
thea-m: the value of `@quantity`, you mean? Yes, that would make sense. I had actually never really paid attention to `@quantity` in `<measure>` (it's only used in these old BL records and some Ethio-SPaRe records, and not present in any example in the guidelines). Would it be useful to systematically add quantity in all records?
(This old BL foliation system and the numbering of the digital images is really unfortunate...)

eu-genia: in E-S records the numbers usually add up ) this is why i asked

eu-genia: But this is again the question what is counted and where; the E-S foliated starting from 1 from the first available page, so from flyleaves, whereas others usually do not number flyleaves, and describe them as part of binding. So am not really sure what the rule should be. Just numbers not adding up are confusing.

eu-genia: I think quantity is good when it is there, one could then filter for mss with e g more than 100 leaves or less than 20 etc. things you cannot do if you do not have a clear numeric value for an attribute.

LondonBritishLibrary/orient/BLorient506.xml
DenisNosnitsin1970: I used to do 175 (174+1)

LondonBritishLibrary/orient/BLorient501.xml
DenisNosnitsin1970: Is it a part of layout?

eu-genia (approved):

LondonBritishLibrary/orient/BLorient500.xml
thea-m: I've added it here, but again, this has been used practically nowhere (only a few ES manuscripts and Zhenia's old BL descriptions). If it is so important, it should be included in the guidelines. Filtering according to leaf number seems not to be possible right now? But maybe I missed it.

PR #1731, opened 2022-06-07
Dr 22 06 wright codicology7

thea-m:
digitise or digitize (and derivates)? I know we agreed to British spelling, but I feel the version with z is really much more common. I don't mind either way, but I could at one point quickly standardise it to one version in all these records.


eu-genia:
It should be ize, this is the preferred spelling in the Oxford English Dictionary

PR #1751, opened 2022-06-23
Genova Sapeto

eu-genia:
stubs from bm:Britaetc2017ThreeCollections


thea-m (approved):
Thank you! Everything looks fine. Are the authors of the inventory informed about the creation of these stubs?

eu-genia:
I have been asked by @abausi to create the stubs on the basis of the handlist in order to be able to display the images; I hope that @antobrita @SusanneHummel @MassimoVilla are ok with it.

thea-m:
I ask because there is another long-term unmerged branch in which work is carried out on the collections described in the inventory, and we need to avoid duplicates and other confusion. As far as I could see, no records exist yet for the Sapeto collection, but maybe you could inform the authors also by mail (GitHub notifications might end up in spam...), so that no confusion arrises later on?

abausi:
Relevant people were informed; we can proceed. No need of GitHub queries

Il 23.06.2022 14:16, thea-m ha scritto:

PR #1752, opened 2022-06-23
Dr 22 06 wright codicology23

thea-m:


thea-m:
Any idea about the dots in BLorient823? There are some, maybe similar, but to a lesser extent, in BLorient826 as well.

PR #1753, opened 2022-06-24
facs added

eu-genia:
@DenisNosnitsin1970 per email
please insert images for
image

@DenisNosnitsin1970 - in order to insert the images as always it is enough to add the facs element with the correct path to images, see the PR
e.g. for DD-036
`<idno facs="DD/036/DD-036" n="157">DD-036</idno>`


DenisNosnitsin1970:
Ok, understood! Thanks.

PR #1754, opened 2022-06-24
Dr 22 06 wright codicology24

thea-m:


LondonBritishLibrary/add/BLadd18993.xml
```diff
@@ -104,18 +104,27 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                     <material key="parchment"/>
                                 </support>
                                 <extent>
-                                    <measure unit="leaf">49</measure>
+                                    <measure unit="leaf" quantity="57">4+1+49+3</measure>
```
DenisNosnitsin1970, line 107: Why is here the string 4+1+49+3? Not 5+49+3?

DenisNosnitsin1970 (changes requested):

LondonBritishLibrary/add/BLadd18993.xml
thea-m: i-iv are modern flyleaves, v probably originally belonged to the manuscript. But it can be changed, in these records this notation mostly reflects the foliation, not the "real" flyleaves.

DenisNosnitsin1970: I would say, it reflect the foliation first (breaks within the main block are not reflected by foliation, too).

thea-m: so better just 5+49+3, you mean?

thea-m: @DenisNosnitsin1970 ? :)

PR #1755, opened 2022-06-27
Update BLorient544.xml

thea-m:
Not completely clear, but hopefully an improvement from Wright. I'm not completely certain whether 112-113 are a bifolium (would fit into the quire structure, but not the textual division, and is not so well visible to me).


DenisNosnitsin1970 (approved):

DenisNosnitsin1970:
Not well visible, true.

eu-genia (approved):

eu-genia:
(to me they seem a bifolium, numbered as a quire...)

PR #1756, opened 2022-06-27
B lorient8192 part2

MarcinKrawczuk:
So this was merged, but it was a "work-in-progress" merge. It would be great if you could do a proper check this time. 

One thing I cannot understand that if you take any work record that appeats in this ms (e.g. https://betamasaheft.eu/works/LIT6519DersanFasika/main) and then press "See these ...  manuscripts in the list view", it does not display correctly.

Thanks very much!


thea-m:
It must have been a syncing issue. Now everything seems to be displayed properly.

PR #1766, opened 2022-07-05
added binding

thea-m:


LondonBritishLibrary/add/BLadd18993.xml
```diff
@@ -160,6 +160,21 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                 </item>
                             </list>
                         </additions>
+                        
+                        <bindingDesc>
+                            <binding contemporary="false" xml:id="binding">
+                                <decoNote xml:id="b1" type="Boards">Two boards.</decoNote>
```
DenisNosnitsin1970, line 166: What is the material of the boards?

LondonBritishLibrary/add/BLadd19660.xml
```diff
@@ -160,6 +160,19 @@ schematypens="http://relaxng.org/ns/structure/1.0"
 
                             </list>
                         </additions>
+                        
+                        <bindingDesc>
+                            <binding contemporary="false" xml:id="binding">
+                                <decoNote xml:id="b1" type="Boards">Two boards.</decoNote>
```
DenisNosnitsin1970, line 166: What is the material of the boards, here and below?

DenisNosnitsin1970 (changes requested):

LondonBritishLibrary/add/BLadd18993.xml
thea-m: I don't know, from what I could see from the images they're completely covered. Could you have a look?

DenisNosnitsin1970: Indeed! Difficult.

thea-m: Can it be merged like this, or would you suggest a different formulation?

DenisNosnitsin1970: If we do not want to make it complicated, it can be left as it is. If we do, perhaps: "Two boards (of wood or pasteboard)."

PR #1776, opened 2022-07-11
Update BLorient637.xml

thea-m:


LondonBritishLibrary/orient/BLorient637.xml
```diff
@@ -50,9 +50,116 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+                                <condition key="good"/>
                             </supportDesc>
                         </objectDesc>
                         
+                        <additions>
+                            <list>
+                                <item xml:id="a1">
+                                    <locus target="#2r"/>
+                                    <desc type="LandGrant">Deeds of gifts, lists of furniture, books, etc., of the reigns of <persName ref="PRS9153TaklaHa"/>, <persName ref="PRS9142TaklaGi"/>
+                                        and <persName ref="PRS5618IyasuII"/>.</desc>
```
DenisNosnitsin1970, line 150: I would say: Three records concerning land possessions and gifts, from the time of...., in two crude and one calligraphic hands. (unless you do not want to treat them separately)

PR #1785, opened 2022-07-21
Feature: CI

line-o:
validate TEI and build XAR


duncdrum (approved):
loks good i like the subshell

PR #1818, opened 2022-08-25
Update BNFabb114.xml

GueshSol:


thea-m (approved):
Very good work, Guesh! Technically, everything is encoded correctly.
There were a few mistakes in the folio numbers. I have corrected them now, in the future, I think it will be good if you double check with the manuscript after finishing the description.
Also note my corrections to ms_i1.29 and ms_i1.30: You had correctly identified the irregularity (maybe establishing the quire structure will help to understand how it happened), I have made the encoding more precise by using two locus elements.

PR #1820, opened 2022-08-25
validation fixes BNF1

eu-genia:
there are several general things I noted:
the schema does not allow two `title `elements in the same `msItem`
many records have an empty `langUsage` (have corrected for now, could be easier if we find a way to tell the schema that empty placeholder elements are OK)

important:
the div `edition `does not allow `locus`, the page reference is expected to only be in `teiHeader`

I have tried to work around these but maybe this could be another discussion for the schema and rules

@thea-m I have not touched BNFet53 with many validation errors in the apparatus, please let me know if any schema modifications are necessary to make it work, or if there are issues with the data, thank you!

### DIFF
--- a/ParisBNF/abb/BNFabb152.xml
+++ b/ParisBNF/abb/BNFabb152.xml
@@ -71,9 +71,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                     
                       <locus from="58v" to="71">Fol. 58v-71</locus>
                      
-                     <title xml:lang="gez">Tarika nagaśt za-gabru la-adbār za-qeddusān</title>
-                     <title xml:lang="fra">Histoire des Rois concernant les impôts des monastères
-                saints</title>
+                     <title xml:lang="gez">Tarika nagaśt za-gabru la-adbār za-qeddusān (Histoire des Rois concernant les impôts des monastères
+                saints)</title>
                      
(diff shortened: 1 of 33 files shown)


thea-m:
Thank you!
If I remember correctly (no guarantee :) ), most of this has historical explanations and should not require further schema adjustments:
- The emtpy langUsage element was the result of an automatic correction carried out by Pietro, because there were many records that had copy pasted langUsage elements that actually did not refer to languages used in the records, which cluttered the  search results. So I would be against reintroducing such elements (e.g. BNFet11 has no French anywhere). I can go through all of them now and update according to what is present in the records.
- div edition does not allow locus: This was also the result of a discussion (whose rationale I forgot), which left old records encoded before this decision invalid. The alternative is `<pb>`, I can also update this in BnF records.

BNFet53 was somewhat ambitious encoding from the beginning of the project that was not further explored...Could we just delete all loc attributes to make it valid?

eu-genia:
feel free to play around ) I have not gone through the rest of BnF yet, was planning after these questions are solved.
I have not found a way to tell the schema that we do not mind empty elements yet, this would solve a whole lot of issues

We also have a problem, maybe there were too many pushes and since yesterday night they are not truly going through to the db, so let us abstain from merging more for now, I will see Monday if the issue has resolved itself, or will have to switch gitsync off and update db manually (very annoying)
image

thea-m (approved):

PR #1830, opened 2022-08-30
Create BSLet175.xml

karljonaskarlsson:


thea-m (approved):

LondonBibleSociety/BSLet175.xml
```diff
@@ -0,0 +1,81 @@
...
+            <p>Encoded according to TEI P5 Guidelines.</p>
+         </projectDesc>
+         <xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="https://raw.githubusercontent.com/BetaMasaheft/Documentation/master/prefixDef.xml">
+            <xi:fallback>
+               <p>Definitions of prefixes used.</p>
+            </xi:fallback>
+         </xi:include>
+      </encodingDesc>
+      <profileDesc>
+         <langUsage><language ident="en">English</language></langUsage>
```
SophiaD-M, line 69: add also Geez?

thea-m: It is not technically necessary, because they language has not been used in the record. Here, all languages used in the record are defined. But since we assume that the manuscript is written in Geez and will be described later, it won't be wrong (adding languages here that are not used in the manuscript or record is confusing the search results.)

karljonaskarlsson: Ah, good to know! :)

PR #1832, opened 2022-08-30
Create BSLet177.xml

karljonaskarlsson:


thea-m (approved):

LondonBibleSociety/BSLet177.xml
```diff
@@ -0,0 +1,90 @@
...
+                     </altIdentifier>
+                     <altIdentifier>
+                        <idno>Cowley IV</idno>
+                     </altIdentifier>
+                  </msIdentifier>
+               <msContents>
+                  <summary/>
+                  <msItem xml:id="ms_i1">
+                     <title ref="LIT1961Mashaf"/>
+                     <textLang mainLang="gez"/>
```
SophiaD-M, line 42: above in title statement you wrote "Sa'atat with Psalms of David", should there then here be two points for msContent?

karljonaskarlsson: I was thinking that the Psalms of David are an integral part of this Sa'atat

SophiaD-M: I think this is an open research question

karljonaskarlsson: I agree, and I wouldn't add the Psalter as a msContent

karljonaskarlsson: To clarify: The Psalter is not to be found from the beginning to the end, but there are different Sa'atat prayers in between. So it is quite well integrated into the manuscript.

PR #1839, opened 2022-09-01
Dr 22 08 provenance

thea-m:
See https://archive.org/details/quaritchcatalogue130/page/n21/mode/2up
(I also find that BDLaethf31 would match the "Gospel of John" on p. 13 - but that would only make sense if the manuscript had been offered for sale while still part of the Phillipps estate, never been sold, and then been sold later with the a part of the Phillipps collection at Sotheby's - that seems a bit far-fetched, I guess)


MassimoVilla:
Very good Thea, thank you! Concerning BDLaethf31, it might indeed be the manuscript described as Gospel of John (the binding features and the "morocco case" also match), although, in the absence of further information, any conjecture about its history seems speculative.
Perhaps a short remark on this possibility is not inappropriate. Others' opinion?

MassimoVilla (approved):

karljonaskarlsson (approved):

thea-m:
added a remark in f. 31

ElianaDalSasso (commented):
All good! As for the binding description in Quaritch 1893, I think that it is so generic that it could match with good number of manuscripts. However, the coincidence with other factors makes it worth noting it!

PR #1843, opened 2022-09-05
ci: fix push workflow

line-o:
Gathering all source files in a separate step will lead to the
following validation step to exit with code 1 because the amount
of file names that will be pasted into the command exceeds ARG_MAX.

While not as readable as in two separate steps it is of greater
value to combine both steps and have the find command be executed
in a subshell.


line-o:
Do not merge yet. The server is still crunching on the last push to master.

duncdrum (approved):
fix does what its supposed to. minor edit for better interop with different shells

PR #1859, opened 2022-09-12
ci: exclude fragment files from validation

line-o:


duncdrum (approved):
there is still the option to bring these in line with other fragment files, which are wrapped inside a full tei file wrapper and then the `xinclude` is pointing at the element in question instead of the full file. 

I found quite a number of fragment files which don't seem to be included anywhere, so that would give us a greater degree of certainty. 

I ll have another look tomorrow before merging.

eu-genia:
now after 
i only see
```
./Ham/DabraLibanosHamGG1/edition.xml fails to validate
./Ham/DabraLibanosHamGG1/msDesc.xml fails to validate
./EMML/6001-7000/EMML6952/EMML6952text.xml fails to validate
./EMML/4001-5000/EMML4398/EMML4398text.xml fails to validate
./ES/ESqdq004/ESqdq004text.xml fails to validate
```
DabraLibanosHamGG1 will be fixed with this PR; `ESdqd004 ``EMML6952 ``EMML4398 `seems to complain about `<space reason="rubrication">` but it is present in the schema as far as I can see

.github/workflows/validate_push.yml
```diff
@@ -32,6 +32,8 @@ jobs:
           find . -type f -name '*.xml' \
             -not -name 'facsimile.xml' \
             -not -name 'transkribus*.xml' \
+            -not -name 'msDesc.xml' \
+            -not -name 'editions.xml' \
```
duncdrum, line 36: typo `edition.xml`

duncdrum:
@eu-genia I get no errors about `<space>` the `*text.xml` fragments when tested as standalone files are invalid, which is to be expected. This should now be all green. Open PRs should be rebased `git rebase master` once this is merged.

PR #1866, opened 2022-09-23
Emip metadata tweed scrolls 20 june 2022

Ralph-Lee-UK:
This is the first set of completed edits that we have done to include Transkribus output. There are no errors in the editor, but there still may be problems with the way we have coded these. All the edits are pretty much identical, so if you could check these, and send back any corrections! We'll make corrections asap.


thea-m (changes requested):
Thank you! That is great. I think these will make a big difference in identifying the textual units contained in scrolls.
Just one question: each of the folders with the extended files contains a file called for example TweedMagicScroll002_tei.xml with both the facsimile and the transcribed text. Are these necessary or are they a product of the workflow?
Before merging, I would let Eugenia also have a look at it.

Ralph-Lee-UK:
@thea-m many thanks. The files like TweedMagicScroll002_tei.xml are the output from Transkribus before the transformation - so indeed they can be deleted

eu-genia:
1) The "merged" _tei files should be of course removed, or on merging doublets with the same `xml:id `will be created on the app, and nothing will show

2) your "main" files (e.g. `EMIPms00551\EMIPms00551.xml`)
contain 
```
<text xml:base="https://betamasaheft.eu/">
      <body>
...
</body>
</text>
```
but your text files (e.g. `EMIPms00551\transkribusTextEMIPms00551.xml`)
are also nested in 
```
<text xmlns="http://www.tei-c.org/ns/1.0" xml:base="https://betamasaheft.eu/">
   <body>
...
</body>
</text>
```
on merging, this would create double text-body sequence.

the main file must not have the text-body at all, just `xi:include` after the end of `teiHeader` as in
```
 </teiHeader>
    <xi:include xmlns:xi="http://www.w3.org/2001/XInclude" 
        href="facsimile.xml">
        <xi:fallback>
            <!-- facsimile -->
        </xi:fallback>
    </xi:include>
 
            <xi:include xmlns:xi="http://www.w3.org/2001/XInclude" 
                href="transkribusText.xml">
                <xi:fallback>
                    <!-- transkribus text -->
                </xi:fallback>
            </xi:include>
```
</TEI>

3) can you please resolve the conflicts?

Ralph-Lee-UK:
@eu-genia thanks for your comments, all clear. I believe that I have made the changes requested!

eu-genia:
Thank you! I still see the body-text at least in EMIPms00470

eu-genia:
There are quite a few conflicts that I cannot resolve easily; I am moving changes to several new branches to try to resolve them in that way

eu-genia:
separate PR were created (new branches from master to avoid conflicts), I am closing this PR without merging, separate PRs should be checked and can be then merged

for the future it would be very helpful if single PR for review does not contain more than 25 files at a time (difficult to review and difficult to sync), thank you very much!

thea-m:
Thanks a lot, @eu-genia !

PR #1878, opened 2022-09-28
EMIP Facs to Bodleian MSS

eu-genia:
facs added


thea-m (approved):

thea-m:
I'm sorry, I was not completely attentive yesterday: Could you (or me, if you can give me a good text here) add a change element to the files that states the source of the images? I remember the library being very strict about publishing images of its manuscripts. I assume that EMIP checked that we have the right to do so?

eu-genia:
you are right, I have now cross checked with Steve and no, we are not supposed to make them available, just keep image copies... I wll comment out the facs, so that only we know that we have the photos (i guess it is still good to know for us that we have them here)

thea-m:
Thank you!

PR #1896, opened 2022-10-10
Bn fabb110

MarcinKrawczuk:
I am creating a stub because I need it for witness list. @DenisNosnitsin1970 @thea-m


thea-m (commented):
Thanks!

thea-m (approved):

eu-genia:
Dear Marcin,
please remember - if you create a record through the app then you MUST download it to the `NEW `folder on your computer, then create a PR folder to sync the github and the app, and only then, after the merge, move the file to the actual repository folder. I know I have already written that, but I see now once again the PR was created from the repository folder and as a result the record was present twice (when you create a record in the app it is automatically created in the `NEW `folder online, and if this is not synced, then on merge it is uploaded to the repository folder AND remains in the `new` folder, and two records with the same ID may not be displayed and result in an error). I have now removed the second `BNFabb110 `copy manually from the backend.
I can see also in the `new `folder on the app a lot of BAV manuscripts that have not been synced yet. They too must now first be synced with the `new `folder and only then moved to the `VaticanBAV `folder and repushed. 
For manuscripts, it may be simpler to create records/stubs without using the app, simply create a file in your XML editor (by using the template from another file) and save it directly where it should be.
Many many thanks!
Zhenia

PR #1902, opened 2022-10-12
Ms 2002.4 guesh

GueshSol:


new/CIA20024.xml
```diff
@@ -0,0 +1,927 @@
...
+                                    <textLang mainLang="gez"/>
+                                </msItem>
+                                
+                                
+                                <msItem xml:id="ms_i2.2.32.3">
+                                    <locus target="#139r"/>
+                                    <title type="complete" ref="LIT5673AkkonuColl#AkkMin32"/>
+                                    <textLang mainLang="gez"/>
+                                </msItem>
+                                <msItem xml:id="ms_i2.2.33">
```
thea-m, line 883: This item is included in ms_i2.2.32, but numbered ms_i2.2.33: It should either be numbered ms_i2.2.32.4 or be moved outside of ms_i2.2.32. Both solutions would be fine, you have to decide which is conceptually better.

thea-m (changes requested):
I think that everything is fine - I did not double check the IDs now, but all seems good. Only one remark in the last item. Thank you!

thea-m (approved):

thea-m:
@GueshSol note that I now moved this manuscript from the folder "new" to the folder "ChicagoIA"

GueshSol:
Thank you very much.

PR #1904, opened 2022-10-17
moved to folders

thea-m:


eu-genia:
Dear @thea-m note that you must always save to a new folder and delete in the old one. Unfortunately an action like this
image
results in this
image
meaning that the github does not know the old record has to be deleted and the doublets may cause multiple errors, for individual mss but also for all records connected to them
note also that the folder `/new` has disappeared completely now from the `manuscripts `repo, it should be recreated, with MDDL001 in it I guess as that is where I see it on the backend but not in github
Thank you!

eu-genia:
(I hope with the new syncing mode on that @line-o and @duncdrum will be setting up this problem won't exist any more but for the moment it does - and at the moment I do not know of the way to quickly search for items in the app with the same xml:id to quickly fix possible problems before they disrupt performance)

thea-m:
Thanks! I was not aware of that. Could you maybe update the guidelines with information on this? I feel that many people are confused with record creation/deletion?

PR #1906, opened 2022-10-19
Binding desc bd laethe33

ElianaDalSasso:
Hi everyone! Here is a brief description for the poorly preserved binding of Aeth. e 33.
I updated also the condition.

### DIFF
--- a/OxfordBodleian/Juel-Jensen/BDLaethe33.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethe33.xml
@@ -352,16 +352,18 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         </additions>
 
                         <bindingDesc>
-                            <binding contemporary="true" xml:id="binding">
+                            <binding contemporary="partly" xml:id="binding">
                                 <decoNote xml:id="b1" type="bindingMaterial">
                                   The manuscript is bound in <material key="wood">wooden</material> boards.
                                 </decoNote>
                                 <decoNote xml:id="b2" type="Boards">
-                                  The boards are cut to the same size as the bookblock and are externally bevelled. The grain runs vertically. On the inner sides the marks left by the turn-ins of a previous leather cover are still visible.
+                                  The boards are cut to the same size as the bookblock and are externally bevelled. The grain runs vertically.
+                                  Marks left by the previous leather cover turn-ins are still visible on the inner sides of the boards.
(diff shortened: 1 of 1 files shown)


thea-m (approved):
Thank you, looks great! Just a question for my better understanding: you write "Since the repair was carried out by stabbing the leaves along the inner margin". What does stabbing mean here, how is it different from sewing?

MassimoVilla (approved):

gnizla:
> Thank you, looks great! Just a question for my better understanding: you write "Since the repair was carried out by stabbing the leaves along the inner margin". What does stabbing mean here, how is it different from sewing?

In my ignorance I also don't know what "stabbing" means

ElianaDalSasso:
@thea-m @gnizla well spotted! So, sewing indicates the passage of thread through the fold of the quires, stabbing (or stitching) indicates that the thread does not pass through the fold but pierces, 'stabs', the single  leaves or the quires through the inner margin.  
Perhaps Ligatus gives a more precise explanation: 
- stitching http://w3id.org/lob/concept/1648
- sewing http://w3id.org/lob/concept/2362

Do you think that stitching would be more appropriate? I used stabbing just as a habit but I don't have any preference. The terms are equivalent.

gnizla:
> @thea-m @gnizla well spotted! So, sewing indicates the passage of thread through the fold of the quires, stabbing (or stitching) indicates that the thread does not pass through the fold but pierces, 'stabs', the single leaves or the quires through the inner margin. Perhaps Ligatus gives a more precise explanation:
> 
> * stitching http://w3id.org/lob/concept/1648
> * sewing http://w3id.org/lob/concept/2362
> 
> Do you think that stitching would be more appropriate? I used stabbing just as a habit but I don't have any preference. The terms are equivalent.

TA! :)

thea-m:
Thank you! @ElianaDalSasso you are in the best position to chose the binding terminology.

karljonaskarlsson:
Thank you, @ElianaDalSasso ! "Leather case" means māḫdar? Maybe one could give the Ethiopic term (which at least to me seems common also in European languages) in parenthesis?

ElianaDalSasso:
Thank Jonas!I remember that the encoding of the māḫdar has been discussed in an old issue, see #970. There, it was advised to not use an Ethiopic term

thea-m:
(This was the issue https://github.com/BetaMasaheft/Documentation/issues/970)

PR #1907, opened 2022-10-19
Emml1763 corr mk 11 aug22

MarcinKrawczuk:
I have added the missing work records, expanded additions a bit.


EMML/1001-2000/EMML1763.xml
```diff
@@ -953,8 +953,21 @@
...
-<locus target="1r"/>
-<desc type="GuestText">Ten lines from the introduction to the Synaxary.</desc>
+<locus target="#1r"/>
+<desc type="GuestText">Ten lines from the introduction to the <ref type="work" corresp="LIT2375Synaxa">Synaxary</ref>.</desc>
+</item>
+
+<item xml:id="a2">
+<locus target="#279r"/>
+<desc type="ScribalSignature"></desc>
+<q xml:lang="gez">ወጸሐፍክዎ፡ ለዝንቱ፡ ድርሳነ፡ አነ፡ መርቆርዮስ፡ ነዳይ፡ ኃጥእ። ጸልዩ፡ ሊተ፡ እለ፡ አንበብክምዎ፡ ለዝንቱ፡ መጽሐፍ። ኵልኩሙ፡ ቅዱሳን፡
```
thea-m, line 963: My impression from a superficial look at the manuscript is that this is not really a secondary addition, but a part of the main production unit, probably written by the main hand. If you agree, a better place to encode this information would be in `<colophon>`.
(I looked at it only briefly, if you think it really is an addition, it can stay here)

thea-m (changes requested):
Great, thanks! One question about the colophon.

MarcinKrawczuk:
You're absolutely right, it's just I was so hard focused on work records that I disregarded everything else. There's a lot of interesting things that should be added to the ms record and I hope to do it at some point.

MarcinKrawczuk:
I think it's OK (or at least more OK) now

thea-m (approved):
Thank you!

PR #1910, opened 2022-10-24
Ba vet151 corr mk

MarcinKrawczuk:
Some of the Vat mss. have been approved and merged without being complete, so I will be correcting them one by one. 

The catalogue gives this info: Numeri vero conspiciuntur tantum 5-9 in principio fasciculorum 3-7, I don't understand what they mean so I have omitted it.


VaticanBAV/et/BAVet151.xml
```diff
@@ -89,16 +91,37 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
 </item>
 
+<item xml:id="e2">
+  <desc>The name of the owner has been erased. On <locus target="#46v">f. 46v</locus> and <locus target="#58v">58v</locus>
+  that of <persName role="owner" ref="PRS13932Walda">Walda Ṣādǝq</persName> has been written.
+</desc>
+</item>
+
+<item xml:id="e3">
+<desc></desc>
```
thea-m, line 115: e3 is empty - if you plan to insert something later, I would just comment it out for the time being.

VaticanBAV/et/BAVet151.xml
```diff
@@ -89,16 +91,37 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
 
 <list>
 
+<item xml:id="a1">
```
thea-m, line 94: If you know that these are later additions, everything is fine, but if they might also have been included during the original production of the manuscript, I would change the addition type to e.

thea-m (changes requested):
Thanks! One comment and one question, but all is fine

MarcinKrawczuk:
I've moved them to e, seems more likely

wt., 25 paź 2022 o 12:46 thea-m ***@***.***> napisał(a):

thea-m (approved):

PR #1916, opened 2022-10-27
Update BerOrFol117.xml

eu-genia:


eu-genia:
it is important that all files validate so i went ahead with this one, hope it is ok

PR #1935, opened 2022-11-30
fixing Miracles of Mary

eu-genia:
I wonder whether in these cases and also in some others here
https://betamasaheft.eu/xpath.html?xpath=collection%28%24config%3Adata-rootMS%29%2F%2Ft%3Atitle%5B%40ref%5Bcontains%28.%2C%27LIT3993Miracle%27%29%5D%5D
(e.g. BDLaethf1) the single MM should be still grouped under the collection ID...


thea-m:
According to the definitions we came up with a while ago you are right, they could be grouped. The appropriate ID would be LIT3585Taamme (without introductory texts)

thea-m (approved):

PR #1940, opened 2022-12-02
Leicester210bis

vitagrazia81:
I have done a few corrections after I have seen the MS on line. Thanks again.


thea-m:
Vita, because I did a layout change when we experimented it seems now like there are conflicts. I reverted my commit and will close this PR, can you make a new one?

vitagrazia81:
It should be without conflicts now, I guess.

PR #1944, opened 2022-12-08
Guesh 2002.4 07 22

GueshSol:


thea-m (changes requested):
Very good, Guesh! Just a few minor remarks. I also carried out a few linguistic corrections directly inside the file.

thea-m (approved):

PR #1947, opened 2022-12-12
Binding desc 2/5

ElianaDalSasso:
Hi! Here are the last descriptions of bindings with wooden boards. Two of them with a slip case.
Sometimes I found myself repeating what's in the condition in the binding description. For example in the case of a broken board.
Do you have any suggestion to avoid it? Or how to regulate it?
Thanks!

### DIFF
--- a/OxfordBodleian/Juel-Jensen/BDLaethe39.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethe39.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng" 
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng"
 schematypens="http://relaxng.org/ns/structure/1.0"?>
-<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng" 
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng"
 type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
 <TEI xmlns="http://www.tei-c.org/ns/1.0"
     xml:id="BDLaethe39"
@@ -13,7 +13,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 <title>Gadla Gabra Manfas Qǝddus, Taʾammǝra Gabra Manfas Qǝddus, Text on councils</title>
                 <editor key="DR"/>
                 <editor key="JG"/>
-                <editor key="MV"/><editor key="SH"/><funder>Akademie der Wissenschaften in Hamburg</funder>
+                <editor key="MV"/>
+                <editor key="SH"/>
+                <editor key="EDS"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
             </titleStmt>
             <publicationStmt>
                 <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
@@ -33,21 +36,22 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                             <idno>Juel-Jensen 17</idno>
                         </altIdentifier>
                     </msIdentifier>
-                    
+
                     <msContents>
                         <summary/>
-                        
+
                         <!--<msItem xml:id="ms_i1">
-                      
+
                         </msItem>-->
-                        
+
                     </msContents>
-                    
+
                     <physDesc>
                         <objectDesc form="Codex">
                             <supportDesc>
                                 <support>
                                     <material key="parchment"/>
+                                    <p>Holes on some leaves (e.g. <locus target="#11 #78 #83"/>), presumably from the parchment making process.</p>
                                 </support>
                                 <extent>
                                     <measure unit="leaf">90</measure>
@@ -64,19 +68,19 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                         <width>170</width>
                                     </dimensions>
                                     <note corresp="#leafdim">Data on leaves dimensions taken from <locus target="#12r"/>.</note>
-                                </extent> 
+                                </extent>
                                 <collation>
                                     <list>
                                         <item xml:id="q1">
                                             <dim unit="leaf">3</dim>
                                             <locus from="1" to="3"/>
                                             <note>The original quire structure is not visible.</note>
-                                            <note>Parchment support around the quire.</note>
+                                            <note>A parchment guard has been places around the quire.</note>
                                         </item>
                                         <item xml:id="q2">
                                             <dim unit="leaf">3</dim>
                                             <locus from="4r" to="7v"/>
-                                            <note>All leaves are detached and probably did not constitute an original quire, see <locus target="#7v"/>, 
+                                            <note>All leaves are detached and probably did not constitute an original quire, see <locus target="#7v"/>,
                                             where remains of the other half of the bifolium are visible.</note>
                                         </item>
                                         <item xml:id="q3">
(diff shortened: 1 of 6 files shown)


MassimoVilla (approved):
Thank you Eliana!
I just noticed some minor typos which I corrected by editing directly the files

OxfordBodleian/Juel-Jensen/BDLaethe39.xml
```diff
@@ -205,37 +213,49 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
-                        
-                     <!--   <bindingDesc>
-                            <binding xml:id="binding">
-                                <decoNote xml:id="b1"></decoNote>
-                                <decoNote xml:id="b2" type="bindingMaterial">
-                           <material key="wood"/>
-                        </decoNote>
+
+                        <bindingDesc>
+                            <binding contemporary="Ethiopian false">
```
thea-m, line 230: by "false" I understand here that the binding is not the original one. Is it here the case that the entire binding has been replaced, including the boards, or that elements (the sewing) have been repaired? I wonder whether it might make sense to make this explicit in a note? Maybe my question does not make conceptual sense.

OxfordBodleian/Juel-Jensen/BDLaethf24.xml
```diff
@@ -65,14 +69,15 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
-                                <foliation>Pencil foliation in the upper right corner carried out by the Bodleian Library during our stay 
+                                <foliation>Pencil foliation in the upper right corner carried out by the Bodleian Library
                                     in January 2020.</foliation>
                                  <collation>
                                     <list>
                                         <item xml:id="q1" n="A">
                                             <dim unit="leaf">2</dim>
                                             <locus from="1r" to="2v"/>
                                         </item>
+                                        <!-- EDS: 06.12.2022 A note on the fact that this is a reused bifolium should be placed here? -->
```
thea-m, line 80: yes, please do it, it cannot hurt. We will then later see anyway about the visualisation of the quire structure

OxfordBodleian/Juel-Jensen/BDLaethg36.xml
```diff
@@ -98,13 +102,17 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                             <locus from="51r" to="54v"/>
                                             1, stub after 4
                                             3 added
-                                            <note><locus target="#54"/>  and <locus target="#52"/> probably formed an original bifolium, which has now become separated.
+                                            <note><locus target="#54"/> and <locus target="#52"/> probably formed an original bifolium, which has now become separated.
                                             <locus target="#53"/> is sewn to <locus target="#54"/>.</note>
+                                            <!-- EDS 01.12.2022: Could the stubs after f. 54 be cut leaves?-->
```
thea-m, line 107: There seems to be text visible on the last one, the only one visible in the images.
I actually would not know how to encode a cut leaf vs a stub formally in the quire structure...but a note could be made at the least.
What do the others think?

thea-m (changes requested):
Thanks a lot! A few questions, not really requested changes

OxfordBodleian/Juel-Jensen/BDLaethg36.xml
gnizla: Definetly let's not invent a new encoding system, could include a note if you believe it's a worthwhile feature!

OxfordBodleian/Juel-Jensen/BDLaethe39.xml
ElianaDalSasso: I think that the boards may be contemporary with the content but the sewing has been repaired certainly after, when the strips of textile and parchment have been placed around the quires. You are right, it should be specified. Which is the best place for the note in your opinion? After @contemporary?

OxfordBodleian/Juel-Jensen/BDLaethg36.xml
ElianaDalSasso: It looks like there are letters on the stubs, doesn't it? I am wondering if a text has been cut out...

OxfordBodleian/Juel-Jensen/BDLaethf24.xml
```diff
@@ -33,21 +36,22 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+
                     </msContents>
-                    
+
                     <physDesc>
                         <objectDesc form="Codex">
                             <supportDesc>
                                 <support>
                                     <material key="parchment"/>
+                                    <p>A large hole, presumably resulting from the parchment making process, is found on <locus target="#2"/>.</p>
```
karljonaskarlsson, line 54: Maybe the word "tear" would be better than "hole", because now I first thought of an intentional hole. Or maybe now :)

OxfordBodleian/Juel-Jensen/BDLaethf24.xml
```diff
@@ -100,8 +105,14 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                         </item>
                                     </list>
                                 </collation>
+                                <condition key="deficient">
+                                  The manuscript is in bad condition. The upper board is broken in two halves and the outer is missing, together with the lower board.
```
karljonaskarlsson, line 109: Do "upper" and "lower" refer to the front and back board? This terminology is a little confusing to me – is this standard?

OxfordBodleian/Juel-Jensen/BDLaethf33.xml
```diff
@@ -33,21 +36,23 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+
                     </msContents>
-                    
+
                     <physDesc>
                         <objectDesc form="Codex">
                             <supportDesc>
                                 <support>
                                     <material key="parchment"/>
+                                    <p>Holes, presumably resulting from parchment making process, are found on <locus target="#1 #25"/>.
```
karljonaskarlsson, line 54: Again, maybe "tears"?

karljonaskarlsson (changes requested):

OxfordBodleian/Juel-Jensen/BDLaethf24.xml
ElianaDalSasso: Since it has a rounded shape, I would say it is an hole, but I am open. What do others think?

OxfordBodleian/Juel-Jensen/BDLaethf24.xml
ElianaDalSasso: Upper and front or lower and back are synonyms, I think I used upper and lower because they were used in the other project I worked. I need to go through all the descriptions and check it! (actually it would be better to talk about right and left board, to avoid any confusion)

OxfordBodleian/Juel-Jensen/BDLaethf33.xml
ElianaDalSasso: Same as before, they are rounded in shape.

OxfordBodleian/Juel-Jensen/BDLaethf24.xml
thea-m: My feeling is that front and back board are used most commonly in Ethiopian studies, and since they are not wrong, I would suggest to stick to them.

OxfordBodleian/Juel-Jensen/BDLaethf24.xml
thea-m: To me, "hole" is the usual term for this phenomenon (I also have the association with roundedness). They are differenciated from holes or tears that happened at a later stage in the life of the manuscript by their place in the description.

karljonaskarlsson: My point was that it would be good to distinguish these (accidental) "holes" from the (intentional) "holes" that, for example, resulted from the binding. It would be possible to use another term for the other piercings as well… or maybe just to specify using "accidental"/"intentional"... or maybe there is no need to clarify, and the average reader will understand what is meant (although it took me some seconds of thought to understand) :)

ElianaDalSasso: Doesn't the reference to the circumstances in which the holes originated (parchment making) help to clear their nature?

thea-m (approved):

OxfordBodleian/Juel-Jensen/BDLaethf24.xml
karljonaskarlsson: My feeling is that the process of parchment making could also leave intentional holes in the parchment, depending on which cuts are used in the manuscript :)  But I think you can use the formulation that you had – after all, Dorothea and Massimo found it clear, so it is good :)

PR #1948, opened 2022-12-16
Update BNFet5.xml

thea-m:
added reference suggested by Carsten
As Carsten pointed out, the information given about dating is not rendered well on the website (dates are not visible and bibliography is not broken down. Would it be better to just repeat the date in each element's contents?


eu-genia:
I have pushed a commit with a suggestion of encoding dates and relevant bibliographies; indeed, dates should be made more explicit (or the code extensively revised...)

thea-m:
Thank you, fine by me! (I can't approve because it's my own PR)

eu-genia (approved):

PR #1956, opened 2022-12-22
Update DW03.xml

thea-m:


eu-genia:
Thank you @thea-m ! In the other branch there was also I guess a suggestion for a new file DG001 (I am not sure about the source) and a change to NazMar001
https://github.com/BetaMasaheft/Manuscripts/pull/1952/files#diff-8eefbad5913decae215e67ff4cf94d5c4b0edf47b0a495fbc2a0e3927ad9f4d9, can you push these here too?

eu-genia:
እሶት and ለኣብ፡ in ll. 1056-57 are correct I guess and should not be changed as proposed previously?

thea-m:
yes, እሶት and ለኣብ are like this in the manuscript. 
The other changes were old ones on the branch from which the new branch for this PR was created (https://github.com/BetaMasaheft/Manuscripts/pull/1305 and for NazMar001, I see only a layout change)

eu-genia:
The branch with https://github.com/BetaMasaheft/Manuscripts/pull/1305/files is also extremely old maybe better move the file here and close that one too?

thea-m:
I'm not sure what needs to be done in the other files, I can look into it later. Let's keep this branch for DW, and create new ones for the other concerned records if needded. I think it will be easier to keep track then.

Getinet309 (approved):
ok I accept it, if it is written in the manuscript the error is from the writter.

PR #1959, opened 2022-12-23
Update DG001.xml

thea-m:


Getinet309 (approved):
It can be. but DGDG indicates dibqgasha debregolgota. if we agree we can make it DG

thea-m:
I see, that makes sense. The important thing is that the filename and the xml:id are identical.

Getinet309:
ok, that is good. Thank you

PR #1962, opened 2023-01-02
Update SinaiEt002.xml

thea-m:


DenisNosnitsin1970 (changes requested):
I think dating 1700-1899 is too broad, could you reduce it to one century?

eu-genia (approved):

thea-m:
We can look at all the datings systematically later, I'll merge for now

PR #1964, opened 2023-01-03
Update BMLacq755.xml

CarstenHoffmannMarburg:


FlorenceBML/BMLacq755/BMLacq755.xml
```diff
@@ -156,6 +156,14 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                               <foreign xml:lang="la">Prob. calami</foreign> and doodles</desc>
                                     
                                 </item>
+                                <item xml:id="e2"><desc>Handwritten pencil note "Acq. e doni 755" in
```
thea-m, line 159: I believe it has been our convention to end such statements as here with a period.

thea-m (approved):
Great, thank you! Only one minor remark

FlorenceBML/BMLacq755/BMLacq755.xml
CarstenHoffmannMarburg: thank you, we have added two more additiones

PR #1979, opened 2023-01-12
merging EMML 682 and EMIP 2746 = UNESCO 1-1

eu-genia:
(data from EMIP 2746 added to EMML 682, relation added, EMIP 2746 removed to avoid double descriptions of different surrogates of the same MS)


thea-m (approved):
Wow :)

PR #1991, opened 2023-01-18
renamed NALE to NLA in xml:id in order to resolve

eu-genia:
records beginning with NA are resolved as narratives

(there are two more records remaining in the NALE folder that may be quoted somewhere)


eu-genia:
@Ralph-Lee-UK NB I have updated NALE 28 with metadata from vHMML https://www.vhmml.org/readingRoom/view/533988 but it is quite different from the metadata for EMIP 2821 in the EMIP spreadsheet (dimensions, contents, page references...)

thea-m (approved):
Thank you!

thea-m:
Can be merged from my point of view - not sure if we wait for the NLA28 problem to be resolved

eu-genia:
I think if @Ralph-Lee-UK protests I will update. In general, as you will have also see from other EMML-EMIP PRs the metadata between EMIP and vHMML (Ted) is often very different (in some cases I was not even sure the same ms was described, but the images did fit!), but I usually go with vHMML as I suppose Ted has had a closer look

PR #2002, opened 2023-02-02
Dr 23 01 senkessar text

thea-m:


thea-m:
Check failed - EMML6952text has many "errors" due to the expansion. Maybe it is because of this.

thea-m:
Fixed one real error, apparently passed the check now :)

eu-genia (approved):

PR #2018, opened 2023-02-14
Create BLorient13156.xml

karljonaskarlsson:


new/BLorient13156.xml
```diff
@@ -0,0 +1,154 @@
...
+                            </recordHist>
+                        </adminInfo>
+                    </additional>
+                </msDesc>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <t:profileDesc xmlns:t="http://www.tei-c.org/ns/1.0">
```
thea-m, line 128: From here on, element names are preceded by t:. This is a technical error that happens occasionally, but needs to be corrected manually

thea-m, line 128: From here on, element names are preceded by t:. This is a technical error that happens occasionally, but needs to be corrected manually

new/BLorient13156.xml
```diff
@@ -0,0 +1,154 @@
...
+                                <locus target="#4v"/>
+                                <desc>The Virgin and Child</desc>
+                            </decoNote>
+                            <decoNote type="miniature" xml:id="d2">
+                                <locus target="#46v"/>
+                                <desc>Michael the Archangel</desc>
+                            </decoNote>
+                            <decoNote type="miniature" xml:id="d3">
+                                <locus target="#62v"/>
+                                <desc>[Our Lady] at prayer while standing</desc>
```
thea-m, line 72: Could you possibly use mark up instead of []? Are the contents of desc in these descriptions translations of captions (if so, it could be indicated) or your descriptions?

thea-m (changes requested):

new/BLorient13156.xml
karljonaskarlsson: They are adaptions of Strelcyn's translations of the captions. I didn't want to type in the complete captions at the moment

new/BLorient13156.xml
karljonaskarlsson: Ah, thank you for noticing! :)

thea-m (approved):

new/BLorient13156.xml
gnizla: Thanks @karljonaskarlsson. There is also always the option of assigning ATs to these when possible, but this is a choice of each encoder, so entirely up to you. Thanks for the work.

thea-m:
From my point of view, everything here is fine, I leave final merging to a project member

PR #2019, opened 2023-02-14
Update BMLacq679.xml

CarstenHoffmannMarburg:
Updated contents and decorations. Again I was not able to see the folia after f. 95r unless I did see them on thursday after Eugenia has restored them.

### DIFF
--- a/FlorenceBML/BMLacq679/BMLacq679.xml
+++ b/FlorenceBML/BMLacq679/BMLacq679.xml
@@ -38,93 +38,97 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                     </msIdentifier>
                     <msContents>
                         <summary/>
-                        <msItem xml:id="ms_i1">
+                    <msItem xml:id="ms_i1">
+                      <locus from="2r" to="132va"/>
+                        <title ref="LIT2701Dawit"/>         
+                        <msItem xml:id="ms_i1.1">
                             <locus from="2r" to="103v"/>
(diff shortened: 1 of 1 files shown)


FlorenceBML/BMLacq679/BMLacq679.xml
```diff
@@ -38,93 +38,97 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                     </msIdentifier>
```
eu-genia, line 38: you cannot see the images because the facs still has the wrong n (98) and the online version I had fixed was overwritten on merge by yours - I had not noticed you did not update the `facs`.
Now you will be able to see them again but only until this branch is merged, unless you correct the value
please replace n="144"
(`<idno facs="Laurenziana/BML_005/BML_005-" n="144">`)

eu-genia (changes requested):

FlorenceBML/BMLacq679/BMLacq679.xml
CarstenHoffmannMarburg: OK. Done.

CarstenHoffmannMarburg: The images no. 99-144 are still not visible to me. Yet alone small grey images/items are on the bottom line. I think I have to wait for a couple of minutes or until tomorrow.

eu-genia (approved):
thank you. and i can see all images, maybe refresh the viewer

PR #2036, opened 2023-03-02
Added ms items and miniatures

Ralph-Lee-UK:
This is a first attempt at describing the scroll contents - with some shortfalls, I am sure - but it would help to be able to see things more clearly!

### DIFF
--- a/EMIP/magicscrolls/EMIPms00469/EMIPms00469.xml
+++ b/EMIP/magicscrolls/EMIPms00469/EMIPms00469.xml
@@ -37,7 +37,29 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                <msContents>
                   <summary/>
                   <msItem xml:id="ms_i1">
+                     <locus target="#facs_1_line_1656341948235_22"/>
+                     <title ref="LIT4302Prayer"/>
+                     <textLang mainLang="gez"/>                     
+                  </msItem>
+                  <msItem xml:id="ms_i2">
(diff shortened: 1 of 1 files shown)


EMIP/magicscrolls/EMIPms00469/EMIPms00469.xml
```diff
@@ -1,100 +1,137 @@
...
+                  <idno facs="EMIP/Scrolls/MS0469/" n="10">Tweed Magic Scroll 1</idno>
+                  <altIdentifier>
+                     <idno>EMIP MS 0469</idno>
+                  </altIdentifier>
+               </msIdentifier>
+
+               <msContents>
+                  <summary/>
+                  <msItem xml:id="ms_i1">
+                     <locus target="#facs_1_line_1656341948235_22"/>
```
eu-genia, line 40: facs cannot be the target, `@facs` is a separate attribute, alongside `@target`, see https://betamasaheft.eu/Guidelines/?id=images Example 8

eu-genia (changes requested):

EMIP/magicscrolls/EMIPms00469/EMIPms00469.xml
Ralph-Lee-UK: @eu-genia I have added a target and facs statement, so I think this is now correct.

eu-genia: Almost ) just move `#` from `facs `to `target`

eu-genia (approved):

PR #2039, opened 2023-03-02
Dr 23 02 sinai et003

thea-m:


SinaiStCatherine/SinaiEt003.xml
```diff
@@ -186,12 +186,24 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
                               <width>125</width>
-                            <!-- <depth></depth>-->
-                           </dimensions>
+                           </dimensions>-->
+                          <dimensions type="outer" unit="mm">
+                            <height>178</height>
+                            <width>127</width>
+                             <depth>47</depth>
+                          </dimensions>
+                          <note>Due to exposure to humidity, the manuscript's depth has expanded to c. 85 mm along the other side.</note>
```
DenisNosnitsin1970, line 201: As far as I remember, the humidity image is only minor (you can see it on the images). This manuscript has a structural problem: the binding is made too tight and does not fit the textblock (the spine should have been wider). As the consequence, the pressure of the boards goes in wrong directions, and the textblock "opens" on the foreedge side, in a not very nice way. Then the leaves are not pressed to each other strong enough, they may start to cockle, and the damage process goes on.  Such volumes occur, but remarkably not many (also this may happen if someone tries to apply a used secondary binding to a new textblock). Smth like this.

DenisNosnitsin1970 (changes requested):

SinaiStCatherine/SinaiEt003.xml
thea-m: Ah, thank you! I have added this in condition, see if it is ok

DenisNosnitsin1970 (approved):

PR #2042, opened 2023-03-06
Dr 23 02 sinai et007

thea-m:
Description not complete (content not identified) because I don't have the images now, but let's to the PR so that all manuscripts are online quickly.
Width measurement is absent from metadata, will need to be estimated from the images :(

### DIFF
--- a/SinaiStCatherine/SinaiEt007.xml
+++ b/SinaiStCatherine/SinaiEt007.xml
@@ -1,0 +1,243 @@
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"
+?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:lang="en" xml:id="SinaiEt007" type="mss">
+  <teiHeader xml:base="https://betamasaheft.eu/">
+      <fileDesc>
+         <titleStmt>
+            <title></title>
+            <editor key="DR"/>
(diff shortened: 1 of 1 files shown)


SinaiStCatherine/SinaiEt007.xml
```diff
@@ -0,0 +1,300 @@
...
+                            <height>130</height>
+                            <width>84</width>
+                          </dimensions>
+                          <note corresp="#leafdim">Data on leaves dimensions taken from <locus target="#39r"/>.</note>
+                        </extent>
+                        <foliation>Foliation in the upper right corner of each folio, carried out in February 2023.</foliation>
+                       
+                        <collation>                      
+                          <signatures>On several quires, small undecorated quire marks in the upper inner corner of the first folio.
+                          It is possible that quire marks had originally been present in every quire, but are not visible everywhere anymore.</signatures>
```
DenisNosnitsin1970, line 91: Should be checked in the manuscript, if there the quire marks are usually visible. It is not uncommon that quire marks were indicated not in every quire.

SinaiStCatherine/SinaiEt007.xml
```diff
@@ -0,0 +1,300 @@
...
+                         <ab type="ruling">Ruling is partly visible.</ab>
+                         <ab type="pricking">Pricking is visible. Primary and secondary pricks are visible.</ab>
+                         <ab type="ruling" subtype="pattern">1A-1A/0-0/0-0/C</ab>
+                       </layout>
+                     </layoutDesc>
+                  </objectDesc>
+                 
+                  <handDesc>                
+                       <handNote script="Ethiopic" xml:id="h1" corresp="#ms_i1">
+                         <date></date>
```
DenisNosnitsin1970, line 192: 19th-century script

DenisNosnitsin1970 (changes requested):

SinaiStCatherine/SinaiEt007.xml
thea-m: As I wrote, I don't have the images now

DenisNosnitsin1970 (approved):

PR #2052, opened 2023-03-17
UM 001 013

DenisNosnitsin1970:
Images

### DIFF
--- a/ES/ESum001.xml
+++ b/ES/ESum001.xml
@@ -27,7 +27,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                <msIdentifier>
                   <repository ref="INS0017ʿUra"/>
                   <collection>Ethio-SPaRe</collection>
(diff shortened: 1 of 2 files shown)


ES/ESum001.xml
```diff
@@ -27,7 +27,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                <msIdentifier>
                   <repository ref="INS0017ʿUra"/>
                   <collection>Ethio-SPaRe</collection>
-                  <idno>UM-001</idno>
+                  <idno facs="ES/ESum001/" n="114">UM-001</idno>
```
eu-genia, line 30: please style facs as we always do that, check other E-S MSS with facs inserted for the model, thank you!

eu-genia (changes requested):

ES/ESum001.xml
eu-genia: `<idno facs="UM/001/UM-001" n="051">`

PR #2056, opened 2023-03-24
Adjusting facs commands to correct values

Ralph-Lee-UK:
I have adjusted the code to have the correct facs values for the transcription, and added facs statements for the images too.


Ralph-Lee-UK:
@eu-genia at least the code passed the validation! I was not clear how to add facs for both transcription and images - but this appears to work.

eu-genia (approved):

PR #2059, opened 2023-03-29
Bml741 updates

eu-genia:
(1) transkribus numbered text pages from 1 instead of from 5 (error in alignment, to be aware of and explain to aligners: empty pages must also be "recognized" with empty text); (2) transkribus export has difficulties with dealing with openings (v-r), it always numbers r-v (should be fixable in the future but for now all exports have this problem) (3) one image was included twice (image 55 and 56 both are ff. 51v-52r)  (4) in some cases the recognition of areas was not correct in transcribus resulting in wrong column assignment (f. 64r) - all these issues are now fixed manually for this ms to be considered for the future


thea-m (approved):
Thank you!

CarstenHoffmannMarburg (approved):

PR #2093, opened 2023-05-02
Update BDLaethg37.xml

gnizla:
Short deco note, plus dating update @thea-m you might want to consider revising handnote? Your choice of course. I have added cross-reference on line 688

### DIFF
--- a/OxfordBodleian/Juel-Jensen/BDLaethg37.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethg37.xml
@@ -486,6 +486,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         <handDesc>
                             <handNote xml:id="h1" script="Ethiopic">
                                 <date notBefore="1550" notAfter="1799"/>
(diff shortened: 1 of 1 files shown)


OxfordBodleian/Juel-Jensen/BDLaethg37.xml
```diff
@@ -519,6 +520,93 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                             </handNote>
                         </handDesc>
 
+                        <decoDesc>
+                          <summary>The prayers in the manuscript are prefaced by a portrait of
+                          their compiler. These are without captions, but their position in relation to each prayer,
```
thea-m, line 525: I took a few seconds to understand "compiler" in this context - at first, I thought the scribe of the manuscript was meant. I got it then :) but wonder if there is a better term? In German, "Betende" would work but "praying person" sounds a bit weird...
In any case, maybe it is obvious to everyone else, then let'ss keep it as it is!

thea-m (approved):
Wonderful, great that the decoration narrowed down the dating!

MassimoVilla (approved):

PR #2097, opened 2023-05-04
GDBS stubs

eu-genia:


Abennat/GDBS001.xml
MershaMengistie, line 1: In case it matters when the information is published, the sticky terms on line 14 ... anAncient ... would be separated.

thea-m (approved):

PR #2098, opened 2023-05-04
Max18

MassimoVilla:
Dear all,
here comes one MS from the Ullendorff collection. Only few photographs are available, so some fields are missing or incomplete. If we manage to organise our stay at the Bodl. Libr., I will be able to add further information, included the incipits and explicits of the inner sections of the book of Henok.

I have one doubt concerning a presumed name of an individual in the supplication formula on fol. 1rb-c በረከተ፡ አምላኩ፡ ተሀሉ፡ ምስለ፡ ፍቁሩ፡ ገብረ፡ እግዚአብሔር፡ አቡነ፡ ሄኖክ፡. It is not clear if ገብረ፡ እግዚአብሔር፡ is a proper name or an epithet introducing the name አቡነ፡ ሄኖክ፡. Shall I create another record for this Gabra ʾƎgziʾabǝḥer (in addition, of course, to ʾAbuna Henok)?

Also, there is one very minor correction regarding another manuscript.

### DIFF
--- a/OxfordBodleian/Ullendorff/BDLeu5.xml
+++ b/OxfordBodleian/Ullendorff/BDLeu5.xml
@@ -10,9 +10,13 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader xml:base="https://betamasaheft.eu/">
         <fileDesc>
             <titleStmt>
-                <title>Henok</title>
+                <title xml:lang="gez" xml:id="title1">ሄኖክ፡</title>
+                <title xml:lang="gez" corresp="#title1" type="normalized">Henok</title>
+                <title xml:lang="en" corresp="#title1">Book of Enoch</title>
                 <editor key="DR"/>
                 <editor key="JG"/>
+                <editor key="MV"/>
+                <editor key="EDS"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
             </titleStmt>
             <publicationStmt>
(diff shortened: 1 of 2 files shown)


gnizla:
Regarding ገብረ፡ እግዚአብሔር፡, leaving it to Thea and Jonas to comment on the plausibility of it being an epiteth or a Personal name, I would say it depends on whether you want it indexed or not in the volume.

gnizla (approved):

OxfordBodleian/Ullendorff/BDLeu5.xml
```diff
@@ -141,25 +277,24 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
-                                  <!--  <listBibl type="secondary">
-                                        <bibl></bibl>
-                                    </listBibl>-->
-                                </source>
-                            </recordHist>
-                        </adminInfo>
-                    </additional>
-                    
+                        <provenance>
+                            The provenance of the manuscript is unknown. It is unclear whether Gabra ʾƎgziʾabǝḥer (lit. 'the servant of God'), mentioned in the supplication
```
thea-m, line 285: I would maybe minimally change the wording: From reading only this description, without consulting the encoding or manuscript, I would not have understood the exact situation in the supplication. Maybe something like:
`The supplication...mentions Gabra ʾƎgziʾabǝḥer (lit. 'the servant of God'), followed by an erasure, over which the name of a certain <persName ref="PRS14014Henok" role="owner">ʾAbuna Henok</persName> has been added. It is unclear....`

thea-m (approved):
Thank you, everything is fine, just a minor suggestion regarding wording. 
On Gabra Egiʾabəḥer, I have no other input than you. If we assume that under the erased part, there was a personal name, it would have been an epithet and there would be no need to create a record. I think that this is likely, but another possibility for the erased part would be maybe the name of his wife '"wa-fəqərtu .... ....", in which case it would have been a name. Maybe we can wait if we get a chance to check the entire manuscript, there might be more attestations.

OxfordBodleian/Ullendorff/BDLeu5.xml
MassimoVilla: Good, I have now revised the wording following your suggestion. Good idea to wait for other possible attestations of this name when the entire MS is accessible to us.
Thank you

ElianaDalSasso (approved):
Dear Massimo, thank you! For me, it is all good!

PR #2106, opened 2023-05-10
EAP336-8 stubs

eu-genia:


MershaMengistie (approved):
Is it not worth to mention that the manuscripts collected in the project are commentary texts according to upper house andemta tradition. The title of the project was _Preserving the lay bet andemta: the Ethiopian intellectual legacy on the verge of extinction._

thea-m (approved):

PR #2109, opened 2023-05-10
BDLaethf35

thea-m:
Hi, 
here is one of the new additions. Not all leaves could be photographed, so the textual description might be completed at a later point.
Have fun!

### DIFF
--- a/OxfordBodleian/BodNewMSS/BDLaethf35.xml
+++ b/OxfordBodleian/BodNewMSS/BDLaethf35.xml
@@ -82,12 +82,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                 <hi rend="rubric">አኰቴተ፡ ቍርባን፡ ዘእግዝእ</hi>ትነ፡ <hi rend="rubric">ማርያም፡</hi> ወላዲተ፡ 
                                 አም<hi rend="rubric">ላክ፡ ጽ</hi>ርደት፡
                                 <hi rend="rubric">ዘነበበ፡ በመንፈ</hi>ስ፡ ቅዱስ፡ አባ፡ ህርያቆስ፡ ኤጲስ፡ ቆጶስ፡ ዘሀገረ፡ ብህንሳ፡ ጸሎቱ፡ ወበረከቱ፡ የሀሉ፡ 
-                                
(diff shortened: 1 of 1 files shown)


gnizla (approved):

OxfordBodleian/BodNewMSS/BDLaethf35.xml
```diff
@@ -34,9 +37,97 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                    <persName ref="PRS14002WaldaS">ወልደ፡ ሥላሴ፡</persName> ለዓለመ<supplied reason="omitted">፡</supplied></hi> 
+                                ዓለም፡ አሜን፡ 
+                                ጐሥዓ፡ ልብ<cb n="b"/><hi rend="rubric">የ፡ ቃለ፡ ሠናየ፡ ጐሥዓ፡ ልብየ፡</hi> ቃለ፡ ሠናየ፡ ወአነ፡ አየድዕ፡
+                                ቅዳ<hi rend="rubric">ሴሃ፡ ለማርያም<supplied reason="omitted">፡</supplied> አኮ፡
+                                    በአብዝኆ፡</hi> አላ፡ በአውህዶ<supplied reason="omitted">፡</supplied> ወአነ፡ አየድዕ<supplied reason="omitted">፡</supplied> ውዳሴሃ፡ ለድንግል።
+                            </incipit>
+                        </msItem>
+                        
+                        <msItem xml:id="ms_i3">
+                            <locus to="107ra"/>
```
MassimoVilla, line 95: `<locus to="107ra"/>`? Text 3.1 is said to begin on fol. 105rb.

MassimoVilla (changes requested):
Thank you Thea. Two very minimal comments

OxfordBodleian/BodNewMSS/BDLaethf35.xml
thea-m: The problem is that the beginning of the text was not photographed. 3.1. ends on 105rb, but we don't have its beginning, so I added only the value for the text's ending. I added a comment in the record to make it clearer.

thea-m:
Thank you, Massimo!

ElianaDalSasso (approved):
Thank you Thea! All good!

PR #2111, opened 2023-05-10
1st commit

MarcinKrawczuk:
I have replaced LIT2432Teachi into LIT6819Teachi318HA in (hopefully) all records.

Also, it would be nice to add LIT2432Teachi "generic record" to the title, but don't know whether it's possible...


thea-m (approved):
Thank you! Yes, it can be added in the work record, manually by writing "(generic title)" in the first English title

PR #2120, opened 2023-05-18
Ambrosiana old fund (pentaglots and the Santo Stefano mss)

eu-genia:
I used most of the data from the catalogues, some texts are not identified as not possible without the images

Should we also create records for Grébaut 3 (fondo Bricchi) and 5, which are not in Ambrosiana any more?


abausi:
I would say: yes, as long as we have a description; the MS could surface 
again at some point or be identified with other MSS presently from 
libraries and it would be useful to be able to identify them.

Il 18/05/2023 01:25, Eugenia ha scritto:

abausi:
It seems to be fully OK. Pentaglotton and Tetraglotton and MSS from 
Santo Stefano are the most historical items from the collection,

Il 18/05/2023 01:25, Eugenia ha scritto:

abausi (approved):

PR #2121, opened 2023-05-18
Update RNBdorn616.xml

eu-genia:
please check the ID (correction suggested) - is this the standard set of miracles?
or are these miracles from the DersanaMikael? But there we have no sub-id for Miracles as they are split by months

### DIFF
--- a/StPetersburg/RNBdorn616.xml
+++ b/StPetersburg/RNBdorn616.xml
@@ -180,7 +180,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
 
                     <msItem xml:id="ms_i1.12">
                         <locus from="61v"/>
-                        <title type="complete" ref="LIT1295Dersan#Miracles">ዜና፡ ተአምራቲሁ፡ ለመልአከ፡ ክቡር፡ ሚካኤል፡ </title>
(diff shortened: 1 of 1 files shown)


DenisNosnitsin1970:
The correction might be ok, but I cannot say with certainty to which extend it is the standard set. Neither catalogue nor the record LIT1295Dersan quote incipits (LIT1295Dersan as only some descriptions). The catalogue says only "12 Miracles of Archangel Michael". Well, possibly it is but I cannot check. However, I assume the adjusting cannot be limited to this correction only. One needs to harmonize LIT4155MiraclesMichael with LIT1295Dersan (where some Miracles are written out but no one is quoted, so all this is not very transparent), and continue in the numerous manuscripts with Dersana Mikael, right?  I think it will be a big work.

eu-genia:
The issue here is that the ID you used "LIT1295Dersan#Miracles" does not exist at all.

eu-genia:
we have LIT1295Dersan for the Dersan, where homilies are followed by salams and miracles (each with an individual sub-ID, no collective one) and we have LIT4155MiraclesMichael for a set of 12 miracles and we have LIT5066MiraclesMichael for any other set/quantity of miracles.
So the MS description has an error that must be fixed.

DenisNosnitsin1970:
Now I see. Theoretically, the miracles under LIT4155MiraclesMichael should be the same as those in LIT1295Dersan, then. OK, I can correct the record.

eu-genia:
I already corrected the record, just approve if OK

PR #2140, opened 2023-06-08
Create BLorient15421.xml

karljonaskarlsson:
Created entry for BL Or 15421. The current scarcity of information is intended.

### DIFF
--- a/new/BLorient15421.xml
+++ b/new/BLorient15421.xml
@@ -64,7 +64,6 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         <handDesc>
                             <handNote xml:id="h1" script="Ethiopic">
                                 <date notBefore="1450" notAfter="1550"/>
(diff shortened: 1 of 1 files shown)


gnizla (approved):
Thanks Jonas, I appreciate you are protecting the work of other colleagues for ethical reasons

SophiaD-M (approved):

SophiaD-M (approved):

SophiaD-M:
I am creating complications, sorry. 
I will stay out of this now.

thea-m:
Thank you!
I have corrected the file, which was invalid due to an issue in `<bindingDesc>`. In case you overlooked this, that's no problem, but in case your editor didn't show you any error messages, it would be necessary to investigate. 
I also delete the erroneous "t:" prefixes from the names of elements, which were added by the app for reasons we don't understand. In case this happens again and you notice them, you can also delete them. 
In general, I would recommend to create new manuscript records not through the app, but by copy pasting an existing record. This avoids such problems.

PR #2146, opened 2023-06-23
Moving the description from mda002 to C4IV123

eu-genia:
on seeing the images it came out that the commentary on the Hebrews corresponds to EAP336-2-4 not EAP336-2-2 as assumed

### DIFF
--- a/Abennat/C4IV123.xml
+++ b/Abennat/C4IV123.xml
@@ -1,0 +1,454 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" type="mss" xml:lang="en" xml:id="C4IV123">
+<teiHeader>
+        <fileDesc>
+            <titleStmt>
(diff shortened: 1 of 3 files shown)


MershaMengistie (approved):

MershaMengistie:
Yes!

eu-genia:
sorry please do not merge yet I think the shelfmark is still wrong

eu-genia:
Mamher Deggu subset as submitted to me has as 002 a different ms of Afawarq
image

but also as 004 a still different ms of Afawarq
image

The MS described by Addisie is 
image

I have not yet been able to map it to any of the MSS photographed for EAP

MershaMengistie:
Mss MD002 and MD004 are two different witnesses of Afawarq. As I see, Addisie did MD004. Both are in the folder under those numbers. I assume I didn't understand your question, Zhenia.

eu-genia:
Addisie did not do 004.
004 is c4-IV-168, her MS is c4-IV-123. It is another MS. I do not know yet where to save it. It does not belong to Mamher Deggu collection

eu-genia:
I have looked through all the images I received as EAP336 collection (including unsubmitted) and this MS is not among them

MershaMengistie:
I think this happened because the copies I gave to Addisie were randomly selected from my folders. And it was ONLY (I am still saying) for her to start exercising not to catalogue them. But...! This very ms might be taken from another folder that is not of EAP336. So I suggest saving it somewhere until we figure out where it is taken from.

eu-genia:
We cannot catalogue unidentified manuscripts, even as exercise, we need reliable metadata first. Please as soon as possible specify exactly which collection this manuscript belongs to. 
In the future we can never do anything with unattributed witnesses. This data cannot be made available on our portal.
Since the MS has a tourism agency shelfmark it is clearly a known manuscript and we cannot ascribe it some invented provenance.

eu-genia:
I have put it for now in the Abennat folder with the original institutional shelfmark as it would be a shame if Addisie's efforts are not even visible. It was a lot of work. We still must by all means identify the repository.

PR #2152, opened 2023-07-12
Update BMLacq811.xml

CarstenHoffmannMarburg:
There are no images available and no information on foliation in Marrassinis catalogue. Alas, we can say nothing on foliation.

### DIFF
--- a/FlorenceBML/BMLacq811.xml
+++ b/FlorenceBML/BMLacq811.xml
@@ -144,7 +144,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                         <width cert="medium">155</width>    
                                     </dimensions>
                                 </extent>
-                                <foliation>No foliation.</foliation>
+                                <foliation/>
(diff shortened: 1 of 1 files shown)


thea-m:
Maybe then it would make sense to write "The catalogue includes no information on the presence of foliation" to prevent readers from wondering whether it was forgotten in the encoding.

CarstenHoffmannMarburg:
I am not sure, if this is useful. I guess most catalogues will not specifically mention the place or kind of foliation. I think, it is better to keep in unrecorded.

thea-m:
It is my experience that if no information is given, readers (or we ourselves in a few years' time) will wonder whether the absence means absence of the feature or absence of information, and since you already researched the topic, it would be a pity to lose the knowledge.

CarstenHoffmannMarburg:
I was mistaken:
"Acq. e doni 811. Membr., sec. XIX-XX, mm. 160 x 155, fr. 153 non num." 
I guess "non num." is for "non numerati". So please better discard this branch.

thea-m:
Ah, good that you checked! Then I will close the PR and the change will be discarded.

PR #2154, opened 2023-07-12
Max21

MassimoVilla:
Dear all, these are my two last scrolls. I worked on them at the same time, so I opted for one single PR for both scrolls. I hope it is not too much.

### DIFF
--- a/OxfordBodleian/Ullendorff/BDLeu10.xml
+++ b/OxfordBodleian/Ullendorff/BDLeu10.xml
@@ -48,7 +48,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                             </incipit>
                             <explicit xml:lang="gez">
                                 እልመንክር፡ ጋኔን፡ ዘትሬኢ፡ በዓይነ፡ ጠባይዕ፡ እሳታዊ፡ ሹሐል፡ እልመክኑን፡ በዝ፡ አስማተ፡ ስምከ፡ ከመ፡ ኢትቅረቡ፡ ኀበ፡ ነፍሳ፡
-                                ወሥጋሃ፡ ለዓመትከ፡ ወለተ፡ <hi rend="rubric">ሰንበት፡ እንጐቻ።</hi>
+                                ወሥጋሃ፡ ለዓመትከ፡ <persName ref="PRS14046WalattaSa" role="owner">ወለተ፡ <hi rend="rubric">ሰንበት፡ እንጐቻ።</hi></persName>
                             </explicit>
                         </msItem>
                         <msItem xml:id="ms_i2">
@@ -62,7 +62,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                             </incipit>
                             <explicit xml:lang="gez">
                                 ጋኔን፡ ዘትሬኢ፡ በዓይነ፡ ጠባይዕ፡ ማያዊ፡ ተአሠር፡ ይቤ፡ በቃለ፡ እግዚአብሔር፡ ከመ፡ ኢትቅረቡ፡ ኀበ፡ ነፍሳ፡
-                                ወሥጋሃ፡ ለዓመትከ፡ ወለተ፡<hi rend="rubric">ሰንበት፡ እንጐቻ።</hi>
+                                ወሥጋሃ፡ ለዓመትከ፡ <persName ref="PRS14046WalattaSa" role="owner">ወለተ፡<hi rend="rubric">ሰንበት፡ እንጐቻ።</hi></persName>
                             </explicit>
                         </msItem>
                         <msItem xml:id="ms_i3">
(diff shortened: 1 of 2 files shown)


OxfordBodleian/Ullendorff/BDLeu10.xml
```diff
@@ -34,59 +37,162 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
-                      
-                        </msItem>-->
-                        
+                        <msItem xml:id="ms_i1">
+                            <locus from="1r1" to="1r67"/>
+                            <title type="complete" ref="LIT6811Protect"/>
+                            <textLang mainLang="gez"/>
+                            <incipit xml:lang="gez">
+                                <hi rend="rubric">በስመ<supplied reason="omitted">፡</supplied> አብ፡ ወወልድ፡ ወመንፈስ፡ ቅዱስ፡ ፩ አምላክ፡ ጸሎት፡ በእንተ፡ ማ</hi>ዕሠረ፡
+                                አጋንንት፡ ወአልቦ፡ ስም፡ ዘየሐዩ፡ በታሕተ፡ ሰማይ፡ ወምድር፡ ዘእንበለ፡ ስሙ፡ ለእግዚእነ፡ ኢየሱስ፡ ክርስቶስ፡ ጋኔን፡ ዘትሬኢ፡ በዓይነ፡ ጠባየእ፡
```
thea-m, line 46: ጠባየእ in the manuscript?

OxfordBodleian/Ullendorff/BDLeu10.xml
```diff
@@ -34,59 +37,162 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                            <title type="complete" ref="LIT6853Prayer"/>
+                            <textLang mainLang="gez"/>
+                            <incipit xml:lang="gez">
+                                <hi rend="rubric">በስመ<supplied reason="omitted">፡</supplied> አብ፡ ወወልድ፡ ወመንፈስ፡ ቅዱስ፡ ፩፡ አምላክ፡ ዘአሠሮ፡ ለጸላኢ፡</hi>
+                                በማዕሠረ፡ <del rend="strikethrough">ጸላኢ፡</del> አሌፍ፡ መደንግፅ፡ ማየ፡ ጠፈረ፡ ሰሎሞን፡ ሎፍሐም፡ ሎፍሐም፡ ሎፍሐም፡ መሐፍሎን፡
+                                መሐፍሎን፡ መሐፍሎን፡ ንምሎስ፡ ንምሎስ፡ ንምሎስ፡
+                            </incipit>
+                            <explicit xml:lang="gez">
+                                ጋኔን፡ ዘትሬኢ፡ በዓይነ፡ ጠባይዕ፡ ማያዊ፡ ተአሠር፡ ይቤ፡ በቃለ፡ እግዚአብሔር፡ ከመ፡ ኢትቅረቡ፡ ኀበ፡ ነፍሳ፡
+                                ወሥጋሃ፡ ለዓመትከ፡ ወለተ፡<hi rend="rubric">ሰንበት፡ እንጐቻ።</hi>
```
thea-m, line 65: could you mark up the personal name? Here and below

OxfordBodleian/Ullendorff/BDLeu10.xml
```diff
@@ -34,59 +37,162 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                    </dimensions>
+                                    <dimensions type="outer" unit="mm" xml:id="strip2">
+                                        <height>790</height>
+                                        <width>97</width>
+                                    </dimensions>
+                                    <dimensions type="outer" unit="mm" xml:id="strip3">
+                                        <height>420</height>
+                                        <width>97</width>
+                                    </dimensions>
+                                    <measure type="weight" unit="g">
```
thea-m, line 108: You're right; we don't have the weight measurements. I think in this case it's better to comment out the entire element

OxfordBodleian/Ullendorff/BDLeu7.xml
```diff
@@ -34,59 +37,210 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                            </incipit>
+                            <explicit xml:lang="gez">
+                                ወቡዳ፡ ተያያዥ፡ ወእጀ፡ ሰብአ፡ ጽላ፡ ወጊ፡ ወፍጌን፡ ወንለ<gap reason="illegible" unit="chars" quantity="1"/>ሽ፡ ዓይነት፡ ወዓይነ፡ ወዓይነ፡ ወርቅ፡ 
+                                መንሶ፡ ወሥራይ፡ ይትሐተሙ፡ በማኅተመ፡ እሌኒ፡ ንግሥት፡ ወበማኅተመ፡ ሰሎሞን፡ ንጉሥ፡ ከመ፡ ኢይቅረቡ፡ ኀበ፡ ነፍሳ፡ ወሥጋሃ፡ 
+                                ለአመተ፡ እግዚአብሔር፡ <hi rend="rubric">ወለተ፡ ኪዳን።</hi> 
+                            </explicit>
+                        </msItem>
+                        <msItem xml:id="ms_i5">
+                            <locus from="1r115" to="1r124"/>
+                            <title type="incomplete">Excerpt from <title ref="LIT5441ApostolicCreed">the Creed of the Apostles</title></title>
```
thea-m, line 97: I'm not sure, but I think that markup inside titles leads to errors. Maybe better to encode as `<title type="incomplete" ref="LIT5441ApostolicCreed">Excerpt from the Creed...`

thea-m (changes requested):
Thank you and congratulations! Just minor remarks

OxfordBodleian/Ullendorff/BDLeu10.xml
MassimoVilla: This means that the name will be indexed also in the Ethiopic text, right? Are we ok with that? Sorry for coming back to this question

thea-m: Yes, it will be. As the index will list all occurrences of each name regardless of page breaks, we will need to edit it manually anyway,

thea-m (approved):

OxfordBodleian/Ullendorff/BDLeu7.xml
```diff
@@ -34,59 +37,212 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                    <dimensions type="outer" unit="mm" xml:id="strip3">
+                                        <height>236</height>
+                                        <width>100-104</width>
+                                    </dimensions>
                                     <measure type="weight" unit="g">39</measure>
-                                </extent> 
+                                </extent>
+                              <condition key="good">The manuscript is in a mediocre state of preservation. 
+                              The parchment is crumpled due to rolling and is darkened toward the lower border; subsequently, the text is barely legible.
+                              Some letters are illegible due to material damage.
```
ElianaDalSasso, line 149: I would add 'Small areas of loss and tears are present, especially along the upper and lower margins'. Or something like that. :)

ElianaDalSasso (changes requested):

PR #2158, opened 2023-08-01
Abb 158 4 2

DenisNosnitsin1970:
Minor corrects

### DIFF
--- a/ParisBNF/abb/BNFabb112.xml
+++ b/ParisBNF/abb/BNFabb112.xml
@@ -280,14 +280,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                 </additions>
                   <bindingDesc>
                      <binding contemporary="Ethiopian" cert="medium">
-                        <decoNote xml:id="b1">Two wooden boards.</decoNote>
-                        <decoNote xml:id="b2" type="Cover">Blind-tooled leather cover.</decoNote>
-                        <decoNote xml:id="b3" type="Boards">Textile inlays.</decoNote>
-                        <decoNote xml:id="b4" type="SewingStations">4</decoNote>
-                           <decoNote xml:id="b5" type="Endbands" cert="low">Endbands.</decoNote>
-                           <decoNote xml:id="b6" type="Other">
-                  <term key="leafStringMark">Leaf string markers inserted in some folia.</term></decoNote>
-                              <decoNote xml:id="b7" type="bindingMaterial">
+                        <decoNote xml:id="b1" type="Boards">Two wooden boards.</decoNote>
+                        <decoNote xml:id="b2" type="Cover">Blind-tooled leather cover, <term key="textileInlay">textile inlays</term>.</decoNote>                        
+                        <decoNote xml:id="b3" type="SewingStations">4</decoNote>
+                           <decoNote xml:id="b4" type="Endbands" cert="low">Endbands.</decoNote>
+                           <decoNote xml:id="b5" type="Other"><term key="leafStringMark">Leaf string markers</term> inserted in some folia.</decoNote>
+                              <decoNote xml:id="b6" type="bindingMaterial">
                                     <material key="wood"/>
                                     <material key="leather"/>
                                     <material key="textile"/>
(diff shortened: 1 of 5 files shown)


StPetersburg/RNBefns2.xml
```diff
@@ -131,10 +127,11 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                </additions>
 
                   <bindingDesc>
-                     <binding contemporary="false" xml:id="binding">
-                        <decoNote xml:id="b1">Two wooden boards, leather quarter-cover.</decoNote>
-                         <decoNote xml:id="b2">Probably two pairs of sewing stations.</decoNote>
-                        <decoNote xml:id="b3" type="bindingMaterial">
+                     <binding contemporary="Ethiopian" xml:id="binding" cert="medium">
```
thea-m, line 130: I find the use of cert="medium" a bit problematic. For one, it will not be visualised (I believe), so the normal user will not be aware of any incertitude. But also when one looks at the XML, the reasons for the incertitude are not obvious (which is a pity, because probably you have though very well about including this encoding). Would it be possible, when cert="medium" is used to also include a short note describing the reasoning behind the encoding?

thea-m (changes requested):
Thank you, I just have one general remark about the use of the attribute cert. Otherwise everything is fine and can be merged.

StPetersburg/RNBefns2.xml
DenisNosnitsin1970: Ok, specifically here not necessary, I wil drop it.  You are right that it is not fully clear how the element should be used.

ParisBNF/abb/BNFabb112.xml
```diff
@@ -280,14 +280,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
-                        <decoNote xml:id="b4" type="SewingStations">4</decoNote>
-                           <decoNote xml:id="b5" type="Endbands" cert="low">Endbands.</decoNote>
-                           <decoNote xml:id="b6" type="Other">
-                  <term key="leafStringMark">Leaf string markers inserted in some folia.</term></decoNote>
-                              <decoNote xml:id="b7" type="bindingMaterial">
+                        <decoNote xml:id="b1" type="Boards">Two wooden boards.</decoNote>
+                        <decoNote xml:id="b2" type="Cover">Blind-tooled leather cover, <term key="textileInlay">textile inlays</term>.</decoNote>                        
+                        <decoNote xml:id="b3" type="SewingStations">4</decoNote>
+                           <decoNote xml:id="b4" type="Endbands" cert="low">Endbands.</decoNote>
+                           <decoNote xml:id="b5" type="Other"><term key="leafStringMark">Leaf string markers</term> inserted in some folia.</decoNote>
```
eu-genia, line 287: marks

eu-genia (changes requested):

ParisBNF/abb/BNFabb112.xml
DenisNosnitsin1970: done.

ParisBNF/abb/BNFabb173.xml
```diff
@@ -272,15 +272,13 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
-                           <decoNote xml:id="b5" type="Endbands">Endbands intact.</decoNote>
-                           <decoNote xml:id="b6" type="Other">
-                  <term key="leafStringMark">Leaf string markers inserted in some folia.</term></decoNote>
-                  <decoNote xml:id="b7" type="SlipCase">The manuscript has a well-preserved leather case made of two parts.</decoNote>
-                              <decoNote xml:id="b8" type="bindingMaterial">
+                       <decoNote xml:id="b1" type="Boards">Two wooden boards.</decoNote>
+                       <decoNote xml:id="b2" type="Cover">Blind-tooled leather cover, <term key="textileInlay">textile inlays</term>.</decoNote>                      
+                        <decoNote xml:id="b3" type="SewingStations">4</decoNote>
+                           <decoNote xml:id="b4" type="Endbands">Endbands intact.</decoNote>
+                           <decoNote xml:id="b5" type="Other"><term key="leafStringMark">Leaf string markers</term> inserted in some folia.</decoNote>
```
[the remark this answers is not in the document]
DenisNosnitsin1970: done. thanks!

PR #2160, opened 2023-08-05
Binding desc non jj 1/2

ElianaDalSasso:
Hi all!Here are some binding descriptions of non Juel Jensen manuscripts that since a while were pending in my Atom. But please don't work on them if it's your summer break! :)

### DIFF
--- a/OxfordBodleian/Ullendorff/BDLeu3.xml
+++ b/OxfordBodleian/Ullendorff/BDLeu3.xml
@@ -306,7 +306,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                  </decoNote>
                                  <decoNote xml:id="b2" type="Boards">
                                    The boards are flush with the bookblock and have shallow external bevels. The grain runs vertically.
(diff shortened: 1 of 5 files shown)


OxfordBodleian/Ullendorff/BDLeu3.xml
```diff
@@ -73,55 +73,56 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
                             </msItem>
                     </msContents>
-                    
+
                     <physDesc>
                         <objectDesc form="Codex">
                             <supportDesc>
                                 <support>
                                     <material key="parchment"/>
+                                    <p>Holes, presumably resulting from parchment making, are found on e.g. <locus target="#23 # #85"/>.</p>
```
MassimoVilla, line 125: One digit missing after the second #

MassimoVilla (changes requested):
Thank you Eliana!
I have just one very minor observation

thea-m (approved):
Great, thank you!

PR #2189, opened 2023-10-25
Sinai Fragments

DenisNosnitsin1970:
New records.

### DIFF
--- a/SinaiStCatherine/SinaiNewEt004.xml
+++ b/SinaiStCatherine/SinaiNewEt004.xml
@@ -1,0 +1,77 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="SinaiNewEt004" xml:lang="en" type="mss">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>Religious text</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="DN"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+               <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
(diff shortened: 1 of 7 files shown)


SinaiStCatherine/SinaiNewEt008.xml
```diff
@@ -0,0 +1,82 @@
...
+
+                    </msContents>
+                   <physDesc>
+                     <objectDesc form="Codex">
+                         <supportDesc>
+                           <support>
+                              <material key="paper"/>
+                               <material key="parchment"/>
+                           </support>
+                             <condition key="deficient">Includes several bigger fragments of a papaer manuscript, and a few small fragments
```
thea-m, line 45: paper

SinaiStCatherine/SinaiNewEt008.xml
```diff
@@ -0,0 +1,82 @@
...
+                    </msContents>
+                   <physDesc>
+                     <objectDesc form="Codex">
+                         <supportDesc>
+                           <support>
+                              <material key="paper"/>
+                               <material key="parchment"/>
+                           </support>
+                             <condition key="deficient">Includes several bigger fragments of a papaer manuscript, and a few small fragments 
+                                 of at least two manuscript (one on paper, one on parchment).</condition>
```
thea-m, line 46: manuscripts

SinaiStCatherine/SinaiNewEt004.xml
```diff
@@ -0,0 +1,77 @@
...
+                        <altIdentifier>
+                            <idno>Sinai New Finds Fragment 1</idno>
+                        </altIdentifier>
+                    </msIdentifier>
+                    <msContents>
+                       <summary/>
+
+                    </msContents>
+                   <physDesc>
+                     <objectDesc form="Codex">
```
thea-m, line 39: To me, it makes more sense to assign "Leaf", because even if it originally was a codex, what we now have is a piece of a leaf. I would even assign "leaf" (or possibly "other") to fragment 5, because its different parts do not constitute one common codex

thea-m (changes requested):
Thank you! only one question on the form

SinaiStCatherine/SinaiNewEt004.xml
DenisNosnitsin1970: I tend to disagree, sorry. I understand "leaf" not as remains of a codex, but as leaf/leaves which were not supposed to be codex, that is, were never bound. They also constitute a manuscript. I agree in the sense that we actually cannot know exactly what it was, but my and other's first assumption is, it was a part of a codex. On the other side, it is somewhat odd to call "codex" what is definitely does not look like  codex now, and assumption is not what we like. Let us ponder on that, and see how the issue is treated elsewhere.

PR #2192, opened 2023-10-30
adding link to images

eu-genia:


karljonaskarlsson:
On the photos, they seem to have used the shelfmark prefix "DS". The reason why Dorothea and I settled for "ADS" was to avoid a conflict with Dayr al-Suryan. I didn't have "DSAE" in my mind, but yes, seeing that Ewa used it, might make sense to change the Bm names of these files to "DSAE" instead of "ADS" (in order to avoid creating yet another abbreviation).

For the individual manuscript, however, I don't know if "DSAE-XVI" (etc for the other fragments) has ever been used. I used "DS-XVI", as did Denis in "Ancient chants for Abba Yohanni" (2018).

DenisNosnitsin1970:
Yes, I know the difference. DSAE stands for Dabra Sahel Arba'ettu Ensasa, which is the same library as the one of "DS", but the shelf mark DSAE (I assume from one to 10) was assigned to sets of images submitted to HMML. Most of them, or all, can be viewed in HMML. The shelfmark has been already used in some publications. DSAE 1-10 are mostly medieval or recent codices, DSAE-001 Four Gospels is the oldest. As to DS, it is reserved to the items of "geniza" only, with Roman numbers (these images were not submitted to HMML, as far as I know). I think it makes sense to keep the signatures as they are (DS also with Roman numbers) and do not change them. That is true, DS comes close to Dayr as-Suryan, I could have remembered this and modified "DS" a litte, fortunately it is actually not "DS" but "DS Ethiop. ...", and one should always add the institution name.

eu-genia:
We keep the shelfmarks, it is the filename (xmlID) which must be clearly disambiguated.
Yet there are DS with Roman numbers (for Genizah fragments), DS with normal numbers (mss on the project page) and DSAE mss, so three separate numbering sets.
I will check HMML...

PR #2131, opened 2023-05-26
Update BMLacq680

CarstenHoffmannMarburg:
The images of BMLacq680 are not complete. The scan images for folio 6v-7r (after scan page 8) as well as for folio 179v-180r (after scan 180). It can be discerned from the folio marks (bottom left) and from the text, that is not fitting.

In encoding the quire structure and the description of hands, I differed a lot from P. Marrassini. As well the last textpart (ms_i3 and ms_i4) are encoded independently contrary to Marrassinis account (See also Issue  #2367). Also I treated the index on f. 4v, I as an addition because it is by another hand as most of ms_i1 but not as the incipit of ms_i1, as Marrassini did.

PR #2195, opened 2023-11-03
Update BMLacq680

CarstenHoffmannMarburg:
I updated BMLacq680 according to P. Marrassinis' catalogue. 

Once again, I had to give a quire structure that differed from Marrassini's description. 

I also found foliation marks that Marrassini did not mention. I have kept the foliation as written in the manuscript for the record, unless it contradicts Marrassini's description, who described the first leaves as protective leaves with I-IV. However, in describing the hands and contents, Marrassini is not consistent in his folio numbers, sometimes counting the protection leaves and sometimes not. I think Marrassini worked on the manuscript before the foliation marks were inserted and got confused several times.
 
In the description of the hands, I also found many different hands that Marrassini did not count and in another round, I found various folios written by the same hand that Marrassini considered to be two or more. 

I decided to give the index to the theological treatises in ms_i1 as a div. I do not know if it is necessary to delete strings in the record that refer to the transcription of the transkribus. 

I had a quick look at the transkribus files. I noticed that the that the folio numbering in the Transkribus record follows Marrassini's system and does not count the protective leaves. The text of ms_i1.2 is given as folio 1r and not 5r as in my record. Is this a problem and how can it be resolved?

In the process of encoding the manuscript, I found a number of salāms that are not recorded as LIT IDs, as I explained in issue [#2438](https://github.com/BetaMasaheft/Documentation/issues/2438). However, since they are part of the commemorations in the Synaxary and do not have an independent circulation, we decided not to encode them with individual LIT IDs.

There are two problems with the viewer. First: Two openings (f. 6v-7r and f. 179v-180r) are missing from the viewer. Most likely they were omitted in the imaging and cannot be restored. Second: for image no. 128 (f. 126v-127r) there is no small image in the bar at the bottom. This is a minor problem, and I don't know if it's easy to fix. I discussed these problems recently in [#2436](https://github.com/BetaMasaheft/Documentation/issues/2436).

### DIFF
--- a/FlorenceBML/BMLacq680/BMLacq680.xml
+++ b/FlorenceBML/BMLacq680/BMLacq680.xml
@@ -35,7 +35,616 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                             <idno>Marrassini ms. 15</idno>
                         </altIdentifier>
                     </msIdentifier>
+                    <msContents>
+                        <summary/>
+                        <msItem xml:id="ms_i1">
+                            <title>Series of theological texts in Amharic</title>
+                            <msItem xml:id="ms_i1.1">
+                                <locus target="#4v">Index</locus>
+                                <title>Index</title>
+                            </msItem>
+                            <msItem xml:id="ms_i1.2">
+                                <locus from="5ra" to="6v"/>
+                                <title>Explanation of the creation</title>
+                                <incipit xml:lang="am"><hi rend="rubric">በመጀመሯ፡ እግዚአብሔር፡ ሰማይና፡ ምድርን፡ ብር</hi>ሃንና፡ መላእክትን፡ ፈጠረ።</incipit>
+                            </msItem>
+                            <msItem xml:id="ms_i1.3">
+                                <locus from="6v" to="9ra"/>
+                                <title>The secret of the flesh</title>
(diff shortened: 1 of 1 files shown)


CarstenHoffmannMarburg:
BNFet126 does not provide images. BNFet677 is not available but provides an error notification to me (HTTP ERROR 500 javax.servlet.ServletException). I don't know, which manuscripts I can consult easily and on which criteria I can build up my judgement.

eu-genia:
If a ms does not load in the app you can always consult the `XML `file, either in your Oxygen or by adding `.xml `in the browser (e.g. `betamasaheft.eu/BNFet677.xml`). 

If you see a bug such as HTTP Error and do not understand how to fix the error usually present in the XML file please in the future open a separate issue for me to resolve it. In this case I have cleaned the xml (but the file is possibly too large to load well).

BnFet126 - maybe @thea-m has images. If no recension can be easily identified we use the general ID

eu-genia:
Please also check https://github.com/BetaMasaheft/Manuscripts/pull/2131 - there seem to be two branches and two PRs with the same MS

CarstenHoffmannMarburg:
Yes, it is true. I made a second record for the same manuscript. I will compare them and will eventually make an update to the first one.

PR #2202, opened 2023-11-28
Update BMLacq679

CarstenHoffmannMarburg:
I updated BMLacq679 by adding four newly created PRS ID in a1. I created four PRS IDs for people mentioned in BMLacq679 as discussed in issue [#2310](https://github.com/BetaMasaheft/Documentation/issues/2310) and pull request #768.

Screenshot 2023_Maqaryos

NB. I did not understand how to reference an issue or a pull request. When I press the 'reference' button in the taskbar, nothing changes except for the # that is introduced.

### DIFF
--- a/FlorenceBML/BMLacq679/BMLacq679.xml
+++ b/FlorenceBML/BMLacq679/BMLacq679.xml
@@ -293,26 +293,31 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                                 </item> 
                                 <item xml:id="a1">
                                     <locus from="132vb" to="133rb"/>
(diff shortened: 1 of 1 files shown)


eu-genia:
to reference an issue you should simply go to the issue, copy the address line, and insert it like this `https://github.com/BetaMasaheft/Documentation/issues/2310`, it will be transformed to an active link https://github.com/BetaMasaheft/Documentation/issues/2310

simply typing # brings you only inside of the repository you are in. so you can type # when you are in an issue to link to another issue and you would get the dropdown menu to choose from, but to link from anywhere else you need simply to paste the address. thank you

eu-genia (approved):

thea-m (approved):

PR #2203, opened 2023-11-29
Update MsEth1.xml

CarstenHoffmannMarburg:
I updated the leporello from Lucerne as discussed in issue [#2401.](https://github.com/BetaMasaheft/Documentation/issues/2401)

### DIFF
--- a/Lucerne/MsEth1.xml
+++ b/Lucerne/MsEth1.xml
@@ -35,8 +35,10 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
                <msContents>
                   <summary>Leporello with prayers</summary>
                   <msItem xml:id="ms_i1">
-                     <title ref="LIT5888AsmatPrayer">Protective prayer</title>
-                     <incipit xml:lang="gez">ቀንዋት፡ ትትአካዕ፡ ማዕከለ፡ ፈያት፡ ፪ቱ፡ አድኅነኒ፡ ክርስቶስ፡ እምፀብዐ፡ ዲያብሎስ፡ ወሠራዊቱ፡ ወእምክረ፡ ልሳን፡ ቀያያፋ፡</incipit>
+                     <title ref="LIT3034RepCh314"/>
+                     <incipit xml:lang="gez"><hi rend="rubric">በስመ፡ አብ፡</hi><gap reason="ellipsis" unit="lines" quantity="2"/> ጸሎት፡ በእንተ፡ 
+                        ልሳነ፡ ሰብእ፡ ዘመድ፡ ወባዕድ፡ ምቀኛ፡ ወተንኮሊኛ፡ <hi rend="rubric">ናሁ፡ ተማኅ<supplied reason="lost">ፀ</supplied>ንኩ፡ በኆኅያተ፡ 
+                        ስምከ፡</hi> ካፍ፡ ወበቀዳማይ፡ የውጣ፡ ዘጥንተ፡ ፊደሉ፡ አሌፍ፡</incipit>
                      <explicit xml:lang="gez"><hi rend="rubric">ኦእ<hi rend="ligature">ግዚ</hi>ኦ፡ ዕቀቦ፡ ወአድኅኖ፡ እምልሳነ፡ ዘ</hi>መድ፡ ወባዕድ፡ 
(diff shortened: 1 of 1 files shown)


Lucerne/MsEth1.xml
```diff
@@ -48,33 +49,47 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
...
                            <dimensions type="outer" unit="mm">
                               <height>52</height>
                               <width>42</width>
-                              
                            </dimensions>
                         </extent>
-                        <foliation></foliation>
-                        <condition key="good"></condition>                        
+                        <foliation>No foliation</foliation>
+                        <collation>Three slender pieces of parchment have been tied together and folded into 27 pages, the first and the last  of which
```
eu-genia, line 61: 27 or 26? I counted 26

eu-genia (changes requested):

Lucerne/MsEth1.xml
```diff
@@ -48,33 +49,47 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
...
+                        <decoNote xml:id="b1" type="Boards">Two wooden boards, tied to the parchment with a leather cord.</decoNote>    
+                        <decoNote xml:id="b2" type="bindingMaterial"><material key="wood"></material></decoNote>
+                     </binding>
+                  </bindingDesc>
                </physDesc>
                <history>
                   <origin>
-                     
-                     <origDate notBefore="1800" notAfter="1900">nineteenth  century (?) </origDate>
+                     <origDate notBefore="1650" notAfter="1950">Second half of 17th to first half of the 20th century</origDate>
```
eu-genia, line 87: The manuscript is unlikely older than the mid 19th, the dating thus is mid-19th to mid-20th century, in my opinion (preference to the more recent half), unless there is any other external evidence (like known acquisition time). the shapes of ደ and ይ show this, also in part  ግ. The scribe may try to imitate, a bit crudely, another "older" hand. The binding, on two boards, is interesting. We should try to avoid datings like 1650-1950 as they are too wide and not really helpful.

_Originally posted by @DenisNosnitsin1970 in https://github.com/BetaMasaheft/Persons/issues/769#issuecomment-1831819144_

Lucerne/MsEth1.xml
CarstenHoffmannMarburg: The last page is glued to the back board. I do not know, how to count and how to make this point clear.

Lucerne/MsEth1.xml
CarstenHoffmannMarburg: ok.

eu-genia (approved):

Lucerne/MsEth1.xml
```diff
@@ -33,12 +33,13 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
...
                </msIdentifier>
                <msContents>
-                  <summary>Leporello with prayers፡</summary>
+                  <summary>Leporello with prayers</summary>
                   <msItem xml:id="ms_i1">
-                     <title>Protective prayer</title>
-                  </msItem>
-                  <msItem xml:id="ms_i2">
-                     <title>Protective prayer</title>
+                     <title ref="LIT5888AsmatPrayer">Protective prayer</title>
```
thea-m, line 38: It is better to identify and if necessary create new work IDs for protective prayers. Experience has shown that they can be often identified. The ID LIT5888AsmatPrayer was (if my memory is correct...) created for quick cataloguing in EMIP. It is not wrong, but if there is time, more specification would be beneficial.

Lucerne/MsEth1.xml
```diff
@@ -48,33 +49,48 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
...
+                     <layoutDesc>
+                        <layout columns="1" writtenLines="10"/>                            
+                     </layoutDesc>
+                  </objectDesc>
+                  <handDesc>                  
+                     <handNote xml:id="h1" script="Ethiopic">
+                        <seg type="script">Script of the 19th or 20th century</seg>
+                        <seg type="ink">Black, red</seg>
+                        <seg type="rubrication">Incipits, name of the owner</seg>
+                        <note>Long and slender script.</note>
```
thea-m, line 75: It is better to transfer the content of these notes into a single `<desc>`

thea-m (changes requested):
Great, thank you!

Lucerne/MsEth1.xml
CarstenHoffmannMarburg: Done.

eu-genia (approved):

thea-m (approved):

PR #2215, opened 2023-12-02
Emml8509

MarcinKrawczuk:


EMML/8001-9000/EMML8509.xml
```diff
@@ -0,0 +1,553 @@
...
+<ab type="cruces">A chi-rho sign on <locus target="102r">f. 102r</locus>.</ab>
+
+    </layout>
+  </layoutDesc>
+
+</objectDesc>
+
+<handDesc>
+<handNote script="Ethiopic" xml:id="h1">
+<desc>The manuscript was not used by Uhlig in his "Äthiopische Paläographie". It shows many palaeographic peculiarities e.g.
```
CarstenHoffmannMarburg, line 485: If I remember it right, all the mentioned features are not described at all in Uhligs palaeography. If this is the case, I would suggest to write rather:
`The manuscript shows many palaeographic peculiarities not mentioned in <bibl><ptr target="bm:Uhlig1988Palaographie"/></bibl>, e.g.`
However, I would not start with the information, that this manuscript was not used by Uhlig, because he took a small selection for his monograph and I would suggest to encode the work with a bibl.

CarstenHoffmannMarburg (changes requested):

CarstenHoffmannMarburg (approved):

PR #2224, opened 2023-12-13
Update BMLacq821

CarstenHoffmannMarburg:
I updated the record for BMLacq821. To my surprise, the transcription by transcribus was inserted in the ms-file. It may be necessary to split this record into two to have the transcribus record seperately.

### DIFF
--- a/FlorenceBML/BMLacq821/BMLacq821.xml
+++ b/FlorenceBML/BMLacq821/BMLacq821.xml
@@ -41,7 +41,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                 		<summary>Protective Prayers</summary>
                 	<msItem xml:id="ms_i1">
                 		<locus from="1r1" to="1r33"/>
-                		<title>Prayer against Bāryā</title>
+                		<title ref="LIT5888AsmatPrayer">Prayer against Bāryā</title>
                 		<note>Prayer against the evil eye of <term key="demon">Bāryā</term>, to bind the demons and to swallow
                 			them; with invocation of magical names.</note>
                 		<incipit xml:lang="gez"><hi rend="rubric">በስመአብ፡ <gap reason="ellipsis" unit="lines" quantity="2"/> ጸሎት፡ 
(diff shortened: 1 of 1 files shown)


CarstenHoffmannMarburg:
I wanted to encode Josef in ms_i6, but I did not find an ID for him. It is supposed to be Joseph, the son of Jacob and Rachel, who is registered as Q145746 in Wikiquote. Is his record missing in Bm?

I checked EMIP00227 and LIT6751PPAynat and discovered some similarities to ms_i6 in the manuscript and in the record respectively. However, I saw, that the link [www.betamasaheft.eu/LIT6751PPAynat](https://betamasaheft.eu/works/LIT6751PPAynat/main) is not available at the moment.

thea-m:
It's PRS5763Joseph. I've added the wikidata ID and the Bm ID to wikidata.

FlorenceBML/BMLacq821/BMLacq821.xml
```diff
@@ -33,8 +33,207 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+                				<condition key="good"/>
+                			</supportDesc>
+                			<layoutDesc><layout columns="1" writtenLines="182">
+                				<ab type="pricking">No pricking/></ab>
+                				<ab type="ruling">No ruling</ab>
+                			</layout>
+                			</layoutDesc>
+                		</objectDesc>
+                		<handDesc><handNote xml:id="h1" script="Ethiopic">
+                			<ab type="script"/>
```
eu-genia, line 187: why this `ab`?

eu-genia (changes requested):

eu-genia:
No, it is not necessary to split the record at all.
We split only when the transcriptions are too long for github.
If you see an error on calling an entry (as in LIT6751PPAynat) it is not because the "link is unavailable" but because there is an encoding issue or a problem with the code, and an issue is needed to settle the bug (done).

FlorenceBML/BMLacq821/BMLacq821.xml
CarstenHoffmannMarburg: Deleted `<ab type="script"/>`

eu-genia (approved):

PR #2228, opened 2023-12-20
Create FCCOSEth002.xml

thea-m:


CairoFCCOS/FCCOSEth002.xml
```diff
@@ -0,0 +1,218 @@
...
+                    
+                    <physDesc>
+                        <objectDesc form="Codex">
+                            <supportDesc>
+                                <support>
+                                    <material key="parchment"/>
+                                </support>
+                                <extent>
+                                    <measure unit="page" quantity="230">230</measure>
+                                    <dimensions type="outer" unit="cm">
```
eu-genia, line 98: we once said it is desirable to use mm

CairoFCCOS/FCCOSEth002.xml
```diff
@@ -0,0 +1,218 @@
...
+                                        <height>20.5</height>
+                                        <width>18</width>
+                                    </dimensions>
+                                </extent>
+                                <condition key="deficient">The binding is broken.</condition>
```
eu-genia, line 103: also repeated under binding below? needed twice?

eu-genia (approved):

CairoFCCOS/FCCOSEth002.xml
thea-m: I agree it is redundant. The reason I encoded it in this way is that this one sentence is the only information we have on the actual presence of a binding, and it equally applies to the condition. What would you propose?

CairoFCCOS/FCCOSEth002.xml
thea-m: The reason I didn't convert them is that I suspect that whole numbers given in centimeters might be rounded up, and converting into mm would simulate more preciseness than might be present here. On the other hand, if the filtering and searching benefits for quantitative analyses would outweigh that, let's do it.

CairoFCCOS/FCCOSEth002.xml
eu-genia: Am not sure, lets leave it

PR #2237, opened 2024-01-16
Create DMGYi135IIIC11b.xml

thea-m:


DenisNosnitsin1970 (approved):

CarstenHoffmannMarburg (approved):

HalleDMG/DMGYi135IIIC11b.xml
```diff
@@ -0,0 +1,170 @@
...
+            </langUsage>
+        </profileDesc>
+        <revisionDesc>
+            <change who="DR" when="2024-01-16">Created record after consultation of the manuscript on 21.12.2023</change>
+        </revisionDesc>
+    </teiHeader>
+    <text xml:base="https://betamasaheft.eu/">
+        <body>
+            <div type="edition">
+                <div type="textpart" xml:lang="ar">
```
eu-genia, line 154: am?

thea-m: other than the first word, the text is in Arabic

eu-genia: I have not found anything 100% fitting in TEI but I guess what they suggest is to use a separate language tag with a sub-type for the script (such as `az-Arab` for Azeri in Arabic script as opposed to `az-Lat` for Azeri in Latin script https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-language.html), which would mean adding something like `ar-Eth` for Arabic in fidel, and then in `langUsage `refer to this specific tag 

Otherwise I wonder if we can use `style `as described https://www.tei-c.org/release/doc/tei-p5-doc/en/html/WD.html (here it is mostly used to describe RL LR etc) and have something like
`xml:lang="ar" style="script: Ethiopic"` (it would not be displayed at the moment but at least it would be somewhere)

eu-genia (approved):

PR #2242, opened 2024-01-24
updated ID

thea-m:
https://github.com/BetaMasaheft/Documentation/issues/1538
https://github.com/BetaMasaheft/Documentation/issues/2455

Where the images were available, I checked that indeed the entire malkǝʾ is contained in the manuscript, in the other cases, it seems very likely as can be assumed by folio range indications.


eu-genia (approved):
thank you!

CarstenHoffmannMarburg (approved):

PR #2248, opened 2024-01-29
added graphic url

eu-genia:
dear @GueshSol, since you know how to work with xml, you can always add link to images as I did here for you

### DIFF
--- a/ChicagoIA/CIA20024.xml
+++ b/ChicagoIA/CIA20024.xml
@@ -1731,6 +1731,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <change who="GS" when="2022-10-28">Added decoration notes </change>
         </revisionDesc>
     </teiHeader>
+    <facsimile>
+        <graphic url="https://www.artic.edu/artworks/158772/bound-manuscript-the-miracles-of-mary-te-amire-maryam"></graphic>
+    </facsimile>
     <text xml:base="https://betamasaheft.eu/">
         <body>
             <ab/>


GueshSol (approved):

GueshSol:
Dear @eu-genia,
 Thank you so much! I just took a lesson.

eu-genia:
https://betamasaheft.eu/Guidelines/?id=images
see under
"Link to images not accessible via IIIF" Example 9

For partial digitizations there is also example 10 in case you ever need it

GueshSol:
Dear @eu-genia,
This is very helpful. 
Many thanks!

GueshSol:
Dear @eu-genia,
In relation to this I have also full image access (on request) of Ms. CBD914 (ʾAkkonu Bǝʾsi type) I think you have already created a stub for it. Some months ago I also created another without noticing the one you did can I delete mine and continue with your stub for a detailed description?

eu-genia:
Dear Guesh, we already have images for the Dublin mss (they were digitized by EMIP and we have all images on our server), but I realize that for some reason the manifest is not working and they are not visible. I will try to switch them on.
It is best if we all work from the same repository, you should be careful to not create doublets.
Feel free to add information to the stub.

GueshSol:
Dear @eu-genia,
You are correct, but I recall that there was no access to the image for nearly a year. Fortunately, thanks to W. Belcher's assistance in communicating with them, the manuscript was finally uploaded and made accessible again. Here is the link for the manuscript.
 https://viewer.cbl.ie/viewer/image/W_914/25/LOG_0000/.

eu-genia:
Now I see, the `facs `was in the wrong place in our record. The images have been with us since several years, from Steve Delamarter, they are also accessible on the shared server and visible https://betamasaheft.eu/manuscripts/CBD914/viewer. But since they are in black and white we can add the second URL or also replace the images in the viewer, since Dublin serves `iiif`. @thea-m @DenisNosnitsin1970 should I switch the `facs `to point directly do Dublin in the viewer, or leave both?

DenisNosnitsin1970:
I think we can leave both.

PR #2251, opened 2024-01-29
facs from emip to cbd

eu-genia:
since for the moment the app does not support multiple manifests we must have one in the viewer
we could leave emip (black and white) as the main one and have a link to CBD or just have the coloured CBD in our viewer where the manifest is available (as proposed here)
https://github.com/BetaMasaheft/Documentation/issues/2100
https://github.com/BetaMasaheft/Manuscripts/pull/2248

### DIFF
--- a/Dublin/CBD911.xml
+++ b/Dublin/CBD911.xml
@@ -24,11 +24,11 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                   <msIdentifier>
                      <repository ref="INS0422Chester_Beatty_Library"/>
                      <collection>Ethiopian</collection>
-                     <idno>CBD W 911</idno>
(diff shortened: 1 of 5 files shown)


eu-genia:
example with both
https://betamasaheft.eu/manuscripts/CBD912/viewer has just EMIP, link to images brings to CBD
https://betamasaheft.eu/manuscripts/CBD914/viewer has just EMIP, link to images brings to CBD

example with just one
https://betamasaheft.eu/manuscripts/CBD936/viewer just CBD
https://betamasaheft.eu/manuscripts/CBD955/viewer

DenisNosnitsin1970 (approved):

eu-genia:
since you approve you agree that we only have a facs pointing to the coloured images at CBD and suppress the BW ones, correct?

thea-m (approved):
Makes sense!

PR #2253, opened 2024-01-30
Emml7602 corr mk

MarcinKrawczuk:
I have added the work records, other than that I have left everything as it was except for bibliography


eu-genia (approved):
thank you

PR #2255, opened 2024-01-31
NALA 64 = UNESCO 2, 64 = EMML 1513 = EMML 7441 = EMIP 2857

eu-genia:


thea-m (approved):

eu-genia:
NB I kept EMML 1513 but inserted a cross-reference; it is a microfilm of a photocopy, and EMML 7441 is the microfilm of the ms itself, so a different object, but the source ms is the same.

PR #2266, opened 2024-02-02
BDLaethf28

thea-m:
Hey everyone, this is the last manuscript from me. See also https://github.com/BetaMasaheft/Works/pull/690
The description is very minimal. A more detailed description of the content would be desirable, but would take up considerable resources. If you think that this is not up to our catalogue's standards, we will reconsider and see what can be done.

### DIFF
--- a/OxfordBodleian/Juel-Jensen/BDLaethf28.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethf28.xml
@@ -115,7 +115,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                             </item>
                                         </list>
                                 </collation>
-                                <condition key="good">The original binding is missing and has been replaced by modern, possibly European wooden boards.
+                                <condition key="good">The original binding is missing and has been replaced by modern European wooden boards.
                                 The first and last page are stained by dirt and in places hardly legible. The corners are stained.</condition>
                             </supportDesc>
                             
@@ -147,7 +147,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         
                         <handDesc>
                             <handNote xml:id="h1" script="Ethiopic">
-                                <desc></desc>
+                                <desc>Mediocre and irregular hand.</desc>
                            <seg type="ink">Black, red.</seg>
                                 <seg type="rubrication">Holy names; elements of numerals and punctuation signs.</seg>
                             </handNote>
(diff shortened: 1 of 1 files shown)


MassimoVilla (changes requested):
Dear Dorothea, thank you.
The text has no internal subdivision, so it is difficult to produce a more detailed description of the content without going through the entire text page after page, which would be much time-consuming and - I think - would exceed the purpose of our catalogue. So, I think we can stick to the current description.
I am uncertain about  the transcription of some poorly readable portions in Amharic. Perhaps they should be revised by a native speaker. Has Solomon already seen them when he prepared the first draft of the description?

gnizla (approved):

ElianaDalSasso (approved):
Thank you Thea! In my opinion the description of the manuscript is fine for the catalogue, even if it is not detailed as the others. The introductory note is informative.
I have just a question, ca we give the information contained in JJ handout? I think it is interesting.

thea-m:
Can this be merged @karljonaskarlsson ?

OxfordBodleian/Juel-Jensen/BDLaethf28.xml
```diff
@@ -133,13 +174,57 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
                        <!-- <decoDesc>
 
                         </decoDesc>-->
 
+                        
+                        <additions>
+                            <list>
+                                <item xml:id="a1">
+                                    <locus target="#38"/>
+                                    <desc type="findingAid">‘Nr. 19’ written in Amharic in the upper margin.</desc>
```
karljonaskarlsson, line 196: Is "Nr." a normal translation of the Amharic? If so, I think we should use "No." instead for "number". Here and some lines below.

OxfordBodleian/Juel-Jensen/BDLaethf28.xml
```diff
@@ -133,13 +174,57 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         <handDesc>
                             <handNote xml:id="h1" script="Ethiopic">
 
+                                <desc>At least two hands were involved in the writing of the manuscript, possibly more. 
+                                    The script changes repeatedly throughout, often within the same page (e.g. <locus target="#35 #90 #115"/>), 
+                                    leading to an overall impression of irregularity. All hands belong to the nineteenth–twentieth century.</desc>
+                                <date notBefore="1800" notAfter="1998"/>
```
karljonaskarlsson, line 180: If we do not use the information from the handout, does it make sense to say "1998"? Maybe rather an even number?

karljonaskarlsson (changes requested):

PR #2293, opened 2024-02-21
Create BLorient13265

CarstenHoffmannMarburg:
I created a record according to the catalogue of S. Strelcyn. I struggled with LIT5066MiraclesMichael in ms_i5. I could not identify the exact miracle, but I can confirm on the basis of comparison of the incipits, that it is not among the twelve miracles that are recorded as LIT4155MiraclesMichael (even though the miracle on the poor man became rich seems to have a similar content). This is why, I decided to use LIT5066MiraclesMichael for this record.

### DIFF
--- a/LondonBritishLibrary/orient/BLorient13265.xml
+++ b/LondonBritishLibrary/orient/BLorient13265.xml
@@ -1,0 +1,188 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="BLorient13265" xml:lang="en" type="mss">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>ʾArdǝʾǝt, Ṣalot za-Muse, Miracles of the Virgin Mary, of Gabra Manfas Qǝddus, and of Archangel Mikāʾel</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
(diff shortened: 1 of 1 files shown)


eu-genia (approved):

eu-genia:
@CarstenHoffmannMarburg , very important:
whenever you create any record on the app, the workflow for all records is the same - download to the NEW folder, save, PR to sync. Only after records can be deleted or moved to other folders.

As I had explained before - whenever you create any record on the app, it is saved IN THE APP, in the NEW folder.
If you download the stub to a different folder (LondonBL e.g.) then at the moment of PR and merge you create TWO records in the app - the initial one which is still there in the NEW folder, and the one you upload. In this case the records CANNOT be accessed from the browser and may cause problems as there are now two files with the same `xml:id`.

The recommended workflow for manuscript records is therefore to NOT use the application for creating them but simply create a new XML file in the correct place, possibly by copypasting an existing model record.

I will now empty the content of the NEW folder in the app, but it is very important that you understand the workflow. Using APP = using NEW folder, for all types of records.

eu-genia (approved):

CarstenHoffmannMarburg:
> @CarstenHoffmannMarburg , very important: whenever you create any record on the app, the workflow for all records is the same - download to the NEW folder, save, PR to sync. Only after records can be deleted or moved to other folders.
> 
> As I had explained before - whenever you create any record on the app, it is saved IN THE APP, in the NEW folder. If you download the stub to a different folder (LondonBL e.g.) then at the moment of PR and merge you create TWO records in the app - the initial one which is still there in the NEW folder, and the one you upload. In this case the records CANNOT be accessed from the browser and may cause problems as there are now two files with the same `xml:id`.
> 
> The recommended workflow for manuscript records is therefore to NOT use the application for creating them but simply create a new XML file in the correct place, possibly by copypasting an existing model record.
> 
> I will now empty the content of the NEW folder in the app, but it is very important that you understand the workflow. Using APP = using NEW folder, for all types of records.

I will try to follow the recommended workflow and avoid saving to the wrong folder as happened in the past.

eu-genia:
the easiest, as said, is simply to create manuscript records directly in your oxygen editor, by saving a model record under a new name and editing it. do not use the app to create manuscript records.

PR #2294, opened 2024-02-22
Update BNFet149.xml

eu-genia:
added foliation line
two pages are not in the set of images (I have written to BnF)


thea-m (approved):
Thank you!

PR #2295, opened 2024-02-23
Create BLorient13223

CarstenHoffmannMarburg:
I created a new record for BLorient13223 according to the description in S. Strelcyns catalogue. For the contents I needed to create two new LIT ID as discussed in https://github.com/BetaMasaheft/Documentation/issues/2487.


eu-genia (approved):
very nice, thank you

PR #2305, opened 2024-03-05
EMDA 81 = MotGeorg001

eu-genia:
Trying to clean and map the records - in this case MotGeorg001 which is the same as EMDA 91 (https://w3id.org/vhmml/readingRoom/view/500194)


nafisa-valieva:
Dear Eugenia, thank you! the identification is correct!

eu-genia:
@nafisa-valieva just in case you refer to this record anywhere note that the xml:id has changed (the old ID remains recorded but the main link will change after the merge)

nafisa-valieva:
ok! thank you for pointing this out!

PR #2312, opened 2024-03-13
Create BLorient9036 (CH_march24)

CarstenHoffmannMarburg:
I created a record for BLorient9036 according to the description in Strelcyns catalogue.

PR #2314, opened 2024-03-14
Update BLorient12467

CarstenHoffmannMarburg:
I updated the subdivision and title markup for LIT1828Mahale, LIT2362Songof and LIT2509Weddas as discussed in https://github.com/BetaMasaheft/Manuscripts/pull/2312. 

I intended to do the same for BLorient6466, but that record was recently merged and is not yet in the master. 

Personally, I am not convinced that marking LIT2362Songof should be done in this way without consulting the manuscripts and edition. I wonder why Strelcyn has divided the text into five sections, but for each section gives a number such as (I, I - II, 7) etc. If the 'sections' in Strelcyn's description are 'canticles' according to our Bm definition, what do the Roman numbers in brackets refer to? Presumably the numbers in brackets refer to what is described as 'canticles' in the edition (and in the BM), and the 'sections' are subdivisions according to the prescribed reading in the service. Do you think I should leave it as it is, or do you think it would be better to return to the more careful encoding and describe the subdivisions as 'sections', as Strelcyn did?

### DIFF
--- a/LondonBritishLibrary/orient/BLorient12467.xml
+++ b/LondonBritishLibrary/orient/BLorient12467.xml
@@ -195,79 +195,79 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                            <note>The general title is joined to that of the first Canticle.</note>
                                             <msItem xml:id="p3_i1.2.1.1">
                                                 <locus from="129v" to="130v"/>
-                                                <title ref="LIT2277salotu"/>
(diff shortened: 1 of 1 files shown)


eu-genia:
The other way around. The Song has in LXX and in other versions 8 Chapters (I to VIII), traditionally organized in Ethiopic tradition into 5 Canticles/Sections (the Ethiopic sections are clearly longer than the standard chapters, Strelcyn always marks the correspondences). Read also the edition in https://archive.org/details/ethiopicversiono0000unse with both chapters and sections marked.

PR #2328, opened 2024-03-22
Create BLorient13268 according to S. Strelcyns catalogue

CarstenHoffmannMarburg:


LondonBritishLibrary/orient/BLorient13268.xml
```diff
@@ -0,0 +1,164 @@
...
+                    <msIdentifier>
+                        <repository ref="INS00001BL"/>
+                        <collection>Oriental</collection>
+                        <idno>BL Oriental 13268</idno>
+                        <altIdentifier><idno>Strelcyn 31</idno></altIdentifier>
+                    </msIdentifier>
+                    <msContents>
+                        <msItem xml:id="ms_i1">
+                            <locus from="3r" to="30vb"/>
+                            <title ref="LIT2701Dawit" type="incomplete"/>
```
thea-m, line 34: If all are fragments, type="incomplete" could be added to the titles, I think

thea-m (approved):

LondonBritishLibrary/orient/BLorient13268.xml
```diff
@@ -0,0 +1,164 @@
...
+                            <list>
+                                <item xml:id="a1">
+                                    <locus from="1v" to="2v"/>
+                                    <desc type="GuestText">Salām to the Virgin Mary (fragment).</desc>        
+                                </item>
+                            </list>
+                        </additions>
+                        <bindingDesc>
+                            <binding xml:id="binding">
+                                <decoNote xml:id="b1" type="unspecified">No binding.</decoNote>
```
eu-genia, line 107: I do not think `decoNote `is suitable to describe the absence of anything, I think it has more to do with condition.
I also wonder if the leaves are completely loose (unbound) or if the thing lacking are just the covers (seems more probable to me, that the description is that of a disintegrated manuscript partially fallen apart)
```
<binding>
                                <condition key="deficient">covers absent or lost</condition>
                            </binding>
```
NB for most elements no `xml:id` needed unless you refer to the element from elsewhere

eu-genia (changes requested):

LondonBritishLibrary/orient/BLorient13268.xml
CarstenHoffmannMarburg: Strelcyn literally says: "unbound". In another place, he speaks about "original flyleafs".
 I therefore assume that it is a pile of loose sheets.

eu-genia: OK then
```
<binding>
                                <condition key="other">unbound</condition>
                            </binding>
```

CarstenHoffmannMarburg:
@eu-genia 

I think, all requests are solved or clarified. Can this branch be merged?

eu-genia (approved):

PR #2341, opened 2024-03-30
Revision condition

ElianaDalSasso:
Dear all,

This is my revision of the condition element. 
I basically:
- ordered the descriptions, describing the condition of the leaves first then the condition of the binding;
- abbreviated where possible the descriptions, eliminating, for example, redundant phrases like 'the manuscript is in good condition';
- checked if the attribution of the state of conservation was coherent;
-  Aeth. c. 14: I moved a phrase regarding the textile which wraps the manuscript from `condition` to `other` in bindingDesc.

Perhaps, what can be discussed is how to deal with the condition of composite manuscripts. I suggest to keep one general element `condition` to avoid redundancy, when the different msParts show the same damage and it cannot be established a clear difference in their conservation history, like in Aeth. b. 6. Different `condition` elements may appear if the different msParts show clearly a different conservation state deriving from a different conservation history, like in Aeth. c. 14 and Aeth. d. 18.

Let me know what you think!

Eliana

### DIFF
--- a/OxfordBodleian/Juel-Jensen/BDLaethf24.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethf24.xml
@@ -10,11 +10,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader xml:base="https://betamasaheft.eu/">
         <fileDesc>
             <titleStmt>
-                <!--  
+                <!--
                 <title xml:lang="gez" xml:id="title1">መዝሙረ፡ ዳዊት፡, መስተብቍዓን፡, ፈትሐት፡ ዘወልድ፡, እሴብሕ፡ ጸጋኪ፡ ኦምልእተ፡ ጸጋ፡</title>
                 <title xml:lang="gez" corresp="#title1" type="normalized">Mazmura Dāwit, Mastabqʷǝʿān, Fǝtḥat za-wald, ʾƎsebbǝḥ ṣaggāki ʾo-mǝlʾǝta ṣaggā</title>
                 -->
-                <title xml:lang="en" xml:id="title1">Psalms of David, Supplications (<hi rendition="simple:italic">Mastabqʷǝʿān</hi>), Absolution of the Son, 
+                <title xml:lang="en" xml:id="title1">Psalms of David, Supplications (<hi rendition="simple:italic">Mastabqʷǝʿān</hi>), Absolution of the Son,
                     ‘I praise your grace, you, full of grace’</title>
                 <editor key="DR"/>
                 <editor key="JG"/>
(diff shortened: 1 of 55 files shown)


OxfordBodleian/Juel-Jensen/BDLaethf24.xml
```diff
@@ -100,32 +100,31 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                     </list>
                                 </collation>
                                 <condition key="deficient">
-                                    The manuscript is in bad condition. The front board is broken in two halves and the outer is missing, together with the back board.
-                                    A large loss is present in the central area of <locus target="#1"/>.
+                                    A large material loss is present in <locus target="q1"/> a reused bifolio, in the central area of <locus target="#1"/>.
```
MassimoVilla, line 103: perhaps "in the `<locus target="q1">first quire</locus>`, a reused bifolio"?

OxfordBodleian/Juel-Jensen/BDLaethf24.xml
```diff
@@ -152,25 +151,25 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
                     <history>
                         <origin>
                             <origDate notBefore="1800" notAfter="1949" evidence="lettering"/>
-                            The manuscript probably consists of three distinct units of production: <ref target="#p1"/> and <ref target="#p2"/>, 
+                            The manuscript probably consists of three distinct units of production: <ref target="#p1"/> and <ref target="#p2"/>,
                             and <ref target="#p3"/>. <ref target="#p3"/>
                             consists of the addition of content on the material support of <ref target="#p1"/> and <ref target="#p2"/>
-                            and of the addition of one more quire, <ref target="#q6"/> (corresponding to the 
-                            transformation model A3 according to <bibl><ptr target="bm:Andrist2013Syntaxe"/><citedRange unit="page">65</citedRange></bibl>). 
+                            and of the addition of one more quire, <ref target="#q6"/> (corresponding to the
```
MassimoVilla, line 163: Will the element `<ref target="#q6"/>` in this position generate any problems in the PDF output? @thea-m

OxfordBodleian/Juel-Jensen/BDLaethg28.xml
```diff
@@ -117,13 +117,13 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
-                                <condition key="deficient">The original sewing is broken and it has been repaired by using a different kind of thread. <!--A brown liquid
+                                <condition key="deficient"><!--A brown liquid
                                      was spilled on the manuscript, see especially on <locus
                                         from="15r" to="16v"/> I think that a different kind of ink, a brownish one, has been used to draw
                                       the figures on f. 15v and f. 16v later damaged by humidity. The same ink is on f. 1v, 2r, 9r, 9v, 10r, 10v, 11r, 11v.-->
-                                        Smudged ink on many folios see especially on <locus
-                                           from="15r" to="16v"/>.</condition>
-
+                                        Smudged ink on many folios, see especially on <locus from="15r" to="16v"/>. Insect droppings and humidity stains
+                                        on most leaves. The original sewing is broken with the risk
```
MassimoVilla, line 125: Did we agree to use "folio/folios" everywhere instead of "leaf/leaves"? Can this be changed automatically at a later stage or should be done now?

OxfordBodleian/Juel-Jensen/BDLaethg35.xml
```diff
@@ -213,8 +213,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                         </item>
                                         </list>
                                 </collation>
-                                <condition key="good">The manuscript folios are in good condition and show only slight traces of use.
-                                  The sewing has become undone since the sewing thread broke in many points. Both boards are detached from the bookblock, together with <ref target="#q1"/>.</condition>
+                                <condition key="good">The manuscript folios show only slight traces of use.
```
MassimoVilla, line 216: Is "manuscript folios" somehow redundant? Perhaps "the bookblock" or "the manuscript"?

OxfordBodleian/Juel-Jensen/BDLaethg41.xml
```diff
@@ -171,7 +171,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                     </dimensions>
                                     <measure type="weight" unit="g">28</measure>
                                 </extent>
-                                <condition key="good">The state of preservation is good. Small areas of loss in the upper part of the scroll. The lower part of the scroll has been crumpled by rolling,
+                                <condition key="good">Small areas of loss are present in the upper part of the scroll. The lower has been crumpled by rolling,
```
MassimoVilla, line 174: "the lower part"

MassimoVilla (changes requested):
Thank you Eliana, information on the condition is so rich!
I have added very few remarks, and one general question: concerning the order of the descriptive elements, did we agree to start from the outside towards the inside (so binding first, then the leaves)? This is the order I followed in my MSS, I hope it didn't cost you too much effort to reverse the descriptive parts during the revision process.

OxfordBodleian/Juel-Jensen/BDLaethf24.xml
thea-m: I think so, yes, as will all other empty refs...it is on the list of PDF problems

thea-m (approved):
Thank you! My only question (aside from Massimo's) is: Do you want the current general evaluation of the condition encoded as a value ("intact", "good" etc) to be printed in the catalogue?

ElianaDalSasso:
Thank you @MassimoVilla and @thea-m!

@MassimoVilla - sorry, I was unaware of this decision. I put the condition of the folios first because I thought that the main concern of the readers will be the texts...but it did not take long, do you think it is better the other way around?

@thea-m - yes, please! If possible it would be good to have it printed. Otherwise I can add it in the description.

MassimoVilla:
> Thank you @MassimoVilla and @thea-m!
> 
> @MassimoVilla - sorry, I was unaware of this decision. I put the condition of the folios first because I thought that the main concern of the readers will be the texts...but it did not take long, do you think it is better the other way around?
> 
> @thea-m - yes, please! If possible it would be good to have it printed. Otherwise I can add it in the description.

No please, if all descriptions agree with this order, better to leave them as they are. I think uniformity is the most important issue. Thank you!

MassimoVilla (approved):

ElianaDalSasso:
Thank you all! Shall I merge? I vaguely remember that Jacopo will revise everything in the end and Jonas for the moment deals with his manuscripts only, do I remember correctly?

PR #2356, opened 2024-04-04
Dr 24 04 dmg yi135 iiib1

thea-m:
Let me know if you think a STU record is necessary :)


eu-genia (approved):
thank you, interesting.
I am a bit ambivalent re: STU, so far we have used it for other things but still it possibly fits

PR #2367, opened 2024-04-10
BAVet141

thea-m:


VaticanBAV/et/BAVet141.xml
```diff
@@ -0,0 +1,367 @@
...
+                          <handNote script="Ethiopic" xml:id="h1">
+                            <seg type="ink">Black, red.</seg>
+                            <seg type="rubrication">Titles.</seg>
+                            <date notBefore="1800" notAfter="1899"/>
+                          </handNote>
+                        </handDesc>
+                      
+                    <decoDesc>
+                      <decoNote type="ornamentation" xml:id="d1">
+                        <desc>The Psalms are divided into ten lines, which are marked alternately with red and black dots.</desc>
```
DenisNosnitsin1970, line 287: I would say, each ten lines of of each psalm is marked by alternatively red of black dots in the margin. The current wording is not clear.

DenisNosnitsin1970 (changes requested):

VaticanBAV/et/BAVet141.xml
```diff
@@ -284,7 +284,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                       
                     <decoDesc>
                       <decoNote type="ornamentation" xml:id="d1">
-                        <desc>The Psalms are divided into ten lines, which are marked alternately with red and black dots.</desc>
+                        <desc>The catalogue statement ‘Psalmi deni lineis dividuntur,
+                          quae constant punctis alterna vice rubris et nigris.’ probably indicates that 
+                          every ten lines of  each psalm are marked by alternativ red or black dots in the margin.</desc>
```
CarstenHoffmannMarburg, line 289: 'alternate' instead of 'alternativ'

VaticanBAV/et/BAVet141.xml
```diff
@@ -284,7 +284,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                       
                     <decoDesc>
                       <decoNote type="ornamentation" xml:id="d1">
-                        <desc>The Psalms are divided into ten lines, which are marked alternately with red and black dots.</desc>
+                        <desc>The catalogue statement ‘Psalmi deni lineis dividuntur,
+                          quae constant punctis alterna vice rubris et nigris.’ probably indicates that
```
CarstenHoffmannMarburg, line 288: Can this be marked as Latin text?

thea-m: I decided against it, because this would them make this manuscript turn up in any searches filtering for Latin text in manuscripts, even when the manuscript itself does not contain any Latin...not ideal

PR #2403, opened 2024-04-30
Emip repository 5

JanetFre:


eu-genia (approved):
NB the ref to EMIP repository stays in altIdentifier so that EMIP repositories in https://github.com/BetaMasaheft/Documentation/issues/2516 remain linked

if facs is missing it is good to add it from https://github.com/BetaMasaheft/Documentation/issues/2017

PR #2409, opened 2024-05-03
Edits to Transkribus Output for Tweed Magic Scrolls

Ralph-Lee-UK:
This is a first attempt at correcting the output from the XSL file to convert output from Transkribus to BM style for the Tweed Magic Scrolls


eu-genia:
Sorry I will recheck later tonight maybe did not store properly... I am not at my desk and cannot access the data now

eu-genia:
It is stored OK, the version I see is the new one with this funny "Ask Ralph about this፡" remark (it was not present in the earlier version)

PR #2419, opened 2024-05-08
Create BLorient11802_CH_may24

CarstenHoffmannMarburg:
I struggled with Strelcyn's statement that "an unnumbered leaf was inserted before (last) folio 28v". I decided to define the leaf, i.e. 28 in the manuscript, as 29 in my description and to explain it in the foliation. 

Screenshot 2024-05-08 142035

### DIFF
--- a/LondonBritishLibrary/orient/BLorient11802.xml
+++ b/LondonBritishLibrary/orient/BLorient11802.xml
@@ -51,8 +51,6 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                         <width>115</width>
                                     </dimensions>
                                 </extent>
-                                <foliation>One unnumbered leaf between the penultimate (<locus target="#27">no. 27</locus>) 
-                                    and the last leaf (<locus target="#29">no. 28 in the manuscript, 29 in this discription</locus>).</foliation>
                             </supportDesc>
                             <layoutDesc>
                                 <layout columns="2" writtenLines="19"/>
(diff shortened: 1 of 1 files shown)


eu-genia:
one could also describe the unnumbered blank leaf as accMat

LondonBritishLibrary/orient/BLorient11802.xml
```diff
@@ -0,0 +1,165 @@
...
+                        </additions>
+                        <bindingDesc>
+                            <binding xml:id="binding">
+                                <decoNote xml:id="b1" type="Boards"><material key="wood"/>Wooden boards.</decoNote>
+                                <decoNote xml:id="b2" type="Cover"><material key="leather"/>The leather back was added during repair in the
+                                    <placeName ref="INS00001BL">British Museum</placeName>.</decoNote>
+                            </binding>
+                        </bindingDesc>
+                        <accMat>An unnumbered blank leaf has been added to the manuscript between the penultimate (<locus target="#27">no. 27</locus>)
+                            and the last leaf (<locus target="#29">no. 28 in the manuscript, 29 in this discription</locus>).</accMat>
```
eu-genia, line 110: description

thea-m (approved):

thea-m:
We have traditionally described such cases in foliation - I'm not saying that this is always conceptually better than accMat, but it is where I would look for this information. If the leaf is bound together with the manuscript (we don't seem to know it here), I think that would be a bit clearer than accmat

PR #2430, opened 2024-05-17
Create BLorient3716a

CarstenHoffmannMarburg:
Strelcyn reported, that this scroll was preserved in a box together with many other items. I decided to encode this information in binding desc. Do you think, this is a suitable place for that?

### DIFF
--- a/LondonBritishLibrary/orient/BLorient3716a.xml
+++ b/LondonBritishLibrary/orient/BLorient3716a.xml
@@ -115,6 +115,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 <keywords>
                     <term key="ChristianLiterature"/>
                     <term key="Poetry"/>
+                    <term key="Magic"/>
                 </keywords>
             </textClass>
             <langUsage>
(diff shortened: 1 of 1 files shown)


LondonBritishLibrary/orient/BLorient3716a.xml
```diff
@@ -0,0 +1,134 @@
...
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p/>
+            </abstract>
+            <textClass>
+                <keywords>
+                    <term key="ChristianLiterature"/>
+                    <term key="Poetry"/>
```
DenisNosnitsin1970, line 117: I would add "poetry" keyword.

DenisNosnitsin1970 (changes requested):

LondonBritishLibrary/orient/BLorient3716a.xml
CarstenHoffmannMarburg: There is already "Poetry" in the list. Do you mean "Chants"?

```
<keywords>
                    <term key="ChristianLiterature"/>
                    <term key="Poetry"/>
                </keywords>
```

eu-genia, line 117: I wonder if Magic should be added as a keyword considering the protective text

eu-genia (approved):

PR #2436, opened 2024-05-27
Create BLorient8818

CarstenHoffmannMarburg:
I created a record for BLorient8818. This is a manuscript, in which a text of Śǝna fǝṭrat (LIT2314senafe) is placed in between the first and the second chapter of LIT1097Ammest. I was puzzled, whether this text should be nested inside ms_i1 as part of LIT1097Ammest or kept independantly. 

For ms_i4, I do not know, whether it is reasonable to create a general record for "List of Patriarchs of Alexandria". Since there is no further information, I do not know, whether it is probably related to the accounts of Walda Amid or Abušāker. 

I found information on B. (= J.) R. T. Lieder. I understood this, as Lieder bought this manuscript and sold it to Curzon, who brought it to England:

Screenshot 2024_Lieder

### DIFF
--- a/LondonBritishLibrary/orient/BLorient8818.xml
+++ b/LondonBritishLibrary/orient/BLorient8818.xml
@@ -1,0 +1,220 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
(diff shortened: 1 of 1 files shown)


eu-genia:
I would say
either we encode/display

1 "5 pillars"
1.1
1.2 
1.3
1.4

2 Beauty of creation, with a note that it is inserted among the pillars (as also clear from the locus)
3 List

or,, if this compilation is something that occurs more than once, then it deserves a collection ID of its own, that is, we create a CAe ID for the Pillars + Beauty compilation (LIT1234) and then have

1. Pillars+Beauty (LIT1234)
1.1. LIT1234#PillarsIntro
1.2 LIT1234#Pillar1
1.3 LIT1234#Senafetrat
1.4 LIT1234#Pillar2
etc

eu-genia (approved):

PR #2440, opened 2024-05-29
Create BLorient11718A

CarstenHoffmannMarburg:
This manuscript has no uniform pagination or foliation. The first part is numbered from page 1 to 82, plus unnumbered and unwritten pages which are most likely at the end (except for one after page 1). The other part consists of the twice typewritten translation of the above text. I did not dare to transform the foliation of the latter according to the pagination of the former or vice versa. I do not think that two systems of pagination / foliation will work in one record. For this reason, I decided not to provide any loci for p2, but to explain the setup explicitly.

I also wondered whether the same text in p2 twice deserves to be defined as p2 and p3. However, Strelcyn did not give any details for the latter copy other than what I noted in the note.

### DIFF
--- a/LondonBritishLibrary/orient/BLorient11718A.xml
+++ b/LondonBritishLibrary/orient/BLorient11718A.xml
@@ -48,6 +48,16 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                 <foliation>The volume has no unified pagination, but each codicological unit has its own folio or page numbers.</foliation>
                             </supportDesc>  
                         </objectDesc>
+                        <additions>
+                            <list>
+                                <item xml:id="a1">
+                                    <locus target="#1"/>
+                                    <desc type="AcquisitionNote">Acquisition note on the verso of the last leaf.</desc>
+                                    <q xml:lang="en">Presented by <persName ref="PRS14446Douglas">Canon John A. Douglas</persName>. 
+                                        <date when="1941">Jan. 1941</date></q>
+                                </item>
+                            </list>
+                        </additions>
                         <bindingDesc>
                             <binding xml:id="binding" contemporary="false">
                                 <decoNote xml:id="b1" type="Boards">European binding provided by the <placeName ref="INS00001BL"/>.</decoNote>
@@ -56,10 +66,6 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     </physDesc>
                     <history>
                         <origin><origDate notBefore="1923" notAfter="1924">1923/1924</origDate></origin>
-                        <provenance></provenance>
-                        <acquisition>On the verso of the last leaf - a note: 
-                            'Presented by <persName ref="PRS14446Douglas">Canon John A. Douglas</persName>. 
-                            <date when="1941">Jan. 1941</date>'</acquisition>
                     </history>
                     <additional>
                         <adminInfo>
@@ -126,7 +132,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                 <list>
                                     <item xml:id="a1">
                                         <locus target="#1"/>
-                                        <desc>A pencil note by <persName ref="PRS5226haylase"><roleName type="title">rās</roleName> 
+                                        <desc type="Letter">A pencil note by <persName ref="PRS5226haylase"><roleName type="title">rās</roleName> 
                                             Tafari Makʷannǝn</persName>: 'Answers given by Ethiopian Doctors to the questions addressed by  
                                             <persName ref="PRS14445BishopGwynne">Bishop Gwynne of Khartoum</persName> to the Heir 
                                             to the Throne of Ethiopia <persName ref="PRS5226haylase">Tafari Makʷannǝn</persName>. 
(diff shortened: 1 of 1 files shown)


LondonBritishLibrary/orient/BLorient11718A.xml
```diff
@@ -0,0 +1,226 @@
...
+                                    <desc>Large careful handwriting.</desc>
+                                    <date notBefore="1923" notAfter="1924"/>
+                                </handNote>
+                            </handDesc>
+                            <additions>
+                                <list>
+                                    <item xml:id="a1">
+                                        <locus target="#1"/>
+                                        <desc>A pencil note by <persName ref="PRS5226haylase"><roleName type="title">rās</roleName> 
+                                            Tafari Makʷannǝn</persName>: 'Answers given by Ethiopian Doctors to the questions addressed by
```
DenisNosnitsin1970, line 130: I would shape this note as other a-additional notes. Attribute `unclear` can be used, if it does not fit the system of the genres; is "Answers..." you translation or continuation of the note? (if so, I would not split it).

LondonBritishLibrary/orient/BLorient11718A.xml
```diff
@@ -0,0 +1,226 @@
...
+                        <msIdentifier/>
+                        <msContents>
+                            <summary/>
+                            <msItem xml:id="p1_i1">
+                                <locus from="2" to="82"/>
+                                <title>Answers given by Ethiopian Doctors to the questions addressed by Bishop Gwynne of Khartoum</title>
+                                <msItem xml:id="p1_i1.1">
+                                    <locus from="2" to="8"/>
+                                    <title>List of sixty questions</title>
+                                    <incipit xml:lang="am"><title>የ፷ጠያቂዎች፡ አርእስት።</title>
```
DenisNosnitsin1970, line 88: please check if it is not ጥያቄዎች

LondonBritishLibrary/orient/BLorient11718A.xml
```diff
@@ -0,0 +1,226 @@
...
+                            <binding xml:id="binding" contemporary="false">
+                                <decoNote xml:id="b1" type="Boards">European binding provided by the <placeName ref="INS00001BL"/>.</decoNote>
+                            </binding>
+                        </bindingDesc>
+                    </physDesc>
+                    <history>
+                        <origin><origDate notBefore="1923" notAfter="1924">1923/1924</origDate></origin>
+                        <provenance></provenance>
+                        <acquisition>On the verso of the last leaf - a note: 
+                            'Presented by <persName ref="PRS14446Douglas">Canon John A. Douglas</persName>.
```
DenisNosnitsin1970, line 61: I would shape this as a separate additional note.

LondonBritishLibrary/orient/BLorient11718A.xml
```diff
@@ -0,0 +1,226 @@
...
+                    <msPart xml:id="p1">
+                        <msIdentifier/>
+                        <msContents>
+                            <summary/>
+                            <msItem xml:id="p1_i1">
+                                <locus from="2" to="82"/>
+                                <title>Answers given by Ethiopian Doctors to the questions addressed by Bishop Gwynne of Khartoum</title>
+                                <msItem xml:id="p1_i1.1">
+                                    <locus from="2" to="8"/>
+                                    <title>List of sixty questions</title>
```
DenisNosnitsin1970, line 87: No ID for this text?

DenisNosnitsin1970 (changes requested):

LondonBritishLibrary/orient/BLorient11718A.xml
CarstenHoffmannMarburg: I think, it is letter (a1) and comment (a2, previously e1).

LondonBritishLibrary/orient/BLorient11718A.xml
CarstenHoffmannMarburg: It is like that in Strelcyns transcription:

Screenshot 2024_Tayaqewoch

LondonBritishLibrary/orient/BLorient11718A.xml
CarstenHoffmannMarburg: I understood, that this is a letter from the community of scholars of the Ethiopian Church around Ras Tafari to Bishop Gwynn in Khartoum. I did not expect any circulation inside Ethiopia for it and did not investigate much until now.

LondonBritishLibrary/orient/BLorient11718A.xml
DenisNosnitsin1970: For sure, this is (the cataloguer's) mistake. ጠያቂ is just a different word from what is intended, ጠ instead of ጥ might have been a typo. -i read instead of -e is a typical mistake as -i and -e are more difficult to distinguish in some modern hands. Maybe just mark it `<sic>`.

CarstenHoffmannMarburg: Do you mean, I should silently correct the ጠ to ጥ or mark the entire word with sic?

eu-genia (approved):

PR #2447, opened 2024-06-07
Create BLorient11718B according to Strelcyn

CarstenHoffmannMarburg:
This record regularly was saved to to the PRS>new folder. I needed to transfer it to BetMas>Manuscripts>LondonBritishLibrary>orient before pushing it. I do not know, what was wrong. I hope, it caused no troubles in the app.


eu-genia:
If it was only saved locally and never committed the app has no idea of it

eu-genia (changes requested):

eu-genia (approved):

PR #2453, opened 2024-06-11
2022 09 15 j kd19

karljonaskarlsson:
Hello!

Here comes a new manuscript description from me. It is an old Psalter manuscript, so if you want to check my description and transcriptions against the actual manuscript images, it will be a pleasant task :)

I analyse it as a composite manuscript and hope to have argued reasonably well for this in the description, because it is not an entirely obvious case (for example, the hand seems to be the same in both parts).

Thank you all for your reviews when you find the time (no stress!)!

### DIFF
--- a/OxfordBodleian/Juel-Jensen/BDLaethd19.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethd19.xml
@@ -52,7 +52,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         
                         <msItem xml:id="ms_i1.1">
                             <locus from="1r" to="4v"/>
-                            <title type="complete"/>
+                            <title type="complete" ref="LIT6719prologue"/>
                             <incipit xml:lang="gez">
                                 <locus target="#1r"/>
                                 <lb/><hi rend="rubric"></hi>
--- a/OxfordBodleian/Juel-Jensen/BDLaethd19.xml
+++ b/OxfordBodleian/Juel-Jensen/BDLaethd19.xml
@@ -46,10 +46,6 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     <msContents>
                         <summary/>
                         
-                        <!--<msItem xml:id="ms_i1">
-                      
-                        </msItem>-->
-                        
                         <msItem xml:id="ms_i1.1">
                             <locus from="1r" to="4v"/>
                             <title type="complete" ref="LIT6719prologue"/>
@@ -58,15 +54,15 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                 <lb/><hi rend="rubric"><gap reason="illegible"></gap> <unclear>ሥር<supplied reason="undefined">አ</supplied>ት፡</unclear> ዘአው<unclear>ሰ</unclear>ብስ፡ <unclear>ና</unclear>ስጥናዊ፡ መጽ
                                     <lb/><supplied reason="lost">ሐ</supplied>ፈ፡ መዝሙር፡ ወግጻ<unclear>ዌ</unclear>ሁ፡ <unclear>ዘ</unclear>በበ፡ ህልው፡ ዘከ
                                     <lb/><supplied reason="lost">መ፡</supplied> ተጽሕፈ፡ ዘበዕብራይስጥ፡ ወዘከመ፡ አዕለውዎ<supplied reason="omitted">፡</supplied></hi>
-                                    <lb/>እለ፡ ኢኮኑ፡ ጽሑፋን፨
-                                <lb/>ወጽሑፋን፨
+                                <lb/>እለ፡ ኢኮኑ፡ ጽሑፋን<hi rend="rubric" rendition="#partialRubric">፨</hi>
+                                <lb/>ወጽሑፋን<hi rend="rubric" rendition="#partialRubric">፨</hi>
                             </incipit>
                             <explicit xml:lang="gez">
                                 <locus target="#4v"/>
                                 <lb/>መጽሐፈ፡ ቅዱሳን፡ ወሕጉ፡ ለእግዚአብሔር፡ እም
                                 <lb/>ድር፡ ወእምሰማይ፡ ኪያከ፡ እግዚኦ፡ ነአኵት፡ ለአብ፡
                                 <lb/>ወወልድ፡ ወመንፈስ፡ ቅዱስ፡ ወማእሰረ፡ ተፍጻሜ
-                                <lb/>ቱ፡ ፍቅር፡ ለዓለመ፡ ዓለም፡ አሜን፨
+                                <lb/>ቱ፡ ፍቅር፡ ለዓለመ፡ ዓለም፡ አሜን<hi rend="rubric" rendition="#partialRubric">፨</hi>
                             </explicit>
                         </msItem>
                         
@@ -76,15 +72,15 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                             <incipit xml:lang="gez">
                                 <locus target="#7r"/>
                                 <lb/><hi rend="rubric">ፍካሬ፡ ዘጻድቃን፡ ወኀጥአን።</hi>
-                                <lb/><hi rend="rubric">፩፡</hi> ብፁዕ፡ ብእሲ፡ ዘኢሖረ፡ በምክረ፡ ረሲዓን፨
-                                <lb/>ወኢቆመ፡ ውስተ፡ ፍኖተ፡ ኃጥአን፨
+                                <lb/><hi rend="rubric">፩፡</hi> ብፁዕ፡ ብእሲ፡ ዘኢሖረ፡ በምክረ፡ ረሲዓን<hi rend="rubric" rendition="#partialRubric">፨</hi>
+                                <lb/>ወኢቆመ፡ ውስተ፡ ፍኖተ፡ ኃጥአን<hi rend="rubric" rendition="#partialRubric">፨</hi>
                             </incipit>
                             <explicit xml:lang="gez">
-                                <locus target="#107r"/>
-                                <lb/>ወረገሙኒ፡ በአማልክቲሆሙ፨
-                                <lb/>ወአንሰ፡ ነሣእኩ፡ ዘእምላዕሌሁ፡ ሰይፈ፨
-                                <lb/>ወመተርኩ፡ ርእሶ፨
-                                <lb/>ወአሰሰልኩ፡ ፅእለተ፡ እምደቂቀ፡ እስራኤል፨
+                                <locus target="#125r"/>
+                                <lb/>ወረገሙኒ፡ በአማልክቲሆሙ<hi rend="rubric" rendition="#partialRubric">፨</hi>
+                                <lb/>ወአንሰ፡ ነሣእኩ፡ ዘእምላዕሌሁ፡ ሰይፈ<hi rend="rubric" rendition="#partialRubric">፨</hi>
+                                <lb/>ወመተርኩ፡ ርእሶ<hi rend="rubric" rendition="#partialRubric">፨</hi>
+                                <lb/>ወአሰሰልኩ፡ ፅእለተ፡ እምደቂቀ፡ እስራኤል<hi rend="rubric" rendition="#partialRubric">፨</hi>
                             </explicit>
                         </msItem>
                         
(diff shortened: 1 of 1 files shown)


OxfordBodleian/Juel-Jensen/BDLaethd19.xml
```diff
@@ -59,21 +53,22 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
                                     </dimensions>
-                                    <dimensions type="leaf" unit="mm" xml:id="leafdim">
-                                        <height>284</height>
-                                        <width>206</width>
-                                    </dimensions>
-                                    <note corresp="#leafdim">Data on the dimensions of folios taken from <locus target="#8r"/>.</note>
-                                    <measure type="weight" unit="g">3340</measure><!--weight variates-->
+
                                 </extent>
+                                <foliation>Pencil foliation, mostly in the upper right corner.</foliation>
```
thea-m, line 62: "Pencil foliation in the upper right corner (1–165)." according to the revision

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
```diff
@@ -184,45 +183,102 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                <desc>
+                                    Fine and trained hand. Characters are regularly spaced.
+                                    The script displays some relatively archaic features: the numeral <foreign xml:lang="gez">፮</foreign> has 
+                                    a compressed shape and the open loop;
+                                    the letter <foreign xml:lang="gez">መ</foreign> has the loops not completely separated; the 
+                                        letter <foreign xml:lang="gez">ዐ</foreign> and 
+                                    numeral <foreign xml:lang="gez">፬</foreign> show "cone-forms" (for the letter 
+                                    <foreign xml:lang="gez">ፀ</foreign>, this is much less 
+                                    salient); the letter <foreign xml:lang="gez">ኍ</foreign> has the vowel marker on the top of the letter body; 
+                                    however, the word <foreign xml:lang="gez">እግዚአ፡ ብሔር፡</foreign> is written as one word. Although the
```
thea-m, line 201: If እግዚአ፡ ብሔር፡ is written as one word, maybe it would be better to use the same spelling here

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
```diff
@@ -184,45 +183,102 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                
+                                
+                                <item xml:id="a4">
+                                    <locus target="#163r"/>
+                                    <desc type="Unclear">Modern scribbles in Ethiopic script in the lower margin.</desc>
+                                </item>
+                                
+                                <item xml:id="e1">
+                                    <desc>A white sticker is glued to the inner side of the front board. The name of <persName ref="PRS5782JuelJen"/> is printed on it
+                                        (<foreign xml:lang="gez">ቤንት፡ ዩውል፡ የንሰን።</foreign>) and it bears partially legible stamp mark which includes the
```
thea-m: it bears **a** partially legible stamp mark?

thea-m: `(‘<foreign xml:lang="gez">ቤንት፡ ዩውል፡ የንሰን።</foreign>’)`

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
```diff
@@ -456,26 +512,958 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                <ab type="ruling">The upper line seems to be written above the ruling, although it is frequently difficult to
+                                    determine with certainty. <!-- ??? --> The bottom line is written above the ruling.
+                                </ab>
+                            </layout>
+                            </layoutDesc>
+                        </objectDesc>
+                        
+                        <decoDesc>
+                            
+                            <decoNote xml:id="d1" type="miniature">
```
thea-m, line 936: I think there is still a trace of a branch conflict here. The miniatures are already described by Jacopo in the common part. Can their list be deleted here? Or should their description be moved from the common part to p1?

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
```diff
@@ -456,26 +512,958 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                    <msContents>
+                        <summary/>
+                        
+                        <msItem xml:id="p2_i1">
+                            <locus from="148ra" to="149vb"/>
+                            <title type="incomplete" ref="LIT2715John"/>
+                            <note>The text is incomplete and contains Jn 20:20–21:21, ending abruptly in the middle of the last verse.</note>
+                            <incipit xml:lang="gez">
+                                <locus target="#148ra"/>
+                                <lb/>እገሪሁ፡ ወገቦሁ<hi rend="rubric" rendition="#partialRubric">፨</hi> ወ
```
thea-m, line 1081: typically, `<lb/> ` has not been used in two-column layout. We might discuss where we want a line break to be printed

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
```diff
@@ -456,26 +512,958 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                   </layout>
+                            </layoutDesc>
+                    
+                        </objectDesc>
+                    
+                        <additions>
+                            <list>
+                                <item xml:id="a5">
+                                    <locus target="#150r"/>
+                                    <desc type="Unclear">In the blank spaces of the headpiece decorating the beginning of the
```
thea-m, line 1384: I would reformulate: Barely legible note....in the blank spaces....

thea-m (changes requested):
Thank you, it is a wonderful description! Some small remarks, mostly related to standardising

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
```diff
@@ -456,26 +506,958 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                            </decoNote>
+                            
+                        </decoDesc>
+                        
+                        <additions>
+                            <list>
+
+                                
+                                <item xml:id="e3">
+                                    There are frequent line takeups and line takedowns (<locus target="#1v #7r #8r"/>), occasional erasures
```
MassimoVilla, line 1053: `(e.g. <locus target="#1v #7r #8r"/>)` (I would add e.g., what do you think?)

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
```diff
@@ -456,26 +506,958 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                            
+                        </decoDesc>
+                        
+                        <additions>
+                            <list>
+
+                                
+                                <item xml:id="e3">
+                                    There are frequent line takeups and line takedowns (<locus target="#1v #7r #8r"/>), occasional erasures 
+                                    (<locus target="#15v #16v #17v #22r #35v #70v #89r"/>), and interlinear additions (<locus target="#47r #54v"/>).
```
MassimoVilla, line 1054: Isn't three the maximum number of instances per phenomenon? I would keep three representative instances and add "e.g."

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
```diff
@@ -414,6 +421,49 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                             </decoNote>
 
                         </decoDesc>
+                        
+                        <additions>
+                            <list>
+                                <item xml:id="a1">
+                                    <desc type="Unclear">On the leather fold on the inside of the cover, there are a few barely legible letters in
```
MassimoVilla, line 428: I would reformulate: There are a few barely legible ... on the leather fold on ....

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
```diff
@@ -456,26 +506,958 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                                            <lb/>ንፈስ፡ ዚኢገስስዎ፡
+                                            <add place="above">ክቡት፡ ዘኢይሬእይዎ፡ እሳት፡ ዘኢይለክፍዎ፡</add>
+                                            <lb/>ኃይል፡ <choice><orig>ዚ</orig><corr>ዘ</corr></choice>ኢይሌብው
+                                            <lb/>ዎ፡ ንጉሥ፡ ኢይሰይም
+                                            <lb/>ዎ፡ አምላክነ፡ በመለ<supplied reason="lost">ኮ</supplied>
+                                            <lb/>ቱ፡ ወአቡነ፡ በኂሩቱ፡ <add place="inline"><unclear>ዘ</unclear></add>አ
+                                            <lb/>ሐደ፡ ይሰገድ፡ <add place="inline">ወ</add>አሐደ፡ ይሴ
+                                            <lb/>ባሕ፡ <add place="inline">ወ</add>አሐደ፡ ይጼለይ፡ ሎ
+                                            <lb/>ቱ፡ ስብሐት፡ ለዓለመ፡ ዓ
+                                            <lb/>ለም፡ አሜን<hi rend="rubric" rendition="#partialRubric">፨=፨</hi>
```
MassimoVilla, line 1414: I think we opted for reducing this punctuation mark ፨=፨ to simple ፨ in our descriptions, but I might be wrong. @thea-m

MassimoVilla (changes requested):
Thank you Jonas for this very detailed description! My comments are few and on secondary issues. On the subdivision of the MS into two original parts and the material traces of their binding, Eliana might tell more.

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
thea-m: I'm afraid I don't remember and cannot find any notes on this in our documentation...

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
thea-m: (or even "A few barely ....on the leather fold...")

MassimoVilla: Yes, I think this depends on the standardization of the syntax during the revision process. We'll see if it is more convenient to remove verbs everywhere or to add them and make full sentences everywhere.

ElianaDalSasso (changes requested):
Thank you Jonas for the detailed description! I have just a minor terminological remark, the leather fold on which there are the barely legible letters is a turn-in :)

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
karljonaskarlsson: Does it cause a problem to have it encoded like this? I was thinking that it won't be rendered

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
karljonaskarlsson: They should be in p1. Is there a way to compare the text in the different parts? I'm not sure if I revised the captions here or not

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
karljonaskarlsson: It seems like the string ' ፨=፨ ' has not been used elsewhere in J-J, so I'll take it away and assume that you have been standardizing :)

karljonaskarlsson:
I think I have addressed all points now, except the two questions I had for Dorothea above :)

ElianaDalSasso (approved):

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
thea-m: Let's keep it for now, it would be a pity to lose the information

thea-m:
Thanks a lot! I have merged and updated the cataloguing and revision overviews, so that this manuscript can be included in all ongoing and future revision categories

OxfordBodleian/Juel-Jensen/BDLaethd19.xml
karljonaskarlsson: It would not be the end of the world, but if it doesn't hurt, it can be there :)

karljonaskarlsson:
Thank you, Dorothea! I'm not sure if we solved the double descriptions of the decorations now… but if not, we will notice it later. It will not be something that falls between the chairs. Thank you!

PR #2457, opened 2024-06-13
Doubled-checked and changed all "supplied reason=lost"

karljonaskarlsson:
Hello!

Here comes the mini branch in which I got rid of "supplied reason=lost", based on re-checking the manuscripts where it had been used.

Have a nice evening!
Jonas


thea-m (approved):
Thank you, Jonas!

PR #2468, opened 2024-06-19
Create BLorient4849.xml

CarstenHoffmannMarburg:
ms_i1 as discussed in https://github.com/BetaMasaheft/Documentation/issues/2563.

In ms_i4, I added new sub-IDs to the synaxary for 3rd Pagumen. I will update the work record of LIT4035SenkessarDL as soon as the current branch (cf. https://github.com/BetaMasaheft/Works/pull/1122) has been merged to the master to avoid conflicts.

### DIFF
--- a/LondonBritishLibrary/orient/BLorient4849.xml
+++ b/LondonBritishLibrary/orient/BLorient4849.xml
@@ -1,0 +1,608 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="BLorient4849" xml:lang="en" type="mss">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>Homilies for festivals of St Michael and miracles of this archangel (Dǝrsāna Mikāʾel)</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                    Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                        licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <msDesc xml:id="ms">
+                    <msIdentifier>
+                        <repository ref="INS00001BL"/>
+                        <collection>Oriental</collection>
+                        <idno>BL Oriental 4849</idno>
+                        <altIdentifier><idno>Or. 4849</idno></altIdentifier>
+                        <altIdentifier><idno>Strelcyn 58</idno></altIdentifier>
+                    </msIdentifier>
+                    <msContents>
+                        <msItem xml:id="ms_i1">
+                            <locus from="21r" to="32r"/> 
+                            <title ref="LIT3961Homily"/>  
+                            <note>The name of the saint is missing, the spaces reserved for the rubrics being left blank.</note>
+                            <incipit xml:lang="gez">ወእምድኅረ፡ ተጠምቁ፡ ነበሩ፡ ወርኃ፡ ፍጹመ፡ ኀበ፡ ጳጳስ፡ እንዘ፡ ይሜህሮሙ፡ ነገረ፡ ሃይማኖት፡ ርትዕት፡ ወማቴዎስ፡ ነግድ፡ እምብዝኃ፡
+                                ፍሥሐሁ፡ ወሀበ፡ ኵሎ፡ ዘረከበ፡ ለቤተ፡ ክርስቲያን፡ ለሊቀ፡ መላእክት፡ ቅዱሰ፡ <add resp="PRS8999Strelcyn">ሚካኤል፡</add> ብዙኃ፡ ወርቀ፡ 
+                                ወብሩረ፡ በእንተ፡ መድኃኒተ፡ ነፍሱ፡</incipit>
+                        </msItem>
+                        <msItem xml:id="ms_i2">
+                            <locus from="33r" to="122v"/>
+                            <title ref="LIT1295Dersan"/>
+                            <note>Homilies for festivals of St Michael on the 12th of each month accompanied by miracles of this Archangel. In the 
+                                catalogue description nothing is said about commemorations or salām prayers, that usually accompany the Dǝrsāna Mikāʾel
+                                for each month.</note>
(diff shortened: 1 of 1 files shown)


LondonBritishLibrary/orient/BLorient4849.xml
```diff
@@ -0,0 +1,608 @@
...
+                            <layoutDesc>
+                                <layout writtenLines="21"/>                                
+                            </layoutDesc>   
+                        </objectDesc> 
+                        <handDesc>
+                            <handNote script="Ethiopic" xml:id="h1">
+                                <desc>Written in a good hand.</desc>
+                                <date notBefore="1600" notAfter="1799" resp="PRS8999Strelcyn"/>
+                            </handNote>
+                            <handNote xml:id="h2" script="Ethiopic">
```
eu-genia, line 369: I think it is a composite MS with two MSS of Dersana Mikael, one partial/incomplete, one full, or possibly this particular homily for Maggabit is not in the main part so was added as msItem 1, see https://github.com/BetaMasaheft/Documentation/issues/2563

LondonBritishLibrary/orient/BLorient4849.xml
```diff
@@ -0,0 +1,608 @@
...
+                                </msItem>
+                            </msItem>
+                            <msItem xml:id="ms_i2.6">
+                                <locus from="66r" to="72v"/>
+                                <title ref="LIT1295Dersan#Maggabit"/>
+                                <msItem xml:id="ms_i2.6.1">
+                                    <locus from="66r" to="71v"/>
+                                    <title ref="LIT1295Dersan#HomilyMaggabit"/>
+                                    <note>The name of Severus is replaced by 'John' in the manuscript.</note>
+                                    <incipit xml:lang="gez">ድርሳን፡ ዘደረሰ፡ ዮሐንስ፡ ሊቀ፡ ጳጳሳት፡ ዘአንጾኪያ፡ ዘይትነበብ፡ በወርኃ፡ መጋቢት፡ በእንተ፡ ቅዱስ፡ ሚካኤል፡ ወበእንተ፡ አፍቅሮተ፡
```
eu-genia, line 145: to check if the same as item 1

eu-genia (changes requested):

LondonBritishLibrary/orient/BLorient4849.xml
CarstenHoffmannMarburg: I see, that is better to define ms_i1 as an individual text part, even though there is no indication in the Strelcyns catalogue of any change of hands or any changes in layout and dimensions. However, I am still wondering what to do with the twenty pages filled with miniatures in the beginning? They can be defined as part of part 1 (= ms_i2) or part 2 (= ms_i2 to ms_i5), they can form a text part on their own or they can be seen as later addition, that should be included into the physDesc of the record concerning the manuscript as a whole. I think, I would opt for creating a third text part for the miniatures in the decoDesc.

CarstenHoffmannMarburg: I have just mentioned that ms_i1 ends at folio 32r, while ms_i2 begins at folio 32v.  I think it is rather unlikely that these are separate codicological units. In that case, either folio 32r or folio 32v must have been added later, after the two parts had been bound together. Whatever the reason for putting ms_i1 first, I think that each ms_item was written one after the other.

LondonBritishLibrary/orient/BLorient4849.xml
```diff
@@ -0,0 +1,609 @@
...
+                            <note>Extract from <ref type="work" corresp="LIT1295Dersan"/>. The name of the saint is missing, the spaces reserved 
+                                for the rubrics being left blank.</note>
+                            <incipit xml:lang="gez">ወእምድኅረ፡ ተጠምቁ፡ ነበሩ፡ ወርኃ፡ ፍጹመ፡ ኀበ፡ ጳጳስ፡ እንዘ፡ ይሜህሮሙ፡ ነገረ፡ ሃይማኖት፡ ርትዕት፡ ወማቴዎስ፡ ነግድ፡ እምብዝኃ፡
+                                ፍሥሐሁ፡ ወሀበ፡ ኵሎ፡ ዘረከበ፡ ለቤተ፡ ክርስቲያን፡ ለሊቀ፡ መላእክት፡ ቅዱሰ፡ <add resp="PRS8999Strelcyn">ሚካኤል፡</add> ብዙኃ፡ ወርቀ፡ 
+                                ወብሩረ፡ በእንተ፡ መድኃኒተ፡ ነፍሱ፡</incipit>
+                        </msItem>
+                        <msItem xml:id="ms_i2">
+                            <locus from="33r" to="122v"/>
+                            <title ref="LIT1295Dersan"/>
+                            <note>Homilies for festivals of St Michael on the 12th of each month accompanied by miracles of this Archangel. In the
```
DenisNosnitsin1970, line 324: This information better belongs to `<title ref="LIT1295Dersan"></>`

DenisNosnitsin1970 (changes requested):
The description is ok, but frankly speaking, I do not see why Eugenia saw the need to represent the manuscript as composite. Strelcyn does not report any change in physical dimensions, neither there is another hand in play (except remark 2, not clear). Deviations from common structure of the work are possible. As to the miniatures, in many richly illuminated manuscripts miniatures are painted in a special quire, this was the way of introduction of pictures into a volume, within one project (contrary to the inserting them into the text, which represent another method). To sum up, in my ipinion not enough evidence is reported in Strelcyn's description (and the ms is not accessible), to make the description that complex.

CarstenHoffmannMarburg:
> The description is ok, but frankly speaking, I do not see why Eugenia saw the need to represent the manuscript as composite. Strelcyn does not report any change in physical dimensions, neither there is another hand in play (except remark 2, not clear). Deviations from common structure of the work are possible. As to the miniatures, in many richly illuminated manuscripts miniatures are painted in a special quire, this was the way of introduction of pictures into a volume, within one project (contrary to the inserting them into the text, which represent another method). To sum up, in my ipinion not enough evidence is reported in Strelcyn's description (and the ms is not accessible), to make the description that complex.

I feel well with both. An argument for the definition of a composite manuscript is the attestation of different owners for p1_i1 and for p3. Alas, it is also not clear from the catalogue, whether the occurrence of PRS14483Silondis and PRS14484LahyaNegest is limited to p2. One could check this, when images are available as they are supposed to be in fall.

CarstenHoffmannMarburg:
> > The description is ok, but frankly speaking, I do not see why Eugenia saw the need to represent the manuscript as composite. Strelcyn does not report any change in physical dimensions, neither there is another hand in play (except remark 2, not clear). Deviations from common structure of the work are possible. As to the miniatures, in many richly illuminated manuscripts miniatures are painted in a special quire, this was the way of introduction of pictures into a volume, within one project (contrary to the inserting them into the text, which represent another method). To sum up, in my ipinion not enough evidence is reported in Strelcyn's description (and the ms is not accessible), to make the description that complex.
> 
> I feel well with both. An argument for the definition of a composite manuscript is the attestation of different owners for p1_i1 and for p3. Alas, it is also not clear from the catalogue, whether the occurrence of PRS14483Silondis and PRS14484LahyaNegest is limited to p2. One could check this, when images are available as they are supposed to be in fall.

Probably it is best, to return to the description as one codicological unit and mark the incertanties somewhere in the physDesc (probably in collation) or in history.

DenisNosnitsin1970:
This is an argument which should be supported by smth else, related to material structure/palaeographic evidence. Besides, Strelcyn does not say that the name is added in a different hand (all three persons may well belong to one family).

DenisNosnitsin1970:
I do not think you need to redo it again. A remark that it may be less complex (under "History"?) will be enough.

CarstenHoffmannMarburg:
> I do not think you need to redo it again. A remark that it may be less complex (under "History"?) will be enough.

I alraedy did. It is not a big deal to recover the older version of this record. We just need a decision, what is the more appropriate representation. The rework in XML should not be the problem. I will push and you will see...

eu-genia (approved):

PR #2493, opened 2024-07-03
BLorient12860_CH_june24

CarstenHoffmannMarburg:
The description of contents was according to the description in Strelcyns catalogue and our discussion in https://github.com/BetaMasaheft/Documentation/issues/2559.

### DIFF
--- a/LondonBritishLibrary/orient/BLorient12860.xml
+++ b/LondonBritishLibrary/orient/BLorient12860.xml
@@ -75,7 +75,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                 </support>
                                 <extent>
                                     <measure unit="leaf">31+1</measure>
-                                    <dimensions type="outer" unit="mm" xml:id="leafdim">
+                                    <dimensions type="leaf" unit="mm" xml:id="leafdim">
                                         <height>210-215</height>
                                         <width>265-270</width>
                                     </dimensions>
--- a/LondonBritishLibrary/orient/BLorient12860.xml
+++ b/LondonBritishLibrary/orient/BLorient12860.xml
@@ -31,41 +31,41 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     <msContents>
                         <msItem xml:id="ms_i1">
                             <locus from="1ra" to="26vb"/> 
-                            <title ref="LIT1512GadlaY">The life of Yāred</title>
+                            <title ref="LIT1512GadlaY"/>
                             <note>With many variations to <bibl><ptr target="bm:ContiRossini1904YaredPantalewonT"/></bibl>.</note>
                             <msItem xml:id="ms_i1.1">
                                 <locus from="1ra"/>
-                                <title ref="LIT000000000000000">The life of Yāred</title>
+                                <title ref="LIT1303Dersan"/>
                                 <incipit xml:lang="gez">በስመ፡ <gap reason="ellipsis"/> ድርሳን፡ ዘቅዱስ፡ ያሬድ፡ ቀርነ፡ ቤተ፡ ክርስቲያን፡ ዘይጼውዖሙ፡ ለሃይማናን፡ ከመ፡
                                     ይሴብሕዎ፡ ለእግዚአብሔር፡ በክላህ፡ ወበዐቢይ፡ ቃል፡ ከመ፡ ሱራፌል፤ ድርሳን፡ ዘያሬድ፡ ካህን፡ </incipit>    
                             </msItem>
                             <msItem xml:id="ms_i1.2">
                                 <locus from="18vb"/>
-                                <title>The three miracles belonging to the Life of Yāred</title>
+                                <title ref="LIT7019MiraclesYared"/>
                                 <msItem xml:id="ms_i1.2.1">
                                     <locus from="18vb"/>
-                                    <title ref="LIT000000000000000000000000" type="incomplete">Miracle 1</title>
+                                    <title ref="LIT7019MiraclesYared#Satan" type="incomplete"/>
                                     <note>The beginning is missing</note>
                                     <incipit xml:lang="gez">ወእምድኅረ፡ አዕረፈ፡ <persName ref="PRS10245Yared">ያሬድ፡</persName> ካህን፡ ፈጺሞ፡ 
                                         ገድሉ፡ ወመዊዖ፡ ሰይጣነ፡</incipit>
                                 </msItem>
                                 <msItem xml:id="ms_i1.2.2">
                                     <locus from="22rb"/>
-                                    <title ref="LIT0000000000000000000000000">Miracle 2</title>
-                                    <incipit xml:lang="gez">በስመ፡ <gap reason="ellipsis"/> ነገር፡ በእንተ፡ ሰዋሬሁ፡ ለያሬድ፡ ካህን፡ ሠርጐ፡ ቤተ፡ ክርስቲያን፡ </incipit>
+                                    <title ref="LIT7019MiraclesYared#Decoration"/>
+                                                   <incipit xml:lang="gez">በስመ፡ <gap reason="ellipsis"/> ነገር፡ በእንተ፡ ሰዋሬሁ፡ ለያሬድ፡ ካህን፡ ሠርጐ፡ ቤተ፡ ክርስቲያን፡ </incipit>
                                 </msItem>
                                 <msItem xml:id="ms_i1.2.3">
                                     <locus from="24vb"/>
-                                    <title ref="LIT0000000000000000000000000">Miracle 3</title>
+                                    <title ref="LIT7019MiraclesYared#Gift"/>
                                     <incipit xml:lang="gez">በስመ፡ <gap reason="ellipsis"/> ነገር፡ በእንተ፡ ሰዋሬሁ፡ ለያሬድ፡ ካህን፡ ዘተወህቦ፡ ጸጋ፡ </incipit>
                                 </msItem>
                             </msItem>
-                        </msItem>
-                        <msItem xml:id="ms_i2">
-                            <locus from="27ra" to="31vb"/>
-                            <title ref="LIT2989RepCh270">Malkǝʾ of Yāred</title>
-                            <incipit xml:lang="gez">በስመ፡ እግዚአብሔር፡ ሥሉስ፡ ዘበሰማያት፡ አቡነ።</incipit>
-                        </msItem>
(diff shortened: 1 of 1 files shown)


LondonBritishLibrary/orient/BLorient12860.xml
```diff
@@ -0,0 +1,207 @@
...
+                    </msIdentifier>
+                    <msContents>
+                        <msItem xml:id="ms_i1">
+                            <locus from="1ra" to="26vb"/> 
+                            <title ref="LIT1512GadlaY"/>
+                            <note>With many variations to <bibl><ptr target="bm:ContiRossini1904YaredPantalewonT"/></bibl>.</note>
+                            <msItem xml:id="ms_i1.1">
+                                <locus from="1ra"/>
+                                <title ref="LIT1303Dersan"/>
+                                <incipit xml:lang="gez">በስመ፡ <gap reason="ellipsis"/> ድርሳን፡ ዘቅዱስ፡ ያሬድ፡ ቀርነ፡ ቤተ፡ ክርስቲያን፡ ዘይጼውዖሙ፡ ለሃይማናን፡ ከመ፡
```
karljonaskarlsson, line 39: I suppose that መሃይምናን፡ must be meant, so maybe it would be good to have a "sic" here. The question is if the "error" is found in the manuscript or only in Strelcyn's catalogue (I checked and you have copied his text faithfully)

LondonBritishLibrary/orient/BLorient12860.xml
```diff
@@ -0,0 +1,207 @@
...
+                                </item>
+                                <item xml:id="e2">
+                                    <locus target="#16ra"/>
+                                    <desc type="OwnershipNote">The place left for the name of the owner has not been filled in (<locus target="#16ra"/> 
+                                        and other places).</desc>
+                                </item>
+                                <item xml:id="e3">
+                                    <locus target="#17r"/>
+                                    <desc type="OwnershipNote">In the lower margin a signature in Cyrillic 
+                                        (<persName ref="PRS14477Pechaguev" role="owner">Т. Печагуев</persName>) is repeated four times.</desc>
```
karljonaskarlsson, line 122: Should the use of a foreign script be marked with "foreign"?

LondonBritishLibrary/orient/BLorient12860.xml
```diff
@@ -0,0 +1,207 @@
...
+                                </item>
+                                <item xml:id="e6">
+                                    <locus target="#Ir"/>
+                                    <desc type="StampExlibris">A rectangular stamp.</desc>
+                                    <q xml:lang="fr">Archives de la Maison Impériale / Angelo Comneno de Thessalie</q>
+                                </item>
+                                <item xml:id="e7">
+                                    <locus target="#Ir"/>
+                                    <desc type="AcquisitionNote">Note in ink.</desc>
+                                    <q xml:lang="fr">Evangile en parchemin originel de XVIIèm siècle en langue gheez donné à moi par un prêtre en
```
karljonaskarlsson, line 143: There are some orthographic peculiarities in the French that Strelcyn marked with "sic". I would take over the "sic"s that he has given.

LondonBritishLibrary/orient/BLorient12860.xml
```diff
@@ -0,0 +1,207 @@
...
+                        </additions>
+                        <bindingDesc>
+                            <binding xml:id="binding">
+                                <decoNote xml:id="b1">European binding</decoNote>
+                            </binding>
+                        </bindingDesc>
+                    </physDesc>
+                    <history>
+                        <origin>
+                            <origDate notBefore="1600" notAfter="1700">17th century</origDate>
```
karljonaskarlsson, line 156: 1699?

karljonaskarlsson (changes requested):

LondonBritishLibrary/orient/BLorient12860.xml
CarstenHoffmannMarburg: I added a `<sic resp="CH">`, even though I could not check it by myself, because images are still not available.

LondonBritishLibrary/orient/BLorient12860.xml
```diff
@@ -0,0 +1,210 @@
...
+                <msDesc xml:id="ms">
+                    <msIdentifier>
+                        <repository ref="INS00001BL"/>
+                        <collection>Oriental</collection>
+                        <idno>BL Oriental 12860</idno>
+                        <altIdentifier><idno>Strelcyn 55</idno></altIdentifier>
+                    </msIdentifier>
+                    <msContents>
+                        <msItem xml:id="ms_i1">
+                            <locus from="1ra" to="26vb"/>
```
eu-genia, line 33: to 31vb, malk is part of the hagiographic dossier

eu-genia (changes requested):

karljonaskarlsson (approved):

LondonBritishLibrary/orient/BLorient12860.xml
```diff
@@ -0,0 +1,210 @@
...
+                    <physDesc>
+                        <objectDesc form="Codex">
+                            <supportDesc>
+                                <support>
+                                    <material key="parchment"/>
+                                </support>
+                                <extent>
+                                    <measure unit="leaf">31+1</measure>
+                                    <dimensions type="outer" unit="mm">
+                                        <height>210-215</height>
```
DenisNosnitsin1970, line 79: I wonder if you can use here `<dimensions type="leaf" unit="mm" xml:id="leafdim"/>` (see Guidelines, Object description). It is a bit strange that there is no clear reference in the catalogue as to what the outer dimensions actually are (but the differences in size recorded are unlikely meaningful).

DenisNosnitsin1970 (changes requested):

LondonBritishLibrary/orient/BLorient12860.xml
CarstenHoffmannMarburg: Do you think, that this is probably not a composite manuscript?

DenisNosnitsin1970: I have doubts about this. No change of hand is noticed, and no significant change of written area/ layout. The differences of size are small and be results of various things (inlcuding trimming in the library). Of course, he noted this probably because he thought it is worth of, but in view of the lack of any clarification I would not make it too complicated.

CarstenHoffmannMarburg: OK. So, I can leave it, like this - just to insert `xml:id="leafdim"`. However, I think, all dimension for all manuscripts of the BL, described by Strelcyn in this catalogue, refer to the dimensions of the leaves. I suspect the same for the manuscript of the BML of Florence. Do you think, it is neccessary to rework all these manuscripts or to insert this string just for this and for every manuscript to come?

DenisNosnitsin1970: This is an open question. I assume that he otherwise used to provide outer dimensions proper (if the text block is fully homogeneous; and this is what is needed first), but tacitly switched to leaf size if the text block shows some breaks. Not a very big deal since the outer dimensions are unlikely very much different, but it generates some uncertainty. I wonder if he explains it anywhere - probably not. Did he assume that this is absolutely clear? You can check this parameter, to understand his way of thinking. :)

CarstenHoffmannMarburg: He does not mention it anywhere. I assume, that all dimensions refer to the size of the leaves, because there are certain manuscripts, for which he give different dimensions for some parts of the codex (e.g. no. 60, BLorient11719). I do not think, that he would have tacitly switched from book dimensions to dimensions of leaves.

DenisNosnitsin1970: Not providing outer dimensions would be quite strange, but older catalogues contain a number of odd things. It would be interesting to check it empirically.

eu-genia (approved):

LondonBritishLibrary/orient/BLorient12860.xml
karljonaskarlsson: It might be relevant to note that the ITIESE project took measurements for two of the BL manuscripts catalogued by Strelcyn (BL Or. 2198 and BL Or. 13156). See https://journals.sub.uni-hamburg.de/aethiopica/article/view/2106/2036. Our measurements compare as follows with those of Strelcyn:

BL Or. 2198:   90 x 102 (Strelcyn)    vs.   110 × 93 × 16 (ITIESE)
BL Or. 13156:   92 x 115 (Strelcyn)    vs.   c. 113 × c. 98 × 63 (ITIESE)

We measured the outer dimensions of the boards. For BL Or. 13156, we also measured the folio dimensions as: 112 x 93. It is not entirely clear, but this might suggest that Strelcyn measured leaves.

DenisNosnitsin1970: It is surprising and I do not know any other cataloguer who would do this (but perhaps I did not pay attention). Another thing - now I see that he gives first the width and then the height, without an explicit remark (this can be easily changed, though).

DenisNosnitsin1970 (changes requested):

LondonBritishLibrary/orient/BLorient12860.xml
CarstenHoffmannMarburg: I remember getting the same impression from Marrassini's catalogue of the Florentine manuscripts. I suppose that researchers who work primarily as philologists are often more interested in the dimensions of the leaves because they are trying to identify codicological units rather than the dimensions of the boards. In Strelcyns' catalogue (1978) the lower number always comes first. I have never come across a codex whose width is greater than its height. For this reason I assume that Strelcyn always gives the width before the height.

DenisNosnitsin1970: Yes, some cataloguers provided indeed "width x height", esp. in the past (this is the case of Conti Rossini's catalogue of d'Abbadie collection), more typical today is of course vice versa, but also the first variant would be ok provided there is a clear hint about this in the guidelines (if there are any). The idea of "codicological units" was less clear in the time of P. Marrassini than today, even though vigilant cataloguers would always indicate meaningful differences. I am not sure that he provided the dimensions of leafs, it looks unlikely to me, and outer dimensions are needed by the librarians for administering the books, leaf sizes would be quite unpractical. Again the lack of clear indications in cataloguing guidelines may create a confusion.

CarstenHoffmannMarburg:
@DenisNosnitsin1970: When all requests are met, can this record be approved and merged?

CarstenHoffmannMarburg:
@DenisNosnitsin1970, @eu-genia: Can this pull request be merged?

PR #2496, opened 2024-07-08
Updated BLorient2198 and BLorient13156

CarstenHoffmannMarburg:
According to information in bm:KarlssonEtAl2023 as recommended by Jonas in https://github.com/BetaMasaheft/Manuscripts/pull/2493.

### DIFF
--- a/LondonBritishLibrary/orient/BLorient13156.xml
+++ b/LondonBritishLibrary/orient/BLorient13156.xml
@@ -122,7 +122,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                         ones, having more visible ruling impressions forming a grid of parallel vertical and horizontal lines used by the illustrator for
                                         outlining the miniatures (e.g. the horizontal line of the lower hem of the tunic is drawn onto one of the horizontal ruling lines,
                                         the vertical line of the nose dorsum follows the vertical ruling). Compass marks for drawing the halos placed at the intersection
-                                        of ruling lines are visible on some of the folios (e.g. fol. 46r).</note>
+                                        of ruling lines are visible on some of the folios (e.g. <locus target="#46r">46r</locus>). 
+                                        Cf. <bibl><ptr target="bm:KarlssonEtAl2023Handlist"/><citedRange unit="page">208</citedRange></bibl>.</note>
                                 </layout>
                             </layoutDesc>
                         </objectDesc>
@@ -369,7 +370,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         <citedRange unit="page">251-253</citedRange>
                     </bibl>
                     <bibl>
-                        <ptr target="bm:Mercier2000LArcheEthiopienne"/>
+                        <ptr target="bm:MercierArt2001"/>
                         <citedRange unit="page">79</citedRange>
                     </bibl>
                 </listBibl>
--- a/LondonBritishLibrary/orient/BLorient13156.xml
+++ b/LondonBritishLibrary/orient/BLorient13156.xml
@@ -103,14 +103,27 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                     <measure unit="leaf">161</measure>
                                     <measure unit="page" type="blank">4</measure><locus target="#46r #112r #127r #144r"/>
                                     <dimensions type="outer" unit="mm">
+                                        <height>113</height>
+                                        <width>98</width>
+                                        <depth>63</depth>
+                                    </dimensions>
+                                    <dimensions xml:id="leafdim" type="outer" unit="mm">
                                         <height>115</height>
                                         <width>92</width>
-                                        <depth/>
                                     </dimensions>
                                 </extent>
                             </supportDesc>
                             <layoutDesc>
-                                <layout columns="1" writtenLines="12 15"/>
+                                <layout columns="1" writtenLines="12 15">   
+                                </layout>
+                                <layout>
+                                    <locus target="#4v #46v #62v #95v #112v #127v #131v #144v"/>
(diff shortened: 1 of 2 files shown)


LondonBritishLibrary/orient/BLorient13156.xml
```diff
@@ -103,14 +103,27 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
                                     </dimensions>
                                 </extent>
                             </supportDesc>
                             <layoutDesc>
-                                <layout columns="1" writtenLines="12 15"/>
+                                <layout columns="1" writtenLines="12 15">   
+                                </layout>
+                                <layout>
+                                    <locus target="#4v #46v #62v #95v #112v #127v #131v #144v"/>
+                                    <note corresp="#d1 #d2 #d3 #d4 #d5 #d6 #d7 #d8">The illustrated folios have a different ruling pattern from the written
```
karljonaskarlsson, line 121: In order not to plagiarize, I think we need to mention the source when quoting (almost) verbatim. What is the Bm practice on this?

LondonBritishLibrary/orient/BLorient13156.xml
```diff
@@ -171,6 +185,49 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                 <locus target="#23v"/>
                                 <desc>Symbolic signs referring to Christ in the margins of <locus target="#23v"/> and on various pages of the codex.</desc>
                             </decoNote>
+                            <decoNote type="band" xml:id="d10">
+                                <locus target="#5r"/>
+                                <desc>Most bands are topped by stepped elements. They vary in appearance, but the use of colours similar to those
```
karljonaskarlsson, line 190: Again, I feel that the source really needs to be mentioned.

Is there a point in repeating the same general description for each of the headpieces? Or could the information be placed somewhere else?

LondonBritishLibrary/orient/BLorient13156.xml
```diff
@@ -233,12 +290,17 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     </physDesc>
                     <history>
                         <origin>
-                            <origDate notBefore="1400" notAfter="1500">15th century</origDate>
+                            <origDate notBefore="1380" notAfter="1450">1380-1450</origDate>
```
karljonaskarlsson, line 293: This is a good way of presenting the information, I think!

LondonBritishLibrary/orient/BLorient2198.xml
```diff
@@ -234,7 +237,27 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         </handDesc>
                         <decoDesc>
                             <decoNote xml:id="d1" type="band">
-                                <desc>Horizontal bands and lines in red and black ink separate the various texts.</desc>
+                                <locus target="#14r"/>
```
karljonaskarlsson, line 240: Again, I would suggest citing the source in connection to the quotations.

karljonaskarlsson (changes requested):

LondonBritishLibrary/orient/BLorient13156.xml
```diff
@@ -290,12 +356,24 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
         <body>
-            <div type="bibliography"/><!---->
+            <div type="bibliography">
+                <listBibl type="secondary">
+                    <bibl>
+                        <ptr target="bm:Uhlig1988Palaographie"/>
+                        <citedRange unit="page">251-253</citedRange>
+                    </bibl>
+                    <bibl>
+                        <ptr target="bm:Mercier2000LArcheEthiopienne"/>
```
eu-genia, line 372: note that Mercier already exists in Zotero as bm:MercierArt2001; you created a doublet, please remove the doublet and update the tag.

eu-genia (changes requested):

LondonBritishLibrary/orient/BLorient13156.xml
CarstenHoffmannMarburg: Dorothea recommended to note each decorative element individually, if possible. This is useful for those, who search for decorations by the number of decorations in one manuscript. This note refers to each of the the decorations, so I repeated this phrase various times.

LondonBritishLibrary/orient/BLorient13156.xml
CarstenHoffmannMarburg: I think we have a legion of verbatim quotes from catalogues, Wikipedia or other websites in our records, with the source cited at the end in <additional> or in <bibliography>, or (in case of wikipedia) not cited at all.

eu-genia (approved):

eu-genia (approved):

LondonBritishLibrary/orient/BLorient13156.xml
karljonaskarlsson: If I understand you correctly, I don't think this is a good practice. The reader must be able to tie specific information to specific sources, no?

 @eu-genia @DenisNosnitsin1970

LondonBritishLibrary/orient/BLorient13156.xml
karljonaskarlsson: Ok, it may not look nice, but it makes sense that it can be useful for some. Thank you! :)

LondonBritishLibrary/orient/BLorient13156.xml
eu-genia: When a source is quoted then the bibliography should be provided at the quotation. Catalogues which are the basis for the description are different, it is assumed that they are the main source of information, so it is even more important to specify the bits that come from elsewhere.

karljonaskarlsson (approved):

PR #2501, opened 2024-07-11
Create UppEt36.xml

karljonaskarlsson:
Created an empty file to be able to refer to it in the description of O Etiop. 3 – will come back and fill it will content in due time


eu-genia:
before merging please make sure that the file validates

karljonaskarlsson:
Sorry, could you explain what I need to do to make sure that it validates?

eu-genia:
As far as I can see, you should fill `msItem `with something, at least an empty `<title></title>` tag

karljonaskarlsson:
Thank you! Now I understand!

PR #2502, opened 2024-07-11
Create UppEt5.xml

karljonaskarlsson:
Created an empty file to be able to refer to it in the description of O Etiop. 3 – will come back and fill it will content in due time


eu-genia (approved):
you could/should better create all stubs in one go /one branch and PR (for the app - the fewer PRs the better, as long as there are no more than 10 files per merge)

PR #2506, opened 2024-07-12
2024 07 04 j kuub o etiop003

karljonaskarlsson:
Created ms record based on Löfgren's and Zetterstéen's catalogues and the online catalogue in Alvin, with transcriptions based on the digital copy available online.

### DIFF
--- a/Uppsala/UppEt3.xml
+++ b/Uppsala/UppEt3.xml
@@ -62,7 +62,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                             </explicit>
                             <explicit type="supplication" xml:lang="gez">ኦአቡየ፡ ወእኁየ፡ <del rend="erasure">
                                 <hi rend="rubric"><unclear>ያ<gap reason="lost" unit="chars" quantity="1"/>ቅበ<gap reason="lost" unit="chars" quantity="1"/><supplied reason="lost">፡</supplied>
-                                    አብ፡</unclear></hi> ኢትርስዐኒ፡ በጸሎ<del rend="erasure">ት</del>ከ፡ ሊተ፡ ለ<del rend="erasure">ይስሐቅ፡</del><gap reason="lost" unit="chars" quantity="12"/>
+                                    አብ፡</unclear></hi> ኢትርስዐኒ፡ በጸሎ<del rend="erasure">ት</del>ከ፡ ሊተ፡ 
+                                ለ<del rend="erasure"><persName ref="PRSXXXXXXXXXXXXXXXXXXXXXXXXXXXXX">ይስሐቅ፡</persName></del><gap reason="lost" unit="chars" quantity="12"/>
                                 አሜን<hi rend="partialRubric">፨</hi>
                             </del>
                             </explicit>
@@ -385,11 +386,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                 </item>
                                 
                                 <item xml:id="e1">
-                                    <locus target=""/>
-                                    <desc type="">White paper sticker on the inside of the front board. It is not mentioned by 
+                                    <desc>White paper sticker on the inside of the front board. It is not mentioned by 
                                         <bibl><ptr target="bm:Zettersteen1899Upsala"/></bibl> or
                                             <bibl><ptr target="bm:Loefgren1974Katalog"/></bibl>.</desc>
-                                 <q xml:lang="sv">Etiopiska. Katal. n<superscript>o</superscript> III.</q>
+                                 <q xml:lang="sv">Etiopiska. Katal. n<sup>o</sup> III.</q>
                                 </item>
                                 
                                 <item xml:id="e2">
--- a/Uppsala/UppEt3.xml
+++ b/Uppsala/UppEt3.xml
@@ -286,8 +286,22 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                             
                         </msItem>
                         
-                        <msItem xml:id="ms_i1">
-                            <title type="complete" ref="LIT2317Senodo"/>
+                        <msItem xml:id="ms_i3">
+                            <locus from="77r" to="79v"/>
+                            <title type="complete" ref=""/>
+                            <note><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">30</citedRange></bibl>
+                                calls it a passion prayer (<foreign xml:lang="de">Passionsgebet</foreign>) and notes that it is partially rhymed.
+                                </note>
+                            <incipit xml:lang="gez">
+                                ኦእግዚኦ፡ ነአኵተከ፡ ወንሴብሐከ፡ ወንሰግድ፡ ለዕበየ፡ ሐይልከ፡ ወለጽንዓ፡ ፍቅርከ፡ እስመ፡ ፈኖከ፡ ለነ፡ ወልድከ፡ 
+                                <sic resp="PRS6337Lofgren">ፍቁርከ፡</sic> ወኢምሕኮ፡ እምኔነ፡
+                            </incipit>
+                            <explicit xml:lang="gez">
+                                ኢነጸረቶ፡ እሙ፡ በጊዜ፡ ሞቱ፡ ወኢኦድዎ፡ አዕርክቲሁ፡ ለሐውጾቱ። ጻመወ፡ በፈቃዱ፡ ባሕቲቱ፡ እስመ፡ መሐሪ፡ 
+                                ውእቱ፡ ወመፍቀሬ፡ ሰብእ፡ ይብትክ፡ ወይፍታሕ፡ መአስረ፡ ኀጣውኢነ፡ ኀበ፡ ሀሎ፡ ነሀሉ፡ በከመ፡ ለሊሁ፡ አሰፈወነ፡ 
+                                ሎቱ፡ ስብሐት፡ <add place="above">ዘ</add>እምኀቤነ፡ ወለነኒ፡ መንግሥቶ፡ ይክፍለነ፡ ይእዜኒ፡ ወዘልፈኒ፡ 
+                                ወለዓለመ፡ ዓለም፡ አሜን።
+                            </explicit>
                         </msItem>
                     </msContents>
                     
--- a/Uppsala/UppEt3.xml
+++ b/Uppsala/UppEt3.xml
@@ -62,7 +62,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                             <locus from="57r" to="76v"/>
                             <title type="complete" ref="LIT1195Bahrah">Maṣḥafa ḥassāb</title>
                             <note><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">29</citedRange></bibl> notes
-                            that among the <foreign xml:lang="gez">ḥassāb</foreign> texts known from manuscripts, the one in 
+                                that among the <foreign xml:lang="gez-tr">ḥassāb</foreign> texts known from manuscripts, the one in 
                                 <ref type="mss" corresp="BNFet64"/>
                                 (and possibly also the one in <ref type="mss" corresp="BerOrQuart370"/>) is closest to this. The
                             transcriptions below are based on those provided by 
@@ -81,8 +81,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                 <locus from="57r" to="57v"/>
                                 <title type="complete">Calculation of 13 monthly cycles (<foreign xml:lang="gez">አቅማር፡</foreign>)</title>
                                 <note><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">29</citedRange></bibl> notes
-                                    that one <foreign xml:lang="gez">qamar</foreign> equals 532 years. Thirteen 
-                                    <foreign xml:lang="gez">qamar</foreign> thus equal 6916 years, to which (according to the text)
+                                    that one <foreign xml:lang="gez-tr">qamar</foreign> equals 532 years. Thirteen 
+                                    <foreign xml:lang="gez-tr">qamar</foreign> thus equal 6916 years, to which (according to the text)
                                     one adds 84 "Years of the Evangelists"
                                     (<foreign xml:lang="gez">ወትዌስክ፡ ዲቤሁ፡ ዓመተ፡ ወንጌ<sic resp="JK">ለ</sic>ዊያን፡ 
                                         <hi rend="rubric">፹፬፡</hi></foreign>),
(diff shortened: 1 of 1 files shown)


Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,552 @@
...
+                    Revelation of John, Maṣḥafa Ḥassāb
+                </title>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/">
+                        <p>
```
eu-genia, line 17: remove `p` in licence

Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,552 @@
...
+                            Attribution-ShareAlike 4.0. </p>
+                    </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <msDesc xml:id="ms">
+                    <msIdentifier>
+                        <repository ref="INS0356Upp"/>
+                        <collection>O Etiop.</collection>
+                        <collection/>
```
eu-genia, line 28: remove empty `collection`

Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,552 @@
...
+            
+        </fileDesc>
+        <encodingDesc>
+            <xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="https://raw.githubusercontent.com/BetaMasaheft/Documentation/master/prefixDef.xml">
+                <xi:fallback>
+                    <p>Definitions of prefixes used.</p>
+                </xi:fallback>
+            </xi:include>
+        </encodingDesc>
+        <profileDesc>
```
eu-genia, line 528: add `textClass `with `keywords`
at least
ChristianLiterature
Bible
Chronography

eu-genia (changes requested):

eu-genia (approved):

Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,557 @@
...
+                                    and suggests that the form with <foreign xml:lang="gez">ጥ</foreign>
+                                    instead of <foreign xml:lang="gez">ፕ</foreign> 
+                                    could be a folk etymology, influenced by <foreign xml:lang="gez">ጥንት፡</foreign>, "beginning".</note>
+                            </msItem>
+                            <msItem xml:id="ms_i2.14">
+                                <locus from="65r" to="65v"/>
+                                <title type="complete">About the "Day of John"</title>
+                                <note>According to <bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">30</citedRange></bibl>,
+                                    the text offers a complicated method (<foreign xml:lang="gez">verwickelte Methode</foreign>).
+                                    .</note>
```
CarstenHoffmannMarburg, line 229: Delete one period (.)

Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,557 @@
...
+                                    and <bibl><ptr target="bm:Dillmann1865Lexicon"/><citedRange unit="column">758</citedRange></bibl>.</note>
+                            </msItem>
+                            <msItem xml:id="ms_i2.16">
+                                <locus target="#66r"/>
+                                <title type="complete">About finding the <foreign xml:lang="gez">መጥቅዕ፡</foreign> 
+                                    (<foreign xml:lang="la">dies novilunii</foreign>)</title>
+                                <note>Regarding the term <foreign xml:lang="gez">መጥቅዕ፡</foreign>,
+                                    <bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">30</citedRange></bibl>
+                                    refers to <bibl><ptr target="bm:Chaine1925Chronologie"/><citedRange unit="page">106</citedRange></bibl>
+                                    and <bibl><ptr target="bm:Dillmann1865Lexicon"/><citedRange unit="column">1124 f.</citedRange></bibl>.
```
CarstenHoffmannMarburg, line 246: It is better to have 1124-1125 to avoid f. (= following).

CarstenHoffmannMarburg (approved):

Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,557 @@
...
+                            <msItem xml:id="ms_i2.3">
+                                <locus from="60r" to="60v"/>
+                                <title type="complete">Chronology of Enoch</title>
+                                <note><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">29</citedRange></bibl> notes
+                                    that it consists of seven <foreign xml:lang="gez-tr">sanābət</foreign> of 700 years each,
+                                    resulting in a sum of 7000 years.</note>
+                            </msItem>
+                            <msItem xml:id="ms_i2.4">
+                                <locus from="60v" to="61r"/>
+                                <title type="complete">Chronology of the Sibyl (<foreign xml:lang="gez">ሰቤላ፡</foreign>)</title>
```
DenisNosnitsin1970, line 123: The ID of "Chronology of the Sybil" is LIT4992HassabSabela. If all mss which Löfgren indicates indeed contain the same work, the ID will hopefully be used in the descriptions. I would find them and see if a kind of harmonisation is needed or not (If BNFet64 contains the same text, the ID may probably be the same; but in BNFet64 it is LIT3851Hassab, here above is the general record LIT1195Bahrah).  I would also strongly recomment to quote a little bit of text for each of the part texts, this will be very helpful (It would be nice if you could quote from the incipit under LIT4992HassabSabela)

Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,557 @@
...
+                        </decoDesc>
+                        
+                        <additions>
+                            <list>
+                                <item xml:id="a1">
+                                    <locus target="55r"/>
+                                    <desc type="Unclear">Under the supplication to <ref target="#ms_i1"/>, between the
+                                        <foreign xml:lang="la">cruces ansatae</foreign>, the remnants of an erased
+                                        note are visible. Only one word has been left untouched. The note is not mentioned by 
+                                        <bibl><ptr target="bm:Zettersteen1899Upsala"/></bibl> or
```
DenisNosnitsin1970, line 373: I do not think this info is necessary (that someone does not mention smth).

Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,557 @@
...
+                                        <bibl><ptr target="bm:Loefgren1974Katalog"/></bibl>.</desc>
+                                    <q xml:lang="gez">እግዚአብሔር፡<del rend="erasure" unit="chars" quantity="15"/></q>
+                                </item>
+                                
+                                <item xml:id="a2">
+                                    <locus target="55v"/>
+                                    <desc type="Supplication">Supplication by a later hand. The note is mentioned by 
+                                        <bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">28</citedRange></bibl> and
+                                        <bibl><ptr target="bm:Zettersteen1899Upsala"/><citedRange unit="page">510</citedRange></bibl>.</desc>
```
DenisNosnitsin1970, line 382: You do not need to refer to both catalogues in this case (I would drop it in a few other cases, whenever it does not bring anything new in comparison to Löfgren), even though the older should be in the bibliography.

Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,557 @@
...
+                                        <bibl><ptr target="bm:Loefgren1974Katalog"/></bibl>.</desc>
+                                </item>
+
+                            </list>
+                        </additions>
+                        
+                        <bindingDesc>
+                            <binding xml:id="binding">
+                                <decoNote xml:id="b1" type="Boards">
+                                    As noted by <bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">31</citedRange></bibl>,
```
DenisNosnitsin1970, line 456: Since good images are available, it is better to describe the binding according to Bm scheme (which would make the part  more clear and informative). To refer to the leather cover and the fact that it is tooled, you do not need to quote the catalogues.

Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,557 @@
...
+                                        <bibl><ptr target="bm:alvin"/></bibl>.
+                                    </note>
+                                </layout>
+                            </layoutDesc>
+                        </objectDesc>
+                        
+                        <handDesc>
+                            <handNote xml:id="h1" script="Ethiopic">
+                                <desc><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">31</citedRange>
+                                </bibl> calls it a "large and even archaic script"
```
DenisNosnitsin1970, line 351: Please note that this characteristic is in part misleading today (15th century handwriting would not be thought as "archaic", this would refer to pre-mid-14th century hands). The characteristic might be: "Regular, careful; by a good scribe. The letters are of big size" (the reference without numerical value is also a bit shaky).

DenisNosnitsin1970 (changes requested):

Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,557 @@
...
+                    </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <msDesc xml:id="ms">
+                    <msIdentifier>
+                        <repository ref="INS0356Upp"/>
+                        <collection>O Etiop.</collection>
+                        <idno>O Etiop. 3</idno>
+                        <altIdentifier><idno>Löfgren 3</idno></altIdentifier>
```
DenisNosnitsin1970, line 28: I assume the link to imges should be included.

DenisNosnitsin1970 (changes requested):

Uppsala/UppEt3.xml
karljonaskarlsson: I included it at the very end of the file, from line 549 and forward. Should they be placed here?

Uppsala/UppEt3.xml
```diff
@@ -0,0 +1,557 @@
...
+                                        <bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">28</citedRange></bibl> and
+                                        <bibl><ptr target="bm:Zettersteen1899Upsala"/><citedRange unit="page">510</citedRange></bibl>.</desc>
+                                    <q xml:lang="gez">ኢትርስኡኒ፡ ለኀጥእ፡ እሁክሙ፡ 
+                                        <persName ref="PRS14496BahaylaSellase">በኀይለ፡ ሥላሴ፡</persName> እለ፡ 
+                                        አንበብክምዋ፡ ለዛቲ፡ መጽሐፍ፡ <sic corresp="JK">አ</sic>አበዊነ፡ ወኣኀዊነ<sic corresp="JK">፡</sic></q>
+                                </item>
+                                
+                                <item xml:id="e1">
+                                    <desc>White paper sticker on the inside of the front board. It is not mentioned by 
+                                        <bibl><ptr target="bm:Zettersteen1899Upsala"/></bibl> or
```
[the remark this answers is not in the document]
karljonaskarlsson: Thank you!

Uppsala/UppEt3.xml
karljonaskarlsson: But if the default is that the descriptions is based on the catalogues mentioned at the end, then there needs to be some way of marking what we add, no?

Uppsala/UppEt3.xml
karljonaskarlsson: Thank you! I can only see one manuscript with LIT4992HassabSabela, and there I only have the incipit in Turaev, which is not exactly the same as in the Uppsala manuscript. Can we be sure that they are the same text?

I looked at the text in BNFet64, and it has a lot of similarities with the text in the Uppsala manuscript, but the elements don't come in the same order and I'm not sure that it has exactly the same contents. I don't know if it's better to say that they are both LIT3851Hassab or to use the more general ID. What do you say?

I can add incipits!

Uppsala/UppEt3.xml
DenisNosnitsin1970: I think so; this is the place to refer to the images online (below you refer to the description in the database), the link may be `http://urn.kb.se/resolve?urn=urn:nbn:se:alvin:portal:record-177135` (Zhenia might know better which exactly you pick).

eu-genia: `Facsimile` graphic url is the correct place for external links ( permalinks)

`Facs` is only for iiif manifests when these have been previously enabled.

Uppsala/UppEt3.xml
DenisNosnitsin1970: Yes, this is a complicated issue, but you will soon find it is impossible to come through if you want to be very exact. We say: we base our descriptions on smth, and this implicates that we as cataloguers take the old foundation and bring smth new/improve older descriptions, as our routine work. If there is smth very important, you can make a note in the edition history.

Uppsala/UppEt3.xml
DenisNosnitsin1970: Of course we cannot be sure (in my opinion, the chance is high), but we can signal their relationship (maybe Löfgren had a chance to check the text closer!). You can also use `<cert=""/>`

DenisNosnitsin1970: To do this interlinking/integrating work with "text units" is the constant background research task done in Bm.  If you are sure that both texts are related but have differences, you can still create another record and interlink both, explaining also your reservations (quoting meaningful chunks of text is always helpful, the lack of quotation is a weakness of old catalogues, like of Turaev).

karljonaskarlsson: Ok, I add LIT4992HassabSabela. It seems reasonable at least until someone takes a closer look at the manuscripts.

Uppsala/UppEt3.xml
karljonaskarlsson: Ok, yes, I will certainly learn with time :) Thank you for the input on this!

CarstenHoffmannMarburg (approved):

eu-genia (approved):

Uppsala/UppEt3.xml
karljonaskarlsson: I don't have any experience of using the Bm scheme for binding descriptions, so if you have time, I would appreciate your feedback, especially if there is anything essential that I missed. Although we have good images, I still don't understand exactly how the leather patches on the spine connect to the rest.

DenisNosnitsin1970: Ok, we can do this.

DenisNosnitsin1970 (changes requested):

CarstenHoffmannMarburg (approved):

karljonaskarlsson:
@DenisNosnitsin1970 , I took away the "negative" descriptions of what is not mentioned in the printed catalogues. Let me know if there is anything else that need to be done before it can be merged. Thank you!

CarstenHoffmannMarburg (approved):

karljonaskarlsson:
@eu-genia , this has been open for a long time, although I believe all requested changes were carried out. Would it be possible for you to merge it, or does that disturb the workflow?

PR #2509, opened 2024-07-17
2024 07 12 jk uppsala o etiop 7

karljonaskarlsson:


Uppsala/UppEt7.xml
```diff
@@ -0,0 +1,336 @@
...
+                            <msItem xml:id="ms_i1.3">
+                                <locus from="18va" to="22ra"/>
+                                <title type="complete">Part 3: Questions 35–43 with answers</title>
+                                <note><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">106</citedRange></bibl>
+                                    notes that the colophon (on <locus target="#22ra"/>) is also found in <ref type="mss" corresp="BerOrQuart344"/>
+                                    and translates it as follows: "Full is Part 3 in the abbreviation by <foreign xml:lang="gez-tr">Yərunāmes</foreign>
+                                    (<persName ref="PRS5340Hieronym">Jerome</persName>), who attributed to <persName ref="PRS7706Palladiu">Palladius</persName>
+                                    the 'Questions' and two chapters, for he wrote in the time of 
+                                    <persName ref="PRS9479Theodosi">Theodosius the Great</persName>, but 
+                                    <persName ref="PRS7706Palladiu">Palladius</persName> wrote his son
```
CarstenHoffmannMarburg, line 122: **while** Palladius wrote **in the time of** Arcadius.

Uppsala/UppEt7.xml
```diff
@@ -0,0 +1,336 @@
...
+            <xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="https://raw.githubusercontent.com/BetaMasaheft/Documentation/master/prefixDef.xml">
+                <xi:fallback>
+                    <p>Definitions of prefixes used.</p>
+                </xi:fallback>
+            </xi:include>
+        </encodingDesc>
+        <profileDesc>
+            <langUsage><language ident="en">English</language></langUsage>
+            <langUsage><language ident="gez">Gəʿəz</language></langUsage>
+            <langUsage><language ident="gez-tr">Gəʿəz</language></langUsage>
```
CarstenHoffmannMarburg, line 318: Gəʿəz transliterated

Uppsala/UppEt7.xml
```diff
@@ -0,0 +1,336 @@
...
+            <langUsage><language ident="de">German</language></langUsage>
+            <langUsage><language ident="sv">Swedish</language></langUsage>
+        </profileDesc>
+        <revisionDesc>
+            <change who="JK" when="2024-07-12">Created record</change>
+            <change who="JK" when="2024-07-16">Added physical description</change>
+            <change who="JK" when="2024-07-16">Added bibliography and place name</change>
```
CarstenHoffmannMarburg, line 325: I think it is enough to have one change record for all the changes within short time.

CarstenHoffmannMarburg (changes requested):

CarstenHoffmannMarburg (approved):

Uppsala/UppEt7.xml
```diff
@@ -0,0 +1,336 @@
...
+                                <date notBefore="1600" notAfter="1699">17th century.</date>
+                            </handNote>
+                        </handDesc>
+                        
+                        <additions>
+                            <list>
+                                <item xml:id="e1">
+                                    <desc type="StampExlibris">Paper sticker on the outside of the front board. It is not mentioned by 
+                                        <bibl><ptr target="bm:Zettersteen1899Upsala"/></bibl> or
+                                        <bibl><ptr target="bm:Loefgren1974Katalog"/></bibl>.</desc>
```
DenisNosnitsin1970, line 215: Here and below: "It is not mentioned..." - this info is not necessary (most of the catalogues do not mention internal marking of the libraries, what kind of stickers existed in Zettersteen's time?). The remarks of this type make the description heavier and more difficult to look though.

Uppsala/UppEt7.xml
```diff
@@ -0,0 +1,336 @@
...
+                                    ተፈጸመ፡ ክፍል፡ ቀዳሚ፡ በረድኤተ፡ እግዚአብሔር፡ ወነገረ፡ ከመ፡ ተኃጥአ፡ እምቀዳሚሁ፡ 
+                                    <hi rend="partialRubric">፵</hi>ተስእሎታት<hi rend="partialRubric">፨</hi>
+                                </explicit>
+                            </msItem>
+                            
+                            <msItem xml:id="ms_i1.2">
+                                <locus from="13ra" to="18va"/>
+                                <title type="complete">Part 2: Questions 17–34 with answers</title>
+                                <note><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">105</citedRange></bibl>
+                                    quotes the colophon (on <locus target="#18va"/>) and translates it as follows: "End of the abbreviation of Part 2"
```
DenisNosnitsin1970, line 93: If it is a colophon, it should be encoded as a colophon (here and in other cases).

Uppsala/UppEt7.xml
```diff
@@ -0,0 +1,336 @@
...
+                                        Bibliotek af <persName ref="PRS2680Bergman">August Bergman</persName>, svenskt missionär
+                                        i <placeName ref="LOC1741Balaza">Belleza</placeName>, Asmara, i Abyssinien</q>
+                                </item>
+                            </list>
+                        </additions>
+                        
+                        <bindingDesc>
+                            <binding contemporary="true" xml:id="binding">
+                                <decoNote xml:id="b1" type="Boards">Wooden boards.
+                                </decoNote>
```
DenisNosnitsin1970, line 247: That is a good and clear description that follows the guidelines. You can add for clarity: "Two pairs of sewing stations".  Perhaps you could count the number of quires to "Collation" (I think they are easy to count).

DenisNosnitsin1970 (changes requested):

Uppsala/UppEt7.xml
karljonaskarlsson: Isn't the question "what kind of stickers existed in Zettersteen's time?" exactly a reason why it may be good to encode this kind of things?

To make the description lighter, would it be possible to include a general note somewhere stating that items "xml:id so-and-so" are not mentioned in this-and-that source?

Uppsala/UppEt7.xml
karljonaskarlsson: Thank you! I will add that.

Uppsala/UppEt7.xml
karljonaskarlsson: I checked another manuscript with the same text (DD-048), and it also has the same colophons, so I will encode them as embedded colophons

karljonaskarlsson: After rereading the guidelines, I think it's possible to argue that this is not a colophon. I will leave them as explicits

CarstenHoffmannMarburg (approved):

Uppsala/UppEt7.xml
DenisNosnitsin1970: In fact, you cannot know if the sticker was not there or Zettersteen's catalogue just does not record this detail (rather the second). If you inlcude a kind of this rule, the descriptions will be full of "negative statements", that this catalogue does not provide this and that. I have three catalogues for Abbadie collection, no one gives exact folia numbers (just to start this; of course no info on signatures/stamps). How the manuscript looked in the time of the older catalogue is the issue which we do not really approach, if we have very good digital duplicates accessible.

CarstenHoffmannMarburg (approved):

karljonaskarlsson:
@DenisNosnitsin1970 , can this be merged or are there any changes that I should still carry out?

karljonaskarlsson:
@eu-genia , perhaps this could also be merged, as I believe all the issues have been solved a while ago

PR #2510, opened 2024-07-17
Update IVorlov7.xml

DenisNosnitsin1970:
Corrects

### DIFF
--- a/StPetersburg/IVorlov7.xml
+++ b/StPetersburg/IVorlov7.xml
@@ -56,13 +56,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                         <foliation/>
                            <collation/>
 
-                      </supportDesc>
-
-                     <layoutDesc>
-                        <layout columns="2" writtenLines="13">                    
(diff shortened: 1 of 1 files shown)


StPetersburg/IVorlov7.xml
```diff
@@ -375,20 +370,18 @@ schematypens="http://relaxng.org/ns/structure/1.0"
                            </supportDesc>
 
                            <layoutDesc>
-                              <layout>
-                                 <p>
+                              <layout>
```
eu-genia, line 373: nothing to say about the layout? columne/lines? which is the layout on these pages? if nothing is known then delete the element, the pages are already specified for the part

eu-genia (changes requested):

StPetersburg/IVorlov7.xml
DenisNosnitsin1970: I revised it a bit, it is difficult to be more precise because the catalogue description of Turaev is very vague and confusing.

eu-genia (approved):

PR #2515, opened 2024-07-25
Create BLorient5085

CarstenHoffmannMarburg:
Some leaves have been displaced. This seems to have led to some confusion on the side of St. Strelcyn, who wrote several chapters as well as the index on folio 2ra and following folioss. This must be a mistake, because folios 2 to 11 cannot carry both the index and chapters 10 to 14. However, I have decided to mark the uncertainty and present a possible solution. It can be checked and corrected when images of the BL are available, hopefully in September.

### DIFF
--- a/LondonBritishLibrary/orient/BLorient5085.xml
+++ b/LondonBritishLibrary/orient/BLorient5085.xml
@@ -1,0 +1,399 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="BLorient5085" xml:lang="en" type="mss">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>Universal Chronology of ʾAbušākǝr; Explanation of the book of Ezekiel</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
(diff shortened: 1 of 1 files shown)


LondonBritishLibrary/orient/BLorient5085.xml
```diff
@@ -0,0 +1,399 @@
...
+                            </msItem>
+                            <msItem xml:id="ms_i1.6">
+                                <locus from="12ra" to="13va"/>
+                                <title type="incomplete">Index of the chapters</title>
+                                <note>The beginning (nos. 1-17) of the Index of chapters is missing. According to 
+                                    <bibl><ptr target="bm:Strelcyn1978BritishLibrary"/><citedRange unit="page">111</citedRange></bibl> the 
+                                    remaining parts of the index (nos. 18-59) occupy <locus from="2ra" to="13va"/>, but this is contradicted by Strelcyn's
+                                    information on the distribution of the chapters, which places ch. 10 on <locus from="2ra"/>, ch. 11 on 
+                                    <locus from="2ra"/>, ch. 12 on <locus from="3va"/>, ch. 13 on <locus from="6rb"/> and ch. 14 on 
+                                    <locus from="8vb" to="11vb"/>. For this reason, it can be assumed, that the index actually runs
```
karljonaskarlsson, line 74: I don't know if we need to assume anything – maybe it's enough to simply state the facts, as you have done up to this sentence?

karljonaskarlsson (approved):
I'm inspired by your encoding! :)

LondonBritishLibrary/orient/BLorient5085.xml
CarstenHoffmannMarburg: Yes, it is better not to be vague. I think that 2va->12va is just a small change, but changing the loci of five subparts is just strange. That is why I favoured this solution. However, if this is too complicated and not convincing on first reading, it is better to delete this suppositional sentence.

karljonaskarlsson (approved):

LondonBritishLibrary/orient/BLorient5085.xml
karljonaskarlsson: I looked at Strelcyn's description and I was confused. I could not figure out what he was intending to convey

CarstenHoffmannMarburg: He gave an overview of the chapters following the numbers of chapters, but we need to follow the manuscript and number the msItems according to their appearance in the manuscript.

PR #2519, opened 2024-07-31
Abb 146 186 219

DenisNosnitsin1970:
Three new descriptions and minor corrections in some others. Description  Abb. 186 is quite complex, I 
decided to analytically treat each part as individual unit since it seems to have been just a bunch of leaves
(the divisions made in Chane's and Conti Rossini's catalogues appeared to me dubious, and again, they did not accord with each other at some points).

### DIFF
--- a/ParisBNF/abb/BNFabb146.xml
+++ b/ParisBNF/abb/BNFabb146.xml
@@ -70,7 +70,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         M. Chaîne dates the manuscript to 18th/19th century. The catalogues of M. Chaîne and C. Conti Rossini
                         provide a number of details which are difficult to reconcile. The manuscript is composite, 
                         but the exact number of units can 
-                        hardly be established, on the basis of the available descritions. 
+                        hardly be established, on the basis of the available descriptions. 
                      </origin>
                      <provenance>The manuscript was bought by
                         <persName ref="PRS1071dAbbadi" role="owner">Antoine d'Abbadie</persName>
@@ -213,7 +213,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                            
                            <msItem xml:id="p2_i2.1">
                               <locus from="112r"/>                             
-                             <title type="complete" ref="LIT2888RepCh169">St Mercury</title>                        
+                              <title type="complete" ref="LIT2888RepCh169">Malkǝʾ-hymn for St Mercury</title>                        
                         <textLang mainLang="gez"/>
                         <incipit xml:lang="gez">ሰላም፡ ለዝክረ፡ ስምከ፡ ገብረ፡ ኢየሱስ፡ ብሂል፡</incipit>
                      </msItem>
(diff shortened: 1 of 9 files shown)


ParisBNF/abb/BNFabb146.xml
```diff
@@ -0,0 +1,344 @@
...
+                     
+                  </physDesc>
+                  
+                  <history>
+                     <origin>
+                        <origDate notBefore="1700" notAfter="1799">18th century</origDate>.
+                        M. Chaîne dates the manuscript to 18th/19th century. The catalogues of M. Chaîne and C. Conti Rossini
+                        provide a number of details which are difficult to reconcile. The manuscript is composite, 
+                        but the exact number of units and scribal hands involved can 
+                        hardly be established, on the basis of the available descritions.
```
CarstenHoffmannMarburg, line 73: These comments are better to be placed in handDesc, contents or collation.

ParisBNF/abb/BNFabb186.xml
```diff
@@ -0,0 +1,733 @@
...
+                     <idno/>
+                  </msIdentifier>
+                  
+                  <msContents>
+                     <summary/>
+                     
+                     <msItem xml:id="p6_i1">
+                        <locus from="14r" to="37v"/>
+                        <title type="complete">Armenian printed booklet concerning calendar and computus</title>
+                        <textLang mainLang="ar" >Armenian</textLang>
```
CarstenHoffmannMarburg, line 515: Remove space in `<textLang mainLang="ar" >`  

I think p6_i1 is better to be defined as accMat.

CarstenHoffmannMarburg (changes requested):

ParisBNF/abb/BNFabb186.xml
DenisNosnitsin1970: Yes, I though of this option for a while. A certain problem is that through the thorough foliation the print was incorporated into the structure of the manuscript (maybe already by d'Abbadie himself). If it would not have been the case (and if this print would have been loosely placed at th end, for instance), accMat would be perfect. It is not the case. I would leave it as it is, for now; but we can ask Zhenia what she thinks.

ParisBNF/abb/BNFabb146.xml
DenisNosnitsin1970: I removed the statement on the scribal hands, units is ok, I assume.

ParisBNF/abb/BNFabb186.xml
CarstenHoffmannMarburg: I would opt to define it as accompanying material nevertheless, but both solutions might be possible.

ParisBNF/abb/BNFabb146.xml
CarstenHoffmannMarburg: Change to "description". I have seen this just now.

CarstenHoffmannMarburg (changes requested):

DenisNosnitsin1970:
Thanks.

CarstenHoffmannMarburg (commented):
You did not remove the space in BNFabb186 (line 515) yet.

DenisNosnitsin1970:
Indeed, sorry.

CarstenHoffmannMarburg (approved):

PR #2547, opened 2024-08-22
Choice fix

eu-genia:
FYI: TEI guidelines specify that the `choice `element must have at least two children (`orig`/`corr `or `sic`/`corr `or similar)


eu-genia:
without this fix you will keep getting validation errors (https://github.com/BetaMasaheft/Manuscripts/actions/runs/10507793132/job/29110317362) on every push to app

DenisNosnitsin1970 (approved):

CarstenHoffmannMarburg:
OK. What happened to the 138 records that are in the link? Are they fixed everyone or do they still need rework?

eu-genia:
all fixed

PR #2556, opened 2024-08-27
Tweed Scrolls Project

Ralph-Lee-UK:
Correct formatting for facsimile files for EMIPms00470-00490


Ralph-Lee-UK:
@eu-genia this was intended to be the final request today - but the validation has failed - I'm not sure what to do to correct this. Can you point me in the right direction?

eu-genia (approved):

EMIP/magicscrolls/EMIPms00470/EMIPms00470.xml
```diff
@@ -89,6 +89,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <change when="2022-09-02" who="CS">added include statements to incorporate transkribus output</change>
          <change when="2022-09-23" who="RL">Corrected include code errors</change>
          <change when="2022-09-26" who="RL">Removed text body code lines</change></revisionDesc>
+         <change when="2024-08-27" who="RL">Corrected formatting of text file</change>
```
eu-genia, line 92: sorry this element must go inside of revisionDesc

eu-genia (changes requested):

EMIP/magicscrolls/EMIPms00470/EMIPms00470.xml
eu-genia: It should be
```
<revisionDesc>
  <change>...</change>
  <change>...</change>
  <change>...</change>
</revisionDesc>
```

eu-genia:
You could also add a condition to your XSL for scrolls that instead of `div type="textpart" subtype="folio"` it writes `div type="textpart" subtype="image"` or any other sensible subtype (I did it now for EMIPms00470 - then in the visualization you have image 1 - image 2 etc as the printed divisions)

Alternatively one could also run 
```
<xsl:stylesheet version="2.0"
  xmlns:xsl="http://www.w3.org/1999/XSL/Transform" 
  xpath-default-namespace="http://www.tei-c.org/ns/1.0" >
  <xsl:output method="xml" omit-xml-declaration="yes" version="1.0" encoding="utf-8" indent="yes"/>
 

  <xsl:template match="div[@type='textpart']">
    <xsl:apply-templates select="node()"/>
  </xsl:template>
  
  <xsl:template match="* | text() | comment() | processing-instruction() | @*">
    <xsl:copy>
      <xsl:apply-templates select="node() | @*"/>
    </xsl:copy>
  </xsl:template>
  
</xsl:stylesheet>
```
to remove `div textpart`

EMIP/magicscrolls/EMIPms00470/EMIPms00470.xml
Ralph-Lee-UK: Now changed - sorry for a simple error here!

Ralph-Lee-UK:
That change makes good sense. I'll see to this tomorrow. That should be a simple change

PR #2567, opened 2024-08-29
Tweed scrolls project august 29 2024

Ralph-Lee-UK:
Reformatted text and facsimile files for scolls


eu-genia (approved):

eu-genia:
just in case: i have recently been using XML refactoring in Oxygen, it is very helpful to batch transform files; you just need a descriptor xml for each xsl and you can apply xslt to as many files at once as needed

Ralph-Lee-UK:
Thank you!

PR #2575, opened 2024-09-02
Create BLorient11763.xml

CarstenHoffmannMarburg:
I created a record for BLorient11763 according to the catalogue description in Strelcyn 1978. Images are still not available as far as I know.

I created a new LIT ID for ms_i2 according to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2591.

I refrained from creating an ID for ms_i3 according to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2592.

### DIFF
--- a/LondonBritishLibrary/orient/BLorient11763.xml
+++ b/LondonBritishLibrary/orient/BLorient11763.xml
@@ -40,7 +40,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                             <textLang mainLang="gez"/>
                         </msItem>
                         <msItem xml:id="ms_i3">
-                            <title>Prayer against ʿaynat</title>
+                            <title ref="LIT00000000000000000000000">Prayer against ʿaynat šumdān</title>
                             <incipit xml:lang="gez">በስመ፡ <gap reason="ellipsis"/> ጸሎት፡ በእንተ፡ ሕማመ፡ ዓይነት፡ ሺምዳን፡ ተክለ፡ ሽርታን፡</incipit>
                             <textLang mainLang="gez"/>
                         </msItem>
--- a/LondonBritishLibrary/orient/BLorient11763.xml
+++ b/LondonBritishLibrary/orient/BLorient11763.xml
@@ -1,0 +1,125 @@
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI xmlns="http://www.tei-c.org/ns/1.0" type="mss" xml:lang="en" xml:id="BLorient11763">
+    <teiHeader xml:base="https://betamasaheft.eu/">
+        <fileDesc>
+            <titleStmt>
+                <title>Magical Scroll</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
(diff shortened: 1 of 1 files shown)


nafisa-valieva (commented):
Hi Karsten, all looks well, besides  the title
<title>Prayer against ʿaynat</title>
                            <incipit xml:lang="gez">በስመ፡ <gap reason="ellipsis"/> ጸሎት፡ በእንተ፡ ሕማመ፡ ዓይነት፡ cተክለ፡ ሽርታን፡</incipit>
I think it is a prayer against  <title>Prayer against ʿaynat</title> it is Prayer against ሺምዳን፡. I asked Solomon and he agrees.

nafisa-valieva (approved):

LondonBritishLibrary/orient/BLorient11763.xml
```diff
@@ -0,0 +1,125 @@
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI xmlns="http://www.tei-c.org/ns/1.0" type="mss" xml:lang="en" xml:id="BLorient11763">
+    <teiHeader xml:base="https://betamasaheft.eu/">
+        <fileDesc>
+            <titleStmt>
+                <title>Magical Scroll</title>
```
eu-genia, line 7: I have a vague feeling that it is better to say "magic scroll", not "magical", as it contains magic, and not works magic (for the same reason for me "magical prayer" sounds better than "magic prayer"). I have seen both used though. @AaronButts as the only native speaker around may have a better idea. Does it make a difference at all? Should we aspire to any sort of standard?

LondonBritishLibrary/orient/BLorient11763.xml
```diff
@@ -0,0 +1,125 @@
...
+        </fileDesc>
+        <encodingDesc>
+            <xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="https://raw.githubusercontent.com/BetaMasaheft/Documentation/master/prefixDef.xml">
+                <xi:fallback>
+                    <p>Definitions of prefixes used.</p>
+                </xi:fallback>
+            </xi:include>
+        </encodingDesc>
+        <profileDesc>
+            <langUsage><language ident="en">English</language><language ident="gez">Gǝʿǝz</language></langUsage>
```
eu-genia, line 114: the `keywords `are missing completely, when you edit in a new branch please add `Magic`

LondonBritishLibrary/orient/BLorient11763.xml
AaronButts: For me, "magic scroll" is much much better (I don't use "magical scroll"). As Genia says, it is a scroll that contains magic. 

I would also personally use "magic prayer", following similar logic: It is a prayer consisting of magic. 

I realize that "magical" is etymologically an adjective for "magic", but it has quite different connotations in usage: supernatural, as well as remarkable, exceptional, incredible. Basically, for me "magical" has some value judgement, whether about the efficacy of the magic, or more generally that something is extraordinary. I personally try to avoid using "magical" as much as possible when describing the practice of magic.

CarstenHoffmannMarburg:
I will be reviewing the BL and BML manuscripts that I have recently pushed or that are in progress.

LondonBritishLibrary/orient/BLorient11763.xml
```diff
@@ -0,0 +1,125 @@
...
+                                <decoNote xml:id="b1" type="SlipCase"><material key="leather"/>Contained in a cylindrical leather case.</decoNote>
+                            </binding>
+                        </bindingDesc>
+                    </physDesc>
+                    <history>
+                        <origin>
+                            <origDate notBefore="1600" notAfter="1699">17th century</origDate>
+                        </origin>
+                        <provenance>Two owners are named: <persName role="owner" ref="PRS14519WaldaSamuel">Walda Sāmuʾel</persName>
+                            and <persName role="owner" ref="PRS14520BakweraSeyon">Bakʷǝra Ṣǝyon</persName></provenance>
```
karljonaskarlsson, line 85: Bakʷra without schwa, no?

CarstenHoffmannMarburg: OK, I will also change it in PRS14520BakweraSeyon.

PR #2602, opened 2024-09-09
Create BAVet124.xml

karljonaskarlsson:
Created stub based on https://github.com/BetaMasaheft/Dillmann/issues/496

### DIFF
--- a/VaticanBAV/et/BAVet124.xml
+++ b/VaticanBAV/et/BAVet124.xml
@@ -26,8 +26,13 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         <repository ref="INS0003BAV"/>
                         <idno>Aeth. 124</idno>
                     </msIdentifier>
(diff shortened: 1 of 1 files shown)


eu-genia (approved):
it is OK if you need to refer to the ID urgently, just keep in mind that stubs for mss in catalogues in the pipeline creates a challenge for knowing which records in the folder are truly there and which are just placeholders, so wherever anyhow possible, avoid creating stubs (except for collections/catalogues we are currently not working with, as then we can assume that the records are all stubs)

MagdaKrzyz:
In that case how shall I refer to the mss that appear in the dictionary?

eu-genia:
It is OK to create stubs if this is necessary for referring to them; we just then must keep track of that somehow and fill in the descriptions as soon as possible.

CarstenHoffmannMarburg (approved):

CarstenHoffmannMarburg (approved):

PR #2620, opened 2024-09-18
Dr 23 06 corrections

thea-m:
Various corrections and commenting out of facsimile as discussed in https://github.com/BetaMasaheft/Documentation/issues/2388 
PR validation fails - could you help me? Thank you!
(I now also commented out the facsimile files in the expanded records EMML4398, ESqdq004, EMIP01859, all leading to error of missing root element. The text files of the expanded records also since always have schema errors)


eu-genia:
You can comment out the included facsimile in the main record but not in the facsimile.xml, this file remains as is.
If you comment the xiinclude out, it will not be expanded anyway.
you cannot comment a root element out, it will clearly rend any xml file invalid, xml files must have root elements

eu-genia (changes requested):

thea-m:
Great, thank you!

eu-genia (approved):
thank you
I just reinstated the facsimile.xml to the previous state to have fewer changes in the merge and in the revision history

PR #2623, opened 2024-09-24
2024 09 20 uub o etiop 21

karljonaskarlsson:
Created ms record based on Löfgren's catalogue, the online catalogue (http://urn.kb.se/resolve?urn=urn:nbn:se:alvin:portal:record-159895), and the images available online

### DIFF
--- a/Uppsala/UppEt21.xml
+++ b/Uppsala/UppEt21.xml
@@ -582,8 +582,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                     <measure unit="page" type="blank">8</measure>
                                     <measure unit="quire">14</measure>
                                     <dimensions type="outer" unit="mm">
-                                        <height>220</height>
+                                        <height>225</height>
                                         <width>180</width>
+                                        <depth>67</depth>
                                     </dimensions>
                                 </extent>
                                 
@@ -725,9 +726,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         <origin>
                             <origDate notBefore="1700" notAfter="1799">18th century</origDate>
                             According to the colophon on <locus target="#141va"/>, the original owner was
-                            <persName ref="PRS14568WaldaKiros" role="owner"/> and 
+                            <persName ref="PRS14568WaldaKiros" role="patron"/> and 
                             the scribe was <persName ref="PRS14569GadlaGiyorgis" role="scribe"/>.
-                            The following family members of <persName ref="PRS14568WaldaKiros" role="owner"/> are mentioned:
+                            The following family members of <persName ref="PRS14568WaldaKiros" role="patron"/> are mentioned:
                             his father <persName ref="PRS14570Hezqeyas"/>, 
                             his mother <persName ref="PRS14571Ebrawit"/>, and
                             his brother <persName ref="PRS14572Awkatewos"/>. The name of one further brother or sister has been erased.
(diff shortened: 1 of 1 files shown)


eu-genia (approved):
very nice, thank you

Uppsala/UppEt21.xml
nafisa-valieva, line 1: Thank you! Looks like a good record to follow!

karljonaskarlsson: If all manuscripts were as well-organized as this one, our lives would be very simple :)

Uppsala/UppEt21.xml
```diff
@@ -0,0 +1,815 @@
...
+                            <list>
+                                <item xml:id="e1">
+                                    <desc type="StampExlibris">Paper sticker on the inside of the front board.</desc>
+                                    <q xml:lang="sv">
+                                        Uppsala Universitets<lb/>
+                                        Bibliotek<lb/>
+                                        Handskr. avd.<lb/>
+                                        O Etiop. 21<lb/>
+                                    </q>
+                                    <note>Below the rest of the text, a note in pencil: "Löfgren nr 38"</note>
```
CarstenHoffmannMarburg, line 710: To my opinion, this note deserves its own extra.

Uppsala/UppEt21.xml
```diff
@@ -0,0 +1,815 @@
...
+                            his mother <persName ref="PRS14571Ebrawit"/>, and
+                            his brother <persName ref="PRS14572Awkatewos"/>. The name of one further brother or sister has been erased.
+                        </origin>
+                        <provenance>
+                            At an unknown point in time, <persName ref="PRS9043Svensson" role="owner">Axel B. Svensson</persName> 
+                            aquired
+                            the manuscript. He subsequently donated the manuscript to <persName ref="PRS14565ErlingEidem" role="owner">
+                                <title>Archbishop</title>
+                                Erling Eidem</persName> on <date when="1948-09-15">15 September 1948</date>, according to
+                            a type-written note that—at least when the manuscript was catalogued by <persName ref="PRS6337Lofgren"/>—was
```
CarstenHoffmannMarburg, line 742: I think, there must be spaces around `-` cf. 
'(...) according to a type-written note that — at least when the manuscript was catalogued by <persName ref="PRS6337Lofgren"/> — was (...)'

CarstenHoffmannMarburg (changes requested):

Uppsala/UppEt21.xml
nafisa-valieva: well seen! I agree

Uppsala/UppEt21.xml
karljonaskarlsson: I think it's common English style to use em dashes here (whereas German and Swedish would have an en dash with spaces, as you suggest). Which style guide does Bm follow?

Uppsala/UppEt21.xml
karljonaskarlsson: Ok! I'll change it

Uppsala/UppEt21.xml
CarstenHoffmannMarburg: I usually check english comments with [Deepl](https://www.deepl.com/en/write#en/en-gb/according%20to%20a%20type-written%20note%20that%E2%80%94at%20least%20when%20the%20manuscript%20was%20catalogued%20by%20%3CpersName%20ref%3D%22PRS6337Lofgren%22%5C%2F%3E%E2%80%94was). The deepl-recommendation in this case is to use either commas or dashes with spaces. However, I think, it is not that important to note it in the guidelines. You may select the option of your choice.

CarstenHoffmannMarburg (approved):

Uppsala/UppEt21.xml
eu-genia: In Bm, as in COMSt, in Aethiopica, and in all other HLCEES publications, we should follow, if we can and care, Oxford Style guide. There should be NO spaces around the m-dashes or around the n-dashes in our publications. We have not been 100% careful in Bm as it is a work in progress, but clearly wherever you submit a catalogue or an article for print you must adhere to Oxford Style. 
https://www.aai.uni-hamburg.de/en/comst/publications/bulletin/authors.html
https://journals.sub.uni-hamburg.de/aethiopica/authorguidelines

PR #2632, opened 2024-10-11
Update EMML50.xml <locus> 

anaiswion:


EMML/1-1000/EMML50.xml
```diff
@@ -231,7 +231,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://ra
                      </msItem>
 
                     <msItem xml:id="ms_i2.19">
-                       <locus from="150"></locus>
+                       <locus target="#150"></locus>
```
eu-genia, line 234: target is to one specific page only, so it must be 150r if you are sure, or if it is 150r to 150v then you need from-to

eu-genia (changes requested):

eu-genia (approved):
thank you!

PR #2635, opened 2024-10-15
Update BAVet142.xml

eu-genia:
@MarcinKrawczuk I can see this branch but I think it has not been merged - are you still working on it? Thank you! Can you also check others?
@MassimoVilla has kindly agreed to try and do some more work on GreTisVat so I am trying to not double the work


MarcinKrawczuk (approved):
I have it in my notes that ms_i5 is the same as ESdmd013 f. 50vb but I have no idea why I thought so or whether it is true

PR #2653, opened 2024-10-16
B lorient555

MarcinKrawczuk:
It seems that I wanted to add refs to the parts but not all were identifiable to me on the basis of the catalogue.


eu-genia:
Here again the version of the master branch at https://github.com/BetaMasaheft/Manuscripts/blob/master/LondonBritishLibrary/orient/BLorient555.xml seems the same if not more complete, so closing this and deleting the branch

PR #2654, opened 2024-10-16
Bl orient763 mk

MarcinKrawczuk:


eu-genia:
it looks to me that there is no data here that is not already in the version on the master branch

eu-genia:
This is a good example why we need to merge ) It looks to me as if @thea-m did the same encoding without noticing that there had been work done. I will try to compare the editions...

eu-genia:
I have compared and updated on master, closing here and deleting the branch

PR #2656, opened 2024-10-18
B lorient578

MarcinKrawczuk:


MarcinKrawczuk:
I really do not know what I intended here, maybe let's remove the branch and leave it as it is

eu-genia:
From the file history it is clear that you were working on your branch, and at the same time @thea-m was not aware of it and was working on hers, trying to complete BL descriptions, so that you produced parallelly two versions of the same record. Hence some variation in the encoding: e. g. what you had as ms_i1.1 was presented by @thea-m as a1, LIT6434Hymn is now listed as [LIT2981RepCh262](https://betamasaheft.eu/LIT2981RepCh262) (corrected in the current version)
It is a bit difficult to compare, since both files are very long. I suggest that we keep the current version, if you notice something you would like to correct please feel free.
For the future - another illustration of how important it is to merge branches to avoid double work in a shared environment. We should have be more attentive before, especially when sharing work on the same catalogue. I am really really sorry that your thorough work on this record may go partially lost though. 
Thank you!

eu-genia:
I would indeed close this one without merging

PR #2667, opened 2024-10-25
stub for Ms Vatican Borgia nr. 27

nafisa-valieva:
Based on Madgalena's request, I have created a stub. Probably I should have asked Massimo's permission first! @MassimoVilla, sorry if I have interfered with your plans! feel free to correct or even delete the record!


MassimoVilla:
Thank you Nafisa. Don't worry. Do you need a full description in a short period of time?

nafisa-valieva:
> Thank you Nafisa. Don't worry. Do you need a full description in a short period of time?

I need it globally :-) but otherwise, the request came from Magdalena, I think she needs the id to be able to refer to it in her work on the Dillmann lexicon. You can work on it when you feel like.

MagdaKrzyz:
Nafisa thank you very much! On my side, I just need the stub.

PR #2676, opened 2024-11-05
2024 10 21 jk uub o etiop 29

karljonaskarlsson:
Description of this manuscript: https://www.alvin-portal.org/alvin/view.jsf?pid=alvin-record:490723, previously catalogued by Löfgren

### DIFF
--- a/Uppsala/UppEt29.xml
+++ b/Uppsala/UppEt29.xml
@@ -66,113 +66,50 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                     <material key="parchment"/>
                                 </support>
                                 <extent>
-                                    <measure unit="leaf">143</measure>
-                                    <measure unit="page" type="blank">8</measure>
-                                    <measure unit="quire">14</measure>
+                                    <measure unit="leaf">38</measure>
+                                    <measure unit="page" type="blank">2</measure>
+                                    <measure unit="quire">5</measure>
                                     <dimensions type="outer" unit="mm">
-                                        <height>225</height>
-                                        <width>180</width>
-                                        <depth>67</depth>
+                                        <height>85</height>
+                                        <width>55</width>
                                     </dimensions>
                                 </extent>
                                 
-                                <foliation>Foliation in pencil in the upper outer corner. The folios <locus from="143" to="144"/>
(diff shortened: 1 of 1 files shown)


CarstenHoffmannMarburg (approved):

Uppsala/UppEt29.xml
```diff
@@ -0,0 +1,280 @@
...
+                                        <width>55</width>
+                                    </dimensions>
+                                </extent>
+                                
+                                <foliation>Foliation in pencil in the upper outer corner.</foliation>
+                                
+                                <collation>
+                                    
+                                    <list>
+                                        <item xml:id="q1" n="1">
```
MassimoVilla, line 89: if this is a quire consisting of guardleaves, I have marked (and found) this quire as `<item xml:id="q1" n="A">` (and similarly, the last quire is, e.g., `<item xml:id="q12" n="B">`). I do not know how widespread this usage is, perhaps it comes from the way guardleaf quires were counted in Ethio-SPaRe. I have not found any specific instruction on this in the guidelines. Are there preferences between `n="A"` and `n="1"`?

Uppsala/UppEt29.xml
```diff
@@ -0,0 +1,280 @@
...
+                                    <q xml:lang="sv">
+                                        1970/21<lb/>
+                                    </q>
+                                </item>
+                            </list>
+                        </additions>
+                        
+                        <bindingDesc>
+                            <binding xml:id="binding">
+                                <decoNote xml:id="b1" type="Boards">Brown wooden boards without leather cover.</decoNote>
```
MassimoVilla, line 209: What about adding `<decoNote xml:id="b2" type="bindingMaterial"><material key="wood"></material></decoNote>` ?

Uppsala/UppEt29.xml
CarstenHoffmannMarburg: I prefer n="A", if it is clearly a guard leaf.

karljonaskarlsson: My thought was that it's better to describe without adding too much interpretation, but I'm open to renumbering the flyleaf quires :)

CarstenHoffmannMarburg (approved):

nafisa-valieva (approved):

PR #2680, opened 2024-11-05
Update Walatta Petros manuscripts

CarstenHoffmannMarburg:
I found six manuscripts containing the phrase "Wälätä P̣eṭros". I corrected the transcriptions as much as I have come across according to Alessandros request in https://github.com/BetaMasaheft/Documentation/issues/2681.


eu-genia:
thank you though I really do not see the much point in fixing EMIP mss, they are many thousand and absolutely all have wrong transcriptions.
in most cases the correct way of dealing with GWP would be add the ref and then no label in the mss records is needed.
@abausi - if you need to fix something specific for quoting, sure we can do it case by case, but we cannot simply fix all EMIP records (or all Cambridge, or all Berlin, or any others, or we willl be doing nothing else).

abausi:
Ok, I never meant all EMIP MSS, as I wrote.

Il mer 6 nov 2024, 10:08 Eugenia ***@***.***> ha scritto:

CarstenHoffmannMarburg:
> BerOrQuart1014 Tanasee179 EMIP2172

I changed the transcription in BerOrQuart1014 and EMIP02172. 
The record of Tanasee179 was all ok, as far as I have seen.

eu-genia (approved):

PR #2714, opened 2024-11-25
Va tet268 corr mk 24.11.24

MarcinKrawczuk:
As in revisionDesc. BTW the reference to Zotero in ms_i7 does not display, is it correct?


CarstenHoffmannMarburg:
> BTW the reference to Zotero in ms_i7 does not display, is it correct?

Thank you for your attentive checking! Me too, I do not understand, why the link does not work. Maybe @eu-genia can help?

PR #2731, opened 2024-12-03
ANLcr1

thea-m:


RomeALincei/ANLcr1.xml
```diff
@@ -138,6 +158,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         </profileDesc>
         <revisionDesc>
             <change who="ES" when="2024-04-10">Created stub</change>
+            <change who="DR" when="2024-12-03">Completed description on the basis of the catalogue description</change>
```
eu-genia, line 161: in accordance with the recent discussion I think you should replace the word `stub` in my `change `with something like "created rudimentary file" or something

eu-genia (approved):

PR #2739, opened 2024-12-10
ANLcr4

thea-m:
I suspect that Conti Rossini was the scribe, but the catalogue does not mention this explicitly. Does anyone happen to know whether this information might be collected somewhere?


nafisa-valieva:
> I suspect that Conti Rossini was the scribe, but the catalogue does not mention this explicitly. Does anyone happen to know whether this information might be collected somewhere?

Interesting! I do not know. In case no one knows, maybe you can formulate your guess in the part where you discuss the origin of the ms. I think it is interesting anyways!

nafisa-valieva (approved):

CarstenHoffmannMarburg (approved):

PR #2746, opened 2024-12-17
Dr 24 12 an lcr3

thea-m:


RomeALincei/ANLcr3.xml
```diff
@@ -28,256 +29,573 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
+                        <summary/>
+                        
                         <msItem xml:id="ms_i1">
-                            <locus from="3ra" to="44vb"></locus>
+                            <locus from="3ra" to="44vb"/>
                             <title ref="LIT4155MiraclesMichael">12 Miracles of St Michael</title>
+                            
                             <msItem xml:id="ms_i1.1">
-                                <locus from="3ra" to="5ra"></locus>
+                                <locus from="3ra" to="5ra"/>
```
eu-genia, line 39: Since all the miracles of St Michael now have individual IDs as well, just like those of Gabriel, I suggest that we refer to them.
Alternatively they can be also referred to with sub-IDs from LIT4155MiraclesMichael (also created), to distinguish cases of transmission as part of the Dersan and part of the Miracles set

eu-genia: So we use LIT1295Dersan#MiracleSailors when the miracle is inside of LIT1295Dersan 
We use LIT7170MiracleMichaelSailors in other occasions
We may also use LIT4155MiraclesMichael#Miracle1 when we have the full set of 12 miracles in LIT4155MiraclesMichael

eu-genia (approved):

CarstenHoffmannMarburg (approved):

PR #2758, opened 2024-12-30
corrected ids

nafisa-valieva:
Having updated lit ids, i have updated ms records as well
https://github.com/BetaMasaheft/Documentation/issues/2707


thea-m (approved):
Great, thank you!

PR #2773, opened 2025-01-12
Create BLorient4943 and BLorient4944

CarstenHoffmannMarburg:
I created records for BLorient4943 and BLorient4944, which are fragments of originally one scroll. They have independant shelfmarks in the library and for this reason, I created two records. I tried to describe the situation in provenance.

grafik

### DIFF
--- a/LondonBritishLibrary/orient/BLorient4943.xml
+++ b/LondonBritishLibrary/orient/BLorient4943.xml
@@ -108,7 +108,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         <handDesc>
                             <handNote xml:id="h1" script="Ethiopic">
                                 <desc>Large, rather careful handwriting.</desc> 
-                                <date notBefore="1700" notAfter="1899" resp="PRS8999Strelcyn"/>
+                                <date notBefore="1700" notAfter="1895" resp="PRS8999Strelcyn"/>
                             </handNote>
                         </handDesc>
                         <decoDesc>
@@ -122,7 +122,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                     </physDesc>
                     <history>
                         <origin>
-                            <origDate notBefore="1700" notAfter="1899">18th/19th century, but not after 1895.</origDate>
+                            <origDate notBefore="1700" notAfter="1895">18th/19th century, but not after 1895.</origDate>
                         </origin>
                         <provenance>A strip from a scroll, probably the last but one of three or four strips. Another slightly longer strip, probably 
                             forming the part of the scroll immediately after, is preserved as <ref type="mss" corresp="BLorient4944"/> in the same library.
--- a/LondonBritishLibrary/orient/BLorient4943.xml
+++ b/LondonBritishLibrary/orient/BLorient4943.xml
@@ -1,0 +1,182 @@
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI xmlns="http://www.tei-c.org/ns/1.0" type="mss" xml:lang="en" xml:id="BLorient4943">
+    <teiHeader xml:base="https://betamasaheft.eu/">
+        <fileDesc>
+            <titleStmt>
+                <title>Fragment of a magic scroll</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                    Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                        licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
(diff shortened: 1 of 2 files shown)


LondonBritishLibrary/orient/BLorient4943.xml
```diff
@@ -0,0 +1,182 @@
...
+                        <msItem xml:id="ms_i1">
+                            <title ref="LIT6609ProtectPrayer" type="incomplete">Prayer against Bāryā, Legewon, Maggāññā and Šotolāy</title>
+                            <note>The beginning is missing.</note>
+                            <incipit xml:lang="gez">ወጸላዒ፡ ባርያ፡ ወሌጌ<sic resp="PRS8999Strelcyn">ዋ</sic>ን፡ መጋኛ፡ ወሾቶላይ፡ ለገብርከ፡</incipit>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i2">
+                            <title ref="LIT7234PPSolomonHand" type="incomplete"/>
+                            <note>Prayer by the virtue of the names which descended from heaven to Solomon and were inscribed on his hands. 
+                                The end is missing.</note>
```
nafisa-valieva, line 40: how do you explain that the end is missing here?

LondonBritishLibrary/orient/BLorient4943.xml
```diff
@@ -0,0 +1,182 @@
...
+                            <incipit xml:lang="gez">አስማተ፡ እዴሁ፡ ለሰሎሞን፡ ኢያስ፡ ኢያያስ፡ አይግባዖስ። በዝንቱ፡ አስማት፡ ወበዝ፡ ቃላት፡ ዘወረደ፡ እምሰማያት፡ ኅቡዕ፡ ቃሉ፡
+                                ለአብ፡ <gap reason="ellipsis"/> ሾተክላሽ፡ ሾላሆም፡ ሽዓት፡ ሸላግ፡ በሲኢለዕ፡ ስሙ፡ ለእግዚአብሔር፡ በዝንቱ፡ አስማተ፡ ሰሎሞን፡ ከመ፡</incipit>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i3">
+                            <title ref="LIT7235PPSolomonSmiths" type="incomplete"/>
+                            <note>The beginning is missing.</note>
```
nafisa-valieva, line 47: how do you explain that the beginning is missing here?

nafisa-valieva (commented):
Thank you, Carsten! it looks confusing and I am not sure that your description is clear enough, although I do not have any better suggestion at hand.

thea-m (changes requested):
Thank you! Technically I think using syntaxe du codex relations to encode the relationship between both records would be possible. It depends on how much time you want to spend on it

LondonBritishLibrary/orient/BLorient4943.xml
```diff
@@ -0,0 +1,182 @@
...
+                        <repository ref="INS00001BL"/>
+                        <idno>BL Oriental 4943</idno>
+                        <altIdentifier><idno>Or. 4943</idno></altIdentifier>
+                        <altIdentifier><idno>Strelcyn 82</idno></altIdentifier>
+                    </msIdentifier>
+                    <msContents>
+                        <summary/>
+                        <msItem xml:id="ms_i1">
+                            <title ref="LIT6609ProtectPrayer" type="incomplete">Prayer against Bāryā, Legewon, Maggāññā and Šotolāy</title>
+                            <note>The beginning is missing.</note>
```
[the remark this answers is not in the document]
CarstenHoffmannMarburg: Strelcyn is not precise on this, but it can be assumed, that the beginning was on the strip before, which is lost. I will try to rephrase it.

LondonBritishLibrary/orient/BLorient4943.xml
CarstenHoffmannMarburg: There is no information in the catalogue and I do not want to give speculative information. Perhaps it was omitted by the scribe or later painted over by one of the pictures.

LondonBritishLibrary/orient/BLorient4943.xml
CarstenHoffmannMarburg: Same as above. Either it was written on the previous strip, which was lost or it was painted over by one of the pictures.

LondonBritishLibrary/orient/BLorient4943.xml
```diff
@@ -0,0 +1,182 @@
...
+                            <decoNote xml:id="d2" type="miniature">
+                                <desc>Coloured <term key="Magic">magic picture</term> in the middle of the strip.</desc>
+                            </decoNote>
+                        </decoDesc>
+                    </physDesc>
+                    <history>
+                        <origin>
+                            <origDate notBefore="1700" notAfter="1895">18th/19th century, but not after 1895.</origDate>
+                        </origin>
+                        <provenance>A strip from a scroll, probably the last but one of three or four strips. Another slightly longer strip, probably
```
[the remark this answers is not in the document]
CarstenHoffmannMarburg: "the last but one of three or four strips". It means, that it probably was the third out of four strips or the second out of three. I was updating this this point after pr. Probably, there was an error in an earlier version. It should be OK, now.

LondonBritishLibrary/orient/BLorient4943.xml
```diff
@@ -0,0 +1,182 @@
...
+                            <supportDesc>
+                                <support>
+                                    <material key="parchment"/>
+                                </support>
+                                <extent>
+                                    <measure unit="leaf">1</measure>
+                                    <dimensions type="outer" unit="mm">
+                                        <height>780</height>
+                                        <width>165</width>
+                                    </dimensions>
```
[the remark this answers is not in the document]
CarstenHoffmannMarburg: This note would lead the user to think that BLorient4943 comprises 3 strips, but infact it is only one out of three, four or even five strips.

CarstenHoffmannMarburg:
> codex relations to encode the relationship between both records would be possible. It depends on how much time you want to spend on it

I would like to spend some time on it. However, I am not sure that such a stratigraphy is helpful in this case. The two strips are not different production units, but simply fragments of what was originally one scroll. I think it is more useful to explain this relationship literally.

CarstenHoffmannMarburg:
@karljonaskarlsson 

See here an example for two fragments of on scroll and the suggestion of @thea-m to encode them according their "syntaxe du codex".

CarstenHoffmannMarburg:
The solution in this record does not work well. The link in BLorient4944 is not represented well and refers to the same record instead of BLorient4943:

grafik

PR #2777, opened 2025-01-15
Create BLorient11804

CarstenHoffmannMarburg:
To my opinion the incipits of ms_i1 and ms_i2 do not contain enough text to make it possible to these prayers with any other record except for the generic records LIT5696MagicPrayer and LIT5888AsmatPrayer respectively. Do you agree, or do you think, it is warranting to create issues for them and look for parallels and eventually create new IDs?

### DIFF
--- a/LondonBritishLibrary/orient/BLorient11804.xml
+++ b/LondonBritishLibrary/orient/BLorient11804.xml
@@ -1,0 +1,156 @@
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI xmlns="http://www.tei-c.org/ns/1.0" type="mss" xml:lang="en" xml:id="BLorient11804">
+    <teiHeader xml:base="https://betamasaheft.eu/">
+        <fileDesc>
+            <titleStmt>
+                <title>Magic scroll</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                    Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                        licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <msDesc xml:id="ms">
+                    <msIdentifier>
+                        <repository ref="INS00001BL"/>
+                        <idno>BL Oriental 11804</idno>
+                        <altIdentifier><idno>Or. 11804</idno></altIdentifier>
+                        <altIdentifier><idno>Strelcyn 84</idno></altIdentifier>
+                    </msIdentifier>
+                    <msContents>
+                        <summary/>
+                        <msItem xml:id="ms_i1">
+                            <title ref="LIT5696MagicPrayer">Prayer for subduing enemies and for catching devils</title>
+                            <note>According to <bibl><ptr target="bm:Strelcyn1978BritishLibrary"/><citedRange unit="page">131</citedRange>
+                                </bibl> with some parts of a 'Hymn to the Cross'.</note>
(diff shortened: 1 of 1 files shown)


CarstenHoffmannMarburg:
I wonder, if I can mark up 'magical characters' as mentioned in a note in ms_i2.

LondonBritishLibrary/orient/BLorient11804.xml
```diff
@@ -0,0 +1,156 @@
...
+                            <title ref="LIT5696MagicPrayer">Prayer for subduing enemies and for catching devils</title>
+                            <note>According to <bibl><ptr target="bm:Strelcyn1978BritishLibrary"/><citedRange unit="page">131</citedRange>
+                                </bibl> with some parts of a 'Hymn to the Cross'.</note>
+                            <incipit xml:lang="gez">በስመ፡ <gap reason="ellipsis"/> ጸሎተ፡ መግረሬ፡ ፀር፡ ወመሣግሪሆሙ፡ ለአጋንንት፡ መስቀል<sic>።</sic> 
+                                መግረሬ<sic>።</sic> ፀር።</incipit>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i2">
+                            <title ref="LIT5888AsmatPrayer">Prayer for expelling devils and demons</title>
+                            <note>Including one line of magic characters</note>
```
eu-genia, line 41: what do you propose as encoding? 

note how Sean Winslow describes them as "charactêres" http://larkvi.com/mss/eth/production/scrolls.php 
which is also the term suggested by the Wikipedia https://en.wikipedia.org/wiki/Charakt%C3%AAres, a Greek term widely used in research https://www.academia.edu/4854168  https://de.wikipedia.org/wiki/Charakt%C3%AAres 

I suggest we use this term (or at least "glyph" which fits better than "character" for things we do not know the meaning of)

I am not sure if it is better to describe them (especially if they are on a separate line as in Sean's figure) as part of text or rather as part of decoration

The latter may be easier.

If we encounter single strange glyphs in text and we decide we must encode them faithfully we could encode each as `g` and then we would need to add a character declaration (`charDecl`) with a `glyph ` to describe or depict each symbol - would mean a schema enhancement and code addition 
https://tei-c.org/release/doc/tei-p5-doc/en/html/WD.html#D25-30

(actually this could also be implemented for the strange letter forms we had previously discussed)

eu-genia (approved):

LondonBritishLibrary/orient/BLorient11804.xml
CarstenHoffmannMarburg: Thank you for your information!

I took that term from Strelcyn's description. 

grafik

I think, these signs/glyphs belong to the text. They might be similar to what Sean Winslows described in his blog, that you linked above. As well Jonas has found recently some of these glyphs in one of the Uppsala manuscripts, that @karljonaskarlsson has described recently (Upp O Etiop 46):

grafik

However, for BLorient11804, I cannot check, because I have no images at hand. It would be useful to make these kind of signs/glyphs searchable even if the individual sign is not available. For now, I will replace the term "magic character" to "magic glyph (charaktêres)" and hope, that this will be understandable to everybody.

CarstenHoffmannMarburg: I wonder, if `<term key="Magic">magic glyphs (charaktêres)</term>` could be helpful?

karljonaskarlsson: I was recently looking for a Bm term for these and found that "Brillenbuchstaben" had been used in some cases. It would probably be good to have a standardized terminology here and also if their presence could be encoded somewhere (a keyword?).

eu-genia: Yes, Brillenbuchstaben was a term used by Uhlig
s. also e.g. https://www.degruyter.com/document/doi/10.1515/9783111697055.150/html?lang=de&srsltid=AfmBOoqs-WSUSJYYtmMozaPL0i1QtbNgsobAgNTeBn6kQXLgDd647skR, they are just heavily stylized.
It seems however that this German term is synonymous with the more general charactêres, see also https://www.fis.uni-hamburg.de/en/publikationen/detail.html?id=9f8feabd-95f0-445f-ba9c-70280fe6c8c1

Indeed, we need to have a standardized glossary/terminology

At the same time we have now introduced the keyword `MagicSymbols `so I think it is appropriate to add there that it can be used to mark up charactêres also known as Brillenbuchstaben in German (in cases when they have circles of course, which they usually do)

karljonaskarlsson: Sounds good!

karljonaskarlsson (approved):

LondonBritishLibrary/orient/BLorient11804.xml
```diff
@@ -38,7 +38,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         </msItem>
                         <msItem xml:id="ms_i2">
                             <title ref="LIT5888AsmatPrayer">Prayer for expelling devils and demons</title>
-                            <note>Including one line of magic characters</note>
+                            <note>Including one line of <term key="Magic">magic glyphs (charaktêres)</term>.</note>
```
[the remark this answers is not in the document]
CarstenHoffmannMarburg: Yes, I see. <term key="MagicSymbols"> is better here.

PR #2778, opened 2025-01-16
Update ESam007.xml

eu-genia:
@DenisNosnitsin1970 I have had a go at this, see decoNote b5


eu-genia:
a very old PR by Pietro who added a b5, never merged, looks OK to me

eu-genia:
but the file is not valid, have to refix

eu-genia:
The PR offers some attributes for textilepastedowns that are not present in schema, probably should be discussed in an issue if they are needed (e.g. type="technique" in `<seg type="technique">hand-drawn, resist-dyed (kalamkari)</seg>`)

I do not know if there has been any development in connection with the Textiles project

DenisNosnitsin1970:
Where die "past downs" come from? Nothing like this visible on the boards, just the common textile inlays.

PR #2779, opened 2025-01-16
BLorient4611_CH_feb24

CarstenHoffmannMarburg:
Apparently, I missed to create a pull request back in february 2024 and therefore this record escaped review and merge.


LondonBritishLibrary/orient/BLorient4611.xml
```diff
@@ -217,7 +217,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                                 </item>
                                 <item xml:id="a2">
                                     <locus from="1v" to="2v"/>
-                                    <desc type="GuestText">Several magic prayers, the first against enemies.</desc>
+                                    <desc type="GuestText">Several magic prayers, 
+                                        the first <ref type="work" corresp="LIT6948PrayerEnemies">against enemies</ref>.</desc>
```
eu-genia, line 221: It seems that the only difference to the version on master is this one and then I wonder if the version on master is more uptodate than the one on this old branch

eu-genia (approved):
If the link to LIT6948PrayerEnemies (the only difference between the version here and the version already on the master branch) is correct, then this PR can be merged.

If the record as it is already merged to master is OK then it does not have to be considered.

LondonBritishLibrary/orient/BLorient4611.xml
CarstenHoffmannMarburg: I think, this was my latest update back in february 2024 and the reason, why it was not merged. The link is not on the master branch, so I think, this is the latest improvement.

PR #2783, opened 2025-01-21
Magda dd 029 10 01 2025

MagdaKrzyz:

PR #2784, opened 2025-01-22
ESdd023 and others

DenisNosnitsin1970:
minor changes


eu-genia:
Editing DD-45 in two branches at the same time (@MagdaKrzyz did https://github.com/BetaMasaheft/Manuscripts/pull/2783) causes conflicts for her.
Please try to avoid working on the same file in different branches as there is a risk in overwriting a more uptodate version (if Magda's branch is merged first, and then this one, the version from this one will be written over hers)

eu-genia (approved):

PR #2787, opened 2025-01-23
Nv 2025 01 23 second commit for titles to appear

nafisa-valieva:
As we discussed yesterday (I hope I understood correctly), I am making the second commit of the files that appear without titles.


CarstenHoffmannMarburg:
I can see no relevant changes beyond the line breaks in oxygen.

nafisa-valieva:
> I can see no relevant changes beyond the line breaks in oxygen.

yes! that´s why I wrote ´no real changes´!

eu-genia (approved):

CarstenHoffmannMarburg:
> yes! that´s why I wrote ´no real changes´! I have assigned you just for your information that the file you created was not fully online (title was missing) there is nothing to review.

It is a mystery to me, how that happened and how I can avoid that in future. Is it, because I did not fill the tag `<title>`?

eu-genia:
no, on first commit manuscripts are visualized without the full shelfmark printed as the title, this is the normal behaviour of the application.

PR #2791, opened 2025-01-23
Magda dd 023 minor changes 23 01 2025 

MagdaKrzyz:
Could you take a look if the visualization after merging will now be ok? 
Dorothea, there are some problems with the way how folia are visualized after merging. Also Philoxenus ID was wrong.
Is the "Extent" section correct? Why are the inlays and the fols. under "Binding decoration" in bold?


thea-m (approved):

thea-m:
Sorry, now I see that there is a validation error in the PR. The file seems to be valid and the error refers to DG001, maybe @eu-genia could have a look? 
Your encoding seems to be fine to me, I cannot say much about the visualisation issue unfortunately. 
Regarding binding: it seems the content of `term` is printed bold. I don't remember the reason for it, I assume it was desirable in some context (maybe decoration description)

thea-m:
I see the validation is a general issue https://github.com/BetaMasaheft/Documentation/issues/2796

thea-m:
And there is already an issue for the extent visualization https://github.com/BetaMasaheft/Documentation/issues/2795

MagdaKrzyz:
Thank you!

eu-genia:
The validation error is because there was an old invalid file merged from an old branch; not everyone has understood what I meant under carefully checking the commits in old branches )

PR #2793, opened 2025-01-24
Maraweh

eu-genia:
This very old branch from Hagos contains a description of a manuscript we do not have; thank you for checking if it can be merged

### DIFF
--- a/DabraSeyon/DabSey002.xml
+++ b/DabraSeyon/DabSey002.xml
@@ -40,6 +40,31 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         </support>
                      </supportDesc>
                   </objectDesc>
+                  <decoDesc>
+<summary>This manuscript circular manuscript and sometimes can be called peackook type manuscript.</summary>
+
+        <decoNote type="frame" xml:id="d1">
+            <locus target="#1"></locus>
+            <desc>circluar hareg decorated in blue and red color</desc>
+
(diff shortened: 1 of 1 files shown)


thea-m (commented):
I would not merge it in its current state, and think that correcting the language to an acceptable level would be so invasive as to need Hagos's approval. So I would be in tentatively in favour of deleting it.

eu-genia:
@HagosAbrha it would be a pity to lose the work you had invested in this - do you feel like revising the record for it to become visible on Bm? Would you be OK with someone heavily revising it? Thank you

DabraSeyon/DabSey002.xml
```diff
@@ -0,0 +1,130 @@
...
+                           <material key="parchment"/>
+                        </support>
+                     </supportDesc>
+                  </objectDesc>
+                  <decoDesc>
+<summary>This manuscript circular manuscript and sometimes can be called peackook type manuscript.</summary>
+
+        <decoNote type="frame" xml:id="d1">
+            <locus target="#1"></locus>
+            <desc>circluar hareg decorated in yelew, blue and red color; it is traditionally called  </desc>
```
nafisa-valieva, line 48: called what?

nafisa-valieva (commented):
I think I know this artistic object and have its images. I wonder, unless Hagos proceeded and did some research, whether we need to keep it. It is very interesting, but for the moment the description is not informative at all.

PR #2794, opened 2025-01-24
Update ESqds004.xml

eu-genia:
this old branch by Hizkiel https://github.com/BetaMasaheft/Manuscripts/compare/master...Hizkiel contains transcriptions for manuscripts, nearly all are already on master but this one is not. If the text is really from the manuscript in question I would suggest on keeping it. Then the old branch can be deleted.


DenisNosnitsin1970:
What should I do, then?

eu-genia:
>If the text is really from the manuscript in question I would suggest on keeping it. 
I need your confirmation that this is the text. And that this can be merged. We have a peer review process.

PR #2795, opened 2025-01-24
Update ESdd023.xml

MagdaKrzyz:
There are some visualization problems under "Origin": there are unnecessary commas among the folia.


eu-genia (approved):

eu-genia:
I do not know what the changes here are; realigning the XML cannot influence anything

eu-genia:
If you are referring to this
image

then the issue is not with XML but with the application code that reads each space as an element in a sequence (it tokenizes by blanks, when you have a line break in XML it equals to several spaces which each is treated as a folio number)

When you see issues with the application you need to create an issue and assign me to change the xquery code behind the visualization, it has nothing to do with this particular file

eu-genia:
but you are lacking a space before the sequence, after `formulas,`

MagdaKrzyz:
The change was that I put the whole string in one line because I thought the programme is disturbed with the line break.  I have corrected the space after "formulas". Can I merge the file now? Then I will send you an issue. Right?

eu-genia:
Merging xml and issues for programming are independent.

PR #2803, opened 2025-01-31
Update BNFabb50.xml

DenisNosnitsin1970:
Update with a new number for Cela Maryam church.


eu-genia:
There is no need to create a second institution

eu-genia (approved):

eu-genia:
not needed as the number had been correct

PR #2807, opened 2025-02-04
Bavet203

MassimoVilla:
Codex containing a Maftǝḥe śǝrāy

### DIFF
--- a/VaticanBAV/et/BAVet203.xml
+++ b/VaticanBAV/et/BAVet203.xml
@@ -47,7 +47,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                       </msItem>
                       <msItem xml:id="ms_i2">
                         <locus from="17r" to="36v"/>
-                        <title type="complete" ref=""/>
+                        <title type="complete" ref="LIT2179Prayer"/>
                         <textLang mainLang="gez"/>
                         <incipit xml:lang="gez">
                           በስመ፡ አብ፡ <gap reason="ellipsis" extent="unknown" resp="PRS4805Grebaut PRS9530Tisseran"/> ጸሎተ፡ ቅዱስ፡ ቆጵርያኖስ፡ ጽንዕ፡ ለነገሥት<supplied reason="omitted">፡</supplied>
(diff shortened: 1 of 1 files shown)


VaticanBAV/et/BAVet203.xml
```diff
@@ -0,0 +1,227 @@
...
+                          <locus from="46v" to="47r"/>
+                          <locus from="47r" to="48r"/>
+                          <dimensions type="outer" unit="mm">
+                            <height>80</height>
+                            <width>55</width>
+                          </dimensions>
+                        </extent>
+                        <collation>
+                          According to the catalogue, the manuscript consists of three quaternions and two quinions, preceded by one protective binion. 
+                          The total amount of folios matches with that provided in the catalogue. However, it is not clear the exact order of the quires.
```
CarstenHoffmannMarburg, line 110: "However, the exact order of the quires is not clear."

VaticanBAV/et/BAVet203.xml
```diff
@@ -0,0 +1,227 @@
...
+                    </handDesc>
+                    <additions>
+                      <list>
+                        <item xml:id="a1">
+                          <desc type="Supplication"/>
+                          <locus target="#47v"/>
+                          <q xml:lang="gez">
+                            አድኅነኒ፡ አም<lb/>ላ<lb/>ኪየ፡ እመከራ፡ <lb/> ሥጋ፡ ወነፍስ፡ <lb/> ለገብርከ፡ ወ<lb/>ልደ፡ ማርያም፡
+                          </q>.
+                          <note>The note mentions a certain <persName ref="PRS14686WaldaMa">Walda Māryām</persName>, possibly a later owner of the manuscript.</note>
```
CarstenHoffmannMarburg, line 135: I used to give information like this in desc, but I do not know, what is the preferred option in Bm.

CarstenHoffmannMarburg (changes requested):

VaticanBAV/et/BAVet203.xml
MassimoVilla: You are right. Better in desc, thank you.

VaticanBAV/et/BAVet203.xml
```diff
@@ -0,0 +1,228 @@
...
+                  </additional>
+                </msDesc>
+            </sourceDesc>
+        </fileDesc>
+      <encodingDesc>
+        <p>A digital born TEI file</p>
+      </encodingDesc>
+      <profileDesc>
+        <particDesc>
+          <listPerson>
```
nafisa-valieva, line 197: I just have noticed that you describe relations in a different way. Is it an alternative that does not affect anything?

MassimoVilla: I haven't found references in the guidelines. Perhaps @thea-m knows more on how people involved should be indicated and if the present way is ok?

nafisa-valieva (approved):

VaticanBAV/et/BAVet203.xml
nafisa-valieva: > I haven't found references in the guidelines. Perhaps @thea-m knows more on how people involved should be indicated and if the present way is ok?

she will be away for a while, but we were using `relations` saws:hasOwned [](uhttps://betamasaheft.eu/Guidelines/?q=relations&start=6&id=relationsrl) (if you check Person to Manuscript)t.

nafisa-valieva: I think you are right, it is in pers records! I am confusing you and myself! I am sorry!

MassimoVilla: Don't worry Nafisa!

CarstenHoffmannMarburg (approved):

PR #2811, opened 2025-02-06
Update ESdd041.xml

MagdaKrzyz:
Two small corrections: removed role from the list, changed "type" in sewing stations.

### DIFF
--- a/ES/ESdd041.xml
+++ b/ES/ESdd041.xml
@@ -317,7 +317,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         <decoNote xml:id="b1" type="Boards">Two wooden boards (the front board is broken and
                            repaired with a cord; the back board is cracked in two places).</decoNote>
                         <decoNote xml:id="b2" type="Cover">Traces of the leather cover.</decoNote>
-                        <decoNote xml:id="b3" type="SewingStations">Two pairs of sewing stations.</decoNote>
+                        <decoNote xml:id="b3" type="Sewing">Two pairs of sewing stations.</decoNote>
                         <decoNote xml:id="b4" type="SewingStations">4</decoNote>
                         <decoNote xml:id="b5" type="bindingMaterial">
                            <material key="wood"/></decoNote>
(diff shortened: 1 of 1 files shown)


ES/ESdd041.xml
```diff
@@ -378,8 +378,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       <profileDesc>
          <particDesc>
             <listPerson>
-               <person><persName ref="PRS14617Zakrestos" role="donor"/>
-                  <persName ref="PRS14616MalkeaKrestos" role="scribe"/>
+               <person><persName ref="PRS14617Zakrestos"/>
```
eu-genia, line 381: you need the roles every time you mention persons with a role in manuscript history

If the persons are already mentioned in the text with roles you do not need to list them again in the particDesc

So what you need to do is add the role everywhere, e.g.
`According to the colophon, the manuscript was commissioned by <persName ref="PRS14617Zakrestos" role="patron"><roleName type="title">ʾabbā</roleName> Zakrǝstos</persName>` etc; the list in line 380 is not needed then

eu-genia: You have the option: either you add the role when you mention the persons, then you do not have to list them. 

Or if you do not mention them or do not want to add the role then you must add a list at the end, and put the role there, only persons with a role in manuscript history are printed as a list in the application (whether appearing in the record or in the separate list, this does not matter)

ES/ESdd041.xml
```diff
@@ -318,7 +317,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         <decoNote xml:id="b1" type="Boards">Two wooden boards (the front board is broken and
                            repaired with a cord; the back board is cracked in two places).</decoNote>
                         <decoNote xml:id="b2" type="Cover">Traces of the leather cover.</decoNote>
-                        <decoNote xml:id="b3" type="Other">Two pairs of sewing stations.</decoNote>
+                        <decoNote xml:id="b3" type="SewingStations">Two pairs of sewing stations.</decoNote>
```
eu-genia, line 320: Sewing, not SewingStations that need the numeric value

ES/ESdd041.xml
MagdaKrzyz: good

ES/ESdd041.xml
MagdaKrzyz: done

DenisNosnitsin1970 (approved):

eu-genia (approved):

PR #2832, opened 2025-02-24
Jacob annuniciation corr emml1763&8509

MarcinKrawczuk:
I know it is confusing but I think when this goes through the manuscript descriptions will reflect what Tedros writes in the article. Then I will clean up the work records and - if it's OK with everybody - create the "parent" work record for CAe 1633, 1683 and 1666.


MarcinKrawczuk:
https://github.com/BetaMasaheft/Documentation/issues/2846

eu-genia (approved):

PR #2835, opened 2025-02-26
Up BL mss (3t., 4t., etc)

CarstenHoffmannMarburg:
I found the abbreviation 3t., 4t., 7t. etc. in various descriptions by St. Strelcyn. First, I did not understand, what this would mean and ignored it, but later I realized, that the meaning was, that the preceding word is written several times. I updated all records with that feature so far.
Screenshot 2025_3t


CarstenHoffmannMarburg:
I will update also LIT records with the texts in question in a minute.

nafisa-valieva (approved):

MagdaKrzyz (approved):

PR #2837, opened 2025-02-27
Magda dd 028 minor changes 2

MagdaKrzyz:
I go for the option `<ref target="#ms_i1.1">Text 1.1, Ps. 100</ref>`, which will be visualized as Text 1.1, Ps. 100 and not 
content.

I have corrected some minor mistakes. 
There is a list of the persons at the end because currently they have not been visualized against the red background in the app.


CarstenHoffmannMarburg (approved):

nafisa-valieva (approved):
Looks good for me! Thank you!

PR #2838, opened 2025-02-27
Magda dd 034 minor changes 3

MagdaKrzyz:
In the current visualization the collation diagrams and the formula do not appear. 
collation

Do you think it is because of the note after quire 2?
Quire table

In the current file I have move the note to the end. Will it help?


CarstenHoffmannMarburg:
I cannot answer your questions on the visualisation, but it must be "further" and not "futher.

eu-genia (approved):

CarstenHoffmannMarburg (approved):

PR #2839, opened 2025-02-28
Magda dd 002 minor changes 

MagdaKrzyz:
The reference to the text is wrong so I have corrected it.
Do you know why the info about the Chi-Rho sign has been duplicated? What shall I do to avoid it? 
And what to do with the hanging dots?! I have noticed that they are taken to the next line after the "locus". 

Here is the encoding: ` <layout>
                                    <ab type="punctuation" subtype="Dividers"/>
                                    <ab type="ChiRho">Chi-Rho sign appears on <locus target="#1v"/>.</ab>
                                 </layout>` 

Visualization: 
chirho

### DIFF
--- a/ES/ESdd002.xml
+++ b/ES/ESdd002.xml
@@ -194,7 +194,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                     <note>All data for <locus target="#1ra"/></note>
                                     <ab type="ruling">Ruling is visible.</ab>
                                     <ab type="pricking">Pricking is not visible.</ab>
-                                    <ab type="ruling" subtype="pattern">Ruling pattern: 1A-1A-1A1A/0-0/0-0/C</ab>
+                                    <ab type="ruling" subtype="pattern">1A-1A-1A1A/0-0/0-0/C</ab>
                                     <ab type="pricking">Primary pricks are not visible.</ab>
                                     <ab type="pricking">Ruling pricks are not visible.</ab>
                                     <ab type="ruling">The upper line is written above the ruling.</ab>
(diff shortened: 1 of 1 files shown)


ES/ESdd002.xml
```diff
@@ -294,7 +294,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                               <material key="parchment"/>
                            </support>
                            <extent><measure unit="leaf" quantity="94"/> 
-                              <measure unit="leaf" type="blank"/>1 <locus target="71v"/>
+                              <measure unit="page" type="blank"/>1 <locus target="71v"/>
```
eu-genia, line 297: 1 must go inside the measure

ES/ESdd002.xml
```diff
@@ -333,8 +333,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         <seg type="ink">Black, red (vivid vermilion).</seg>
                         <desc>Fine, careful. The handwriting remains uniform throughout the manuscript and is characterized by finely shaped, slender and upright letters.</desc>
                        
-                        <seg type="rubrication">Divine names; a few lines on the incipit page of <ref target="#ms_i2"/> and <ref target="#ms_i3"/> alternating with black lines; 
-                           two initial lines of each section of <ref target="#ms_i1"/>, <ref target="#ms_i2"/> and <ref target="#ms_i3"/>; elements of the punctuation signs;
+                        <seg type="rubrication">Divine names; a few lines on the incipit page of <ref target="p2_i1"/> and <ref target="p2_i2"/> alternating with black lines;
```
eu-genia, line 336: targets always need `#`

ES/ESdd002.xml
eu-genia: you might also want to remove the empty element `<ab type="punctuation" subtype="Dividers"/>` or fill it with data

eu-genia (changes requested):

ES/ESdd002.xml
MagdaKrzyz: done

ES/ESdd002.xml
MagdaKrzyz: done
How about the lines `  <licence target="http://creativecommons.org/licenses/by-sa/4.0/">`
`<ptr target="bm:EthioSpare"/>`.

eu-genia: only internal targets of course, not external ones

eu-genia (approved):

PR #2868, opened 2025-03-12
Update ESdd007.xml

MagdaKrzyz:
I had to make some changes in the msContent. I have deleted some roles if they were stated between the encoding.


thea-m (approved):
EVerything is fine, but for my understanding, could you explain to me the improvement of deleting the person roles from within the description?

MagdaKrzyz:
I deleted them where it is already said that it is a donor or an owner. 
Like here: 
Role_name

I added the list at the end so the persons will appear against the red background. 
I do not merge and wait for your response.

nafisa-valieva (approved):
Wonderful! Thank you! (I see your point about roles and wait to see what Dorothea replies)

thea-m:
It's fine, I was just curious :)

PR #2872, opened 2025-03-14
Update ESdd007.xml

MagdaKrzyz:
The placement of the common ID for Haymanote Abew and Mashafa Tomar is not felicitous. Now in the summary there are 4 texts! I have changed it so that now there is only a note with the ID.

### DIFF
--- a/ES/ESdd007.xml
+++ b/ES/ESdd007.xml
@@ -273,7 +273,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                      </origPlace>
                      <origDate when="1610" source="#coloph1">According to the <ref target="#coloph1">colophon</ref>, <ref target="#p1">Unit 1</ref> was completed in <date>1610</date>.
                      </origDate>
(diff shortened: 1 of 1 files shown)


nafisa-valieva (approved):
Looks good! Thank you!

thea-m:
I am not sure whether with this encoding the manuscript will be shown as witness on the page LIT5769HaymanTomar. You can merge and try whether it will be or not.

MagdaKrzyz:
The visualisation looks good now but.... as you predicted the ms. is not shown as a witness. What can we do?

eu-genia:
you can manually add the ms to the witness list of the work record, no other way if it is not pointed from title. you can add there in the abstract that the composition was sometimes achieved "artificially".

one could also possibly think of adding the link to LIT5769HaymanTomar as msItem in the main msContents, before the parts start

MagdaKrzyz:
@eu-genia  "one could also possibly think of adding the link to LIT5769HaymanTomar as msItem in the main msContents, before the parts start"
I did before but in the summary there are 3 texts and so it is written at the top: Number of Text units: 3.
Do you think I can avoid this somehow?

PR #2890, opened 2025-03-24
Dormitio 24.03.25

MarcinKrawczuk:
I have corrected some work records so now I think the distribution of work records is correct. 

Also, I think it would be justified to link the two work records with a relation (see the attached entry from CANT which would suggest that 2619 is an epitome of 1928) - do you think that saws:isRelatedTo would be appropriate?

https://github.com/BetaMasaheft/Documentation/issues/2860

[
Zrzut ekranu 2025-03-24 230105
](url)

### DIFF
--- a/ParisBNF/abb/BNFabb128.xml
+++ b/ParisBNF/abb/BNFabb128.xml
@@ -35,7 +35,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                     <msContents>
                       <msItem xml:id="ms_i1">
                         <locus from="1r" to="18v" cert="medium"/>
-                         <title type="complete" ref="LIT1928Mashaf"/>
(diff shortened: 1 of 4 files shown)


eu-genia (approved):
thank you! a relation is always good

CarstenHoffmannMarburg (approved):

eu-genia:
@MarcinKrawczuk thank you! I have deleted the branch on the remote, but you may want to check that you also delete used branches on your desktop as old branches eat up a lot of space (you can also every now and again delete the entire folder, e.g. Manuscripts, and reclone it, this would delete local branches)

MarcinKrawczuk:
Sorry, I had no idea that the old branches should be deleted 😬 Will do of
course.

wt., 25 mar 2025 o 13:25 Eugenia ***@***.***> napisał(a):

PR #2892, opened 2025-03-25
2025 03 21 jk uub o etiop. 9

karljonaskarlsson:
Description based on the catalogues. No images available online.

### DIFF
--- a/Uppsala/UppEt9.xml
+++ b/Uppsala/UppEt9.xml
@@ -34,8 +34,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                     <msContents>
                         <msItem xml:id="ms_i1">
                             <locus from="5" to="172r"/>
-                            <title type="complete"><foreign xml:lang="gez">Wəddāse ʾAmlāk</foreign> and
-                                <foreign xml:lang="gez">ʾAstabqʷəʿot ʾənta za-Səmʿon ʿAmdāwi</foreign></title>
+                            <title type="complete"><ref type="work" corresp="LIT2505Weddas"/> and
+                                <ref type="work" corresp="LIT6468Astebeqweot"/>
+                                </title>
                             <note><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">46</citedRange></bibl>
                                 notes that the prayers of the two works are either placed before or after each other. The chapters are not 
                                 numbered, and the distribution of texts on days differs from <ref type="mss" corresp="UppEt8"/>.
--- a/Uppsala/UppEt9.xml
+++ b/Uppsala/UppEt9.xml
@@ -1,0 +1,335 @@
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/Schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI xmlns="http://www.tei-c.org/ns/1.0" type="mss" xml:lang="en" xml:id="UppEt9">
+    <teiHeader xml:base="https://betamasaheft.eu/">
+        
+            <fileDesc>
+            <titleStmt>
+                <title xml:id="t1">
+                    Wəddāse ʾAmlāk, ʾAstabqʷəʿot ʾənta za-Səmʿon ʿAmdāwi
+                </title>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/">
+                            This file is licensed under the Creative Commons
+                            Attribution-ShareAlike 4.0.
+                    </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <msDesc xml:id="ms">
+                    <msIdentifier>
+                        <repository ref="INS0356Upp"/>
+                        <collection>O Etiop.</collection>
+                        <idno>O Etiop. 9</idno>
+                        <altIdentifier><idno>Löfgren 15</idno></altIdentifier>
+                        <altIdentifier><idno>Zetterstéen IX</idno></altIdentifier>
+                        <altIdentifier><idno>Äthiop. Ms. 2</idno></altIdentifier>
+                    </msIdentifier>
+                    
+                    <msContents>
+                        <msItem xml:id="ms_i1">
+                            <locus from="5" to="172r"/>
+                            <title type="complete"><title type="complete" ref="LIT2505Weddas"/> and 
+                                <title type="complete" ref="LIT6468Astebeqweot"/></title>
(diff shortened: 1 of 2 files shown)


Uppsala/UppEt9.xml
```diff
@@ -0,0 +1,323 @@
...
+                    </msIdentifier>
+                    
+                    <msContents>
+                        <msItem xml:id="ms_i1">
+                            <locus from="5" to="172r"/>
+                            <title type="complete"><title type="complete" ref="LIT2505Weddas"/> and 
+                                <title type="complete" ref="LIT6468Astebeqweot"/></title>
+                            <note><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">46</citedRange></bibl>
+                                notes that the prayers of the two works are either placed before or after each other. The chapters are not 
+                                numbered, and the distribution of texts on days differs from <ref type="mss" corresp="#UppEt8"/>.
```
thea-m, line 41: no `#` in corresp, here and later

Uppsala/UppEt9.xml
```diff
@@ -0,0 +1,323 @@
...
+                        <idno>O Etiop. 9</idno>
+                        <altIdentifier><idno>Löfgren 15</idno></altIdentifier>
+                        <altIdentifier><idno>Zetterstéen IX</idno></altIdentifier>
+                        <altIdentifier><idno>Äthiop. Ms. 2</idno></altIdentifier>
+                    </msIdentifier>
+                    
+                    <msContents>
+                        <msItem xml:id="ms_i1">
+                            <locus from="5" to="172r"/>
+                            <title type="complete"><title type="complete" ref="LIT2505Weddas"/> and
```
thea-m, line 37: I am not sure if the title inside title encoding works - if there is a precedent, great, otherwise I would suggest to write out the content `<title type="complete">Wəddāse ʾAmlāk and ʾAstabqʷəʿot ʾənta za-Səmʿon ʿAmdāwi</title>`

Uppsala/UppEt9.xml
```diff
@@ -0,0 +1,323 @@
...
+                    
+
+                    <physDesc>
+                        <objectDesc form="Codex">
+                            <supportDesc>
+                                <support>
+                                    <material key="parchment"/>
+                                </support>
+                                <extent>
+                                    <measure unit="leaf">174</measure>
```
thea-m, line 180: `   <measure unit="leaf" quantity="174">174</measure>` is better

Uppsala/UppEt9.xml
```diff
@@ -0,0 +1,323 @@
...
+                            </handNote>
+                        </handDesc>
+                        
+                        <additions>
+                            <list>
+                                <item xml:id="a1">
+                                    <locus from="1v" to="4"/>
+                                    <locus target="#172v"/>
+                                    <desc type="GuestText"><title type="complete" ref="LIT7328MalkeaYostos"/>
+                                    </desc>
```
thea-m, line 218: `<q xml:lang="gez"/>`

Uppsala/UppEt9.xml
```diff
@@ -0,0 +1,323 @@
...
+                                </item>
+                            </list>
+                            <list>
+                                <item xml:id="a2">
+                                    <desc type="Unclear">
+                                        <bibl>Without specifying on which folio,
```
thea-m, line 224: I think the element bibl should surround only the reference proper

Uppsala/UppEt9.xml
```diff
@@ -0,0 +1,323 @@
...
+                                            <ptr target="bm:Zettersteen1899Upsala"/><citedRange unit="page">519</citedRange></bibl>
+                                        notes that the names of <persName ref="PRS14760KenfaGabreel"/> and of 
+                                        his son <persName ref="PRS14761TaklaHaymanot"/>
+                                        are written underneath <ref type="item" corresp="#a1"/>.
```
thea-m, line 228: `<ref target="#a1"/>`

Uppsala/UppEt9.xml
```diff
@@ -0,0 +1,323 @@
...
+                            </binding>
+                        </bindingDesc>
+                        
+                    </physDesc>
+                    
+                    <history>
+                        <origin>
+                            <origDate notBefore="1700" notAfter="1899">18th–19th century</origDate>
+                        </origin>
+                        <provenance>The name of the original owner has been erased and substituted
```
thea-m, line 251: This information should also be given as an extra item ,see https://betamasaheft.eu/Guidelines/?id=additionsVaria
But I think it's very useful to repeat it in provenance

thea-m (changes requested):
Great, thank you!

Uppsala/UppEt9.xml
karljonaskarlsson: You are right!

Uppsala/UppEt9.xml
karljonaskarlsson: The guidelines seem to suggest that this CAN be done, not that it has to be done. Or am I reading it wrong?

thea-m (approved):

Uppsala/UppEt9.xml
karljonaskarlsson: Yes, I guess it makes sense both in extras (when it's based on notes/supplications in the manuscript) and in history

Uppsala/UppEt9.xml
```diff
@@ -0,0 +1,329 @@
...
+                        <idno>O Etiop. 9</idno>
+                        <altIdentifier><idno>Löfgren 15</idno></altIdentifier>
+                        <altIdentifier><idno>Zetterstéen IX</idno></altIdentifier>
+                        <altIdentifier><idno>Äthiop. Ms. 2</idno></altIdentifier>
+                    </msIdentifier>
+                    
+                    <msContents>
+                        <msItem xml:id="ms_i1">
+                            <locus from="5" to="172r"/>
+                            <title type="complete"><foreign xml:lang="gez">Wəddāse ʾAmlāk</foreign> and
```
MagdaKrzyz, line 37: This might be a stupid question but why you do not give the work IDs for both texts but only later for their parts.

Uppsala/UppEt9.xml
```diff
@@ -0,0 +1,329 @@
...
+                                    <desc>One secondary owner mentioned in
+                                        supplications is <persName ref="PRS14762Zamaryam" role="owner"></persName>.</desc>
+                                    <q xml:lang="gez"></q>
+                                </item>
+                            </list>
+                        </additions>
+                        
+                        <bindingDesc>
+                            <binding contemporary="true" xml:id="binding">
+                                <decoNote xml:id="b1" type="Boards"><material key="wood">Wooden</material> boards.
```
MagdaKrzyz, line 242: As far as I know we do not give the type "Boards". Please look at the current issue on binding and the guideliness. I do not know why it is so. 
As for the material, there is a separate type, like here: `<decoNote xml:id="b4" type="bindingMaterial"><material key="wood"></material></decoNote>`.
Under "Cover" you might say simply "Blind-tooled leather "

Uppsala/UppEt9.xml
karljonaskarlsson: Not stupid! :) I've tried different possibilities… The schema doesn't allow two LIT IDs within one ref nor two separate refs within the `<title>`. So then I first provided both LIT IDs as text within the title element. But then Dorothea remarked that she wasn't sure that would visualise well and suggested this encoding instead. I'm open to improved ways of encoding it!

Uppsala/UppEt9.xml
karljonaskarlsson: You are right. Thanks for pointing it out!

Uppsala/UppEt9.xml
MagdaKrzyz: I see! Still I think it would be important to have the general LIT IDs. maybe in the note where you say `notes that the prayers of the **two works** (here the IDs) are either placed before or after each other.` 

In your current way of encoding, will the ms. appear in the work records of both texts? You may add them manually as witnesses.

PR #2899, opened 2025-03-31
Create BLorient11620.xml

CarstenHoffmannMarburg:
With new LIT IDs according to our discussions in https://github.com/BetaMasaheft/Documentation/issues/2901 and
https://github.com/BetaMasaheft/Documentation/issues/2900.

### DIFF
--- a/LondonBritishLibrary/orient/BLorient11620.xml
+++ b/LondonBritishLibrary/orient/BLorient11620.xml
@@ -1,0 +1,168 @@
+<?xml version="1.0" encoding="UTF-8"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI xmlns="http://www.tei-c.org/ns/1.0" type="mss" xml:lang="en" xml:id="BLorient11620">
+    <teiHeader xml:base="https://betamasaheft.eu/">
+        <fileDesc>
+            <titleStmt>
+                <title>Magic scroll</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                    Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                        licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <msDesc xml:id="ms">
+                    <msIdentifier>
+                        <repository ref="INS00001BL"/>
+                        <idno>BL Oriental 11620</idno>
+                        <altIdentifier><idno>Or. 11620</idno></altIdentifier>
+                        <altIdentifier><idno>Strelcyn 93</idno></altIdentifier>
+                    </msIdentifier>
+                    <msContents>
+                        <summary/>
+                        <msItem xml:id="ms_i1">
+                            <title ref="LIT7334Nahabt"/>
+                            <incipit xml:lang="gez">በስመ፡ <gap reason="omitted"/> ጸሎት፡ በእንተ፡ ነሀብት፡ አስማተ፡ ሰሎሞን፡ ዘረበቦሙ፡ ለአጋንንት፡ ከመ፡ መርበብ፡ ዘዓሣ፡
+                                ዘባሕር፡ እንዘ፡ ይብል፡ ሰደቃኤል፡ ጸደቃታኤል፡</incipit>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i2">
+                            <title ref="LIT6609ProtectPrayer"/>
+                            <note>Prayer against demons and about the healing of a man possessed with a demon, containing verses from 
+                                <ref type="work" corresp="LIT1560Gospel">Mt. 9:32-4</ref>.</note>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i3">
+                            <title ref="LIT4620Prayer"/>
+                            <incipit xml:lang="gez">በስመ፡ <gap reason="omitted"/> ጸሎት፡ በእንተ፡ አላሁማ፡ ወያኑራሽም፡ ዲጂ፡ <gap reason="omitted"/>
+                                በኃይለ፡ ዝንቱ፡ አስማቲከ፡ ዘረበቦሙ፡ ለአጋንንት፡ ከመ፡ መርበብተ፡ ዓሣ፡ ዘባሕር፡ ወይቤ፡ ሰሎሞን፡ ሎፍሐም፡ ሎፍሐም፡ ሎፍሐም፡ <gap reason="omitted"/>
+                                በኃይለ፡ ዝንቱ፡ አስማቲከ፡ አድኅኖ፡ እምሕማመ፡ ነሀብት፡</incipit>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i4">
+                            <title ref="LIT7276MarkOne"/>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i5">
+                            <title ref="LIT000000000000000000">Prayer for undoing charms, maftǝḥe śǝrāy</title>
+                            <incipit xml:lang="gez">አኽያ፡ ሹራህያ፡ አልሻዳይ፡ <gap reason="omitted"/> በኃይለ፡ ዝንቱ፡ አስማቲከ፡ ፍታህ፡ ሥራዮሙ፡ 
+                                ለመሠርያ<supplied reason="undefined" resp="PRS8999Strelcyn">ን</supplied>፡ እምላዕለ፡ ለ<gap reason="illegible"/>
+                                <gap reason="omitted"/> በከመ፡ ፈታህኮሙ፡ ለጴጥሮስ፡ ወለጳውሎስ፡ ወፈታህከ፡ ውስተ፡ ቴርም፡ ሥራየ፡ ቆጵርያኖስ፡ ከማሁ፡ ፍታህ፡ ሥራየ፡ ብእሲት፡
+                                ወብእሲት፡ ሥራየ፡ ባርያ፡ ወአጋንንት፡ አቅጥናኤል፡ ክፍትናኤል፡ <gap reason="omitted"/> በኃይለ፡ ዝንቱ፡ <!-- Strelcyn added: (MS. ''ዝ sic) -->
+                                አስማቲከ፡ ዘርዝር፡ ወርጽሽጅ፡ ዝንቱ፡ <sic resp="PRS8999Strelcyn">ሥራይ፡</sic> በእሲ፡ ወብእሲት፡ ሥራየ፡ ማሪ፡ ወማሪት፡</incipit>
+                            <explicit xml:lang="gez"></explicit>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                    </msContents>
+                    <physDesc>
+                        <objectDesc form="Scroll">
+                            <supportDesc>
+                                <support>
+                                    <material key="parchment"/>
+                                </support>
+                                <extent>
+                                    <measure unit="leaf">1</measure>
+                                    <dimensions type="outer" unit="mm">
+                                        <height>1490</height>
+                                        <width>140-145</width>
+                                    </dimensions>
+                                    <note>Scroll composed of <measure unit="strips">3</measure> strips.</note>
+                                </extent>
+                                <condition key="deficient">The scroll is cut at the bottom, and the end is missing.</condition>
+                            </supportDesc>
+                        </objectDesc>
+                        <handDesc>
+                            <handNote xml:id="h1" script="Ethiopic">
+                                <desc>Large, careful handwriting.</desc>
+                                <date notBefore="1800" notAfter="1899" resp="PRS8999Strelcyn"/>
+                            </handNote>
+                        </handDesc>
+                        <decoDesc>
+                            <decoNote xml:id="d1" type="miniature">
+                                <desc><term key="MagicDrawing">Coloured magic picture</term>.</desc>
+                            </decoNote>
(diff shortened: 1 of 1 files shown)


LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                                በኃይለ፡ ዝንቱ፡ አስማቲከ፡ ዘረበቦሙ፡ ለአጋንንት፡ ከመ፡ መርበብተ፡ ዓሣ፡ ዘባሕር፡ ወይቤ፡ ሰሎሞን፡ ሎፍሐም፡ ሎፍሐም፡ ሎፍሐም፡ <gap reason="omitted"/>
+                                በኃይለ፡ ዝንቱ፡ አስማቲከ፡ አድኅኖ፡ እምሕማመ፡ ነሀብት፡</incipit>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i4">
+                            <title ref="LIT7276MarkOne"/>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i5">
+                            <title ref="LIT000000000000000000">Prayer for undoing charms, maftǝḥe śǝrāy</title>
```
thea-m: Here an ID is still missing

LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                    <msIdentifier>
+                        <repository ref="INS00001BL"/>
+                        <idno>BL Oriental 11620</idno>
+                        <altIdentifier><idno>Or. 11620</idno></altIdentifier>
+                        <altIdentifier><idno>Strelcyn 93</idno></altIdentifier>
+                    </msIdentifier>
+                    <msContents>
+                        <summary/>
+                        <msItem xml:id="ms_i1">
+                            <title ref="LIT7334Nahabt"/>
```
thea-m, line 32: type="complete" for all items? If this information is available

LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                        </msItem>
+                    </msContents>
+                    <physDesc>
+                        <objectDesc form="Scroll">
+                            <supportDesc>
+                                <support>
+                                    <material key="parchment"/>
+                                </support>
+                                <extent>
+                                    <measure unit="leaf">1</measure>
```
thea-m, line 72: `  <measure unit="leaf" quantity="1">1</measure>` is better

LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                            </list>
+                        </additions>
+                        <accMat type="Comment">A notice accompanies the scroll: "A parchment scroll containing a number of magical prayers in the Ethiopic language. 
+                            Written in a good bold hand of the seventeenth century".</accMat>
+                    </physDesc>
+                    <history>
+                        <origin>
+                            <origDate notBefore="1800" notAfter="1899">19th century</origDate>
+                        </origin>
+                        <provenance>The name of the owner is erased, and partly preserved only at the end of <ref target="ms_i3"/>:
```
thea-m, line 118: `<ref target="#ms_i3"/>`

LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                                        <date when="1937-03-09">9 March 1937</date></q>
+                                </item>
+                            </list>
+                        </additions>
+                        <accMat type="Comment">A notice accompanies the scroll: "A parchment scroll containing a number of magical prayers in the Ethiopic language. 
+                            Written in a good bold hand of the seventeenth century".</accMat>
+                    </physDesc>
+                    <history>
+                        <origin>
+                            <origDate notBefore="1800" notAfter="1899">19th century</origDate>
```
thea-m, line 116: evidence="lettering?" If this information is provided

LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                </xi:fallback>
+            </xi:include>
+        </encodingDesc>
+        <profileDesc>
+            <textClass>
+                <keywords>
+                    <term key="Magic"/>
+                    <term key="Prayers"/>
+                    <term key="Asmat"/>
+                    <term key="demon"/>
```
thea-m, line 151: "demon" is categorized as an Art Keyword in the taxonomy. I'm not sure if it can be used here? If there is a precedent, all good.

LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                            <title ref="LIT7276MarkOne"/>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i5">
+                            <title ref="LIT000000000000000000">Prayer for undoing charms, maftǝḥe śǝrāy</title>
+                            <incipit xml:lang="gez">አኽያ፡ ሹራህያ፡ አልሻዳይ፡ <gap reason="omitted"/> በኃይለ፡ ዝንቱ፡ አስማቲከ፡ ፍታህ፡ ሥራዮሙ፡ 
+                                ለመሠርያ<supplied reason="undefined" resp="PRS8999Strelcyn">ን</supplied>፡ እምላዕለ፡ ለ<gap reason="illegible"/>
+                                <gap reason="omitted"/> በከመ፡ ፈታህኮሙ፡ ለጴጥሮስ፡ ወለጳውሎስ፡ ወፈታህከ፡ ውስተ፡ ቴርም፡ ሥራየ፡ ቆጵርያኖስ፡ ከማሁ፡ ፍታህ፡ ሥራየ፡ ብእሲት፡
+                                ወብእሲት፡ ሥራየ፡ ባርያ፡ ወአጋንንት፡ አቅጥናኤል፡ ክፍትናኤል፡ <gap reason="omitted"/> በኃይለ፡ ዝንቱ፡ <!-- Strelcyn added: (MS. ''ዝ sic) -->
+                                አስማቲከ፡ ዘርዝር፡ ወርጽሽጅ፡ ዝንቱ፡ <sic resp="PRS8999Strelcyn">ሥራይ፡</sic> በእሲ፡ ወብእሲት፡ ሥራየ፡ ማሪ፡ ወማሪት፡</incipit>
```
thea-m, line 60: I think here the expected form would have been ሥራየ

LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                        <msItem xml:id="ms_i4">
+                            <title ref="LIT7276MarkOne"/>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i5">
+                            <title ref="LIT000000000000000000">Prayer for undoing charms, maftǝḥe śǝrāy</title>
+                            <incipit xml:lang="gez">አኽያ፡ ሹራህያ፡ አልሻዳይ፡ <gap reason="omitted"/> በኃይለ፡ ዝንቱ፡ አስማቲከ፡ ፍታህ፡ ሥራዮሙ፡ 
+                                ለመሠርያ<supplied reason="undefined" resp="PRS8999Strelcyn">ን</supplied>፡ እምላዕለ፡ ለ<gap reason="illegible"/>
+                                <gap reason="omitted"/> በከመ፡ ፈታህኮሙ፡ ለጴጥሮስ፡ ወለጳውሎስ፡ ወፈታህከ፡ ውስተ፡ ቴርም፡ ሥራየ፡ ቆጵርያኖስ፡ ከማሁ፡ ፍታህ፡ ሥራየ፡ ብእሲት፡
+                                ወብእሲት፡ ሥራየ፡ ባርያ፡ ወአጋንንት፡ አቅጥናኤል፡ ክፍትናኤል፡ <gap reason="omitted"/> በኃይለ፡ ዝንቱ፡ <!-- Strelcyn added: (MS. ''ዝ sic) -->
```
thea-m, line 59: Maybe he means that the mansucript has ዝንቱዝ፡?

thea-m (changes requested):
Very nice, thank you!

LondonBritishLibrary/orient/BLorient11620.xml
CarstenHoffmannMarburg: There is no information about some parts wanting or missing. Therefore, I can assume, that the prayers are complete as far as this can be stated for a often more or less fluid kind of literature.

LondonBritishLibrary/orient/BLorient11620.xml
CarstenHoffmannMarburg: Yes, I have updated ms_i5 in a second commit.

LondonBritishLibrary/orient/BLorient11620.xml
CarstenHoffmannMarburg: This is not clear from the catalogue description. Strelcyn may have been guided by the lettering, the miniatures and the overall condition of the scroll.

LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                                በኃይለ፡ ዝንቱ፡ አስማቲከ፡ ዘረበቦሙ፡ ለአጋንንት፡ ከመ፡ መርበብተ፡ ዓሣ፡ ዘባሕር፡ ወይቤ፡ ሰሎሞን፡ ሎፍሐም፡ ሎፍሐም፡ ሎፍሐም፡ <gap reason="omitted"/>
+                                በኃይለ፡ ዝንቱ፡ አስማቲከ፡ አድኅኖ፡ እምሕማመ፡ ነሀብት፡</incipit>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i4">
+                            <title ref="LIT7276MarkOne"/>
+                            <textLang mainLang="gez"/>
+                        </msItem>
+                        <msItem xml:id="ms_i5">
+                            <title ref="LIT7335BahaylaAsm">Prayer for undoing charms, maftǝḥe śǝrāy</title>
```
MagdaKrzyz, line 55: The guidliness says that we should have first of all the transcribed title, so maftǝḥe śǝrāy. The translation is not necessary, but I would add also. Dorothea says that there is some inconsistency about giving the titles. She never gives the translation, I give always because I think it is good for the reader. Perhaps we should have an issue on that? 

"The title should be intended as a generic title for this manuscript record, which might refer to contents or identifiers. Other rules apply for [Works](https://betamasaheft.eu/Guidelines?id=works). The title should be made of all the normalized titles of the works contained in the manuscript. If there are many different works which would make the title too long, it is possible to create a more generic title where this makes sense."

LondonBritishLibrary/orient/BLorient11620.xml
thea-m: It's not a problem (it's present in most older records), just for the future for better filtrability

LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                                <desc><term key="MagicDrawing">Coloured magic picture</term>.</desc>
+                            </decoNote>    
+                            <decoNote xml:id="d4" type="frame">
+                                <desc>Frame</desc>
+                            </decoNote>
+                        </decoDesc>
+                        <additions>
+                            <list>
+                                <item xml:id="a1">
+                                    <desc type="AcquisitionNote">On the back of the frame a notice on the acquisition of the scroll.</desc>
```
MagdaKrzyz, line 105: Please check, but instead of "notice" (announcemnet) "note" is better to my mind.

LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                                    <q>Presented by Mrs. <persName role="bequeather" ref="PRS14769Needham">Needham</persName>,
+                                        <date when="1937-03-09">9 March 1937</date></q>
+                                </item>
+                            </list>
+                        </additions>
+                        <accMat type="Comment">A notice accompanies the scroll: "A parchment scroll containing a number of magical prayers in the Ethiopic language.
```
MagdaKrzyz, line 111: Here the same.

LondonBritishLibrary/orient/BLorient11620.xml
CarstenHoffmannMarburg: Yes, we discussed the use of demon for texts in magic contexts in https://github.com/BetaMasaheft/Documentation/issues/2627#issuecomment-2468093215. 

I have created many work records with the keyword 'demon' since then (e.g. LIT7282ColicPleurisy).

LondonBritishLibrary/orient/BLorient11620.xml
CarstenHoffmannMarburg: This was also my assumption, even though I did not get the grammatical reason for this. In this case, the encoding
`<sic resp="PRS8999Strelcyn">ሥራይ፡</sic>` might be OK, right?

thea-m: Yes, I think it can stay like this

LondonBritishLibrary/orient/BLorient11620.xml
CarstenHoffmannMarburg: OK. I understood, ዝንቱ፡ may be replaced by ዝ, but I think, your suggested solution is better. I will change to 
`<sic>ዝንቱዝ፡</sic>`

LondonBritishLibrary/orient/BLorient11620.xml
CarstenHoffmannMarburg: You are right! I forgot to delete the title from Strelcyns catalogue. I will do in the next commit.

```
<msItem xml:id="ms_i5">
                            <title type="complete" ref="LIT7335BahaylaAsm"/>
```
It is better to have the same titles in work records and in manuscript records to be more user friendly. It is only for generic titles, that I often try to give more precise titles in the manuscript records.

LondonBritishLibrary/orient/BLorient11620.xml
```diff
@@ -0,0 +1,168 @@
...
+                                    </dimensions>
+                                    <note>Scroll composed of <measure unit="strips">3</measure> strips.</note>
+                                </extent>
+                                <condition key="deficient">The scroll is cut at the bottom, and the end is missing.</condition>
+                            </supportDesc>
+                        </objectDesc>
+                        <handDesc>
+                            <handNote xml:id="h1" script="Ethiopic">
+                                <desc>Large, careful handwriting.</desc>
+                                <date notBefore="1800" notAfter="1899" resp="PRS8999Strelcyn"/>
```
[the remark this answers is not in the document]
CarstenHoffmannMarburg: OK, but "nineteenth century script" in this case.

I think, that such a phrase could be misleading in this case, because Strelcyn gives no information, on whether his dating is based on the lettering or on other grounds (as decorations, provenance or the overall condition of the scroll/manuscript). I think, I did the encoding like this for all BL records so far.

grafik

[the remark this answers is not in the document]
MagdaKrzyz: sure, I just took an example from the guidliness

MagdaKrzyz (approved):

thea-m (approved):

PR #2900, opened 2025-03-31
Update ESdd007.xml

MagdaKrzyz:


thea-m (approved):

ES/ESdd007.xml
```diff
@@ -38,6 +38,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                </msIdentifier>
                <msContents>
                   <summary/>
+                  <msItem xml:id="ms_i1">
```
CarstenHoffmannMarburg, line 41: I wonder, how ms_i1 is correct for a composite manuscript. I did not know, that the same texts can be encoded both as ms_i1 and as p1_i1/p2_i1.

CarstenHoffmannMarburg (approved):

ES/ESdd007.xml
CarstenHoffmannMarburg: I see, that it is a useful solution. As far as the visualisation turns out to be well formed, I agree, that you should keep it. Another option could be to mention LIT5769HaymanTomar in a note in p1_i1 and p2_i1  to make it more easy to understand to the user. E.g. 

`<note>The work was transmitted together with ... forming the collection <ref type="work" corresp="LIT5769HaymanTomar"/> as it is usually the case.</note>`

But this is your choice.

MagdaKrzyz: Yes, this is what I had last time, but the ms. is not listed under LIT5769HaymanTomar.

CarstenHoffmannMarburg: Yes, this may be the best solution.

MagdaKrzyz: Ok, thanks!

CarstenHoffmannMarburg (approved):

PR #2902, opened 2025-03-31
Update BL mss (type in ms_items and quantity in measure)

CarstenHoffmannMarburg:
I updated some of the BL mss-records for the type in ms-Items and for the quantity in measure (both leaf and strips as well as page in case of one concertina). Please check and tell me, if you like it. If it is Ok, I want to update the rest of around 90 manuscripts in these points.


thea-m (approved):
It looks good to me!

PR #2904, opened 2025-04-02
2024 12 02 j kd23

karljonaskarlsson:
I prematurely created a PR for an unfinished file. Please wait with reviewing it until I tell you :)


thea-m:
@karljonaskarlsson if you foresee that still several working session are necessary, you could also close the PR and open a new one when the record is finished (this would reduce the amount of notifications we get)

karljonaskarlsson:
> @karljonaskarlsson if you foresee that still several working session are necessary, you could also close the PR and open a new one when the record is finished (this would reduce the amount of notifications we get)

So I simply close it here and then continue working in the same branch?

thea-m:
Exactly!

PR #2906, opened 2025-04-03
Magda dd 013 minor changes 

MagdaKrzyz:
There were some mistakes. Do I now refer correctly to ms. DD-025 (l. 350)? 

How do we refer to the items of the Senkessar: commemorative NOTES or NOTICES?
I think NOTE is correct. In the work ID there is "notices" in the titles. So in the description I have both words. Which word do you use? 

synk


thea-m (approved):
Thank you, it looks good to me! 
I say commemorations, but this is from my point of view not prescriptive. 
`<ref type="mss" corresp="ESdd025"/> (l. 350)` would work, I think :) It is very commendable to try to encode all details, but I think here there is no need for it.

CarstenHoffmannMarburg (approved):

MagdaKrzyz:
commemorations or commemorative notes (we have commemorativeNote as a type) but rather not NOTICE. I would change that in the main record.

PR #2818, opened 2025-02-10
Create BNFabb88.xml

nafisa-valieva:

PR #2907, opened 2025-04-03
2025 03 28 jk o etiop. 17

karljonaskarlsson:
Created description based on Löfgren

### DIFF
--- a/Uppsala/UppEt17.xml
+++ b/Uppsala/UppEt17.xml
@@ -33,6 +33,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         <msItem xml:id="ms_i1">
                             <locus from="3" to="64r"/>
                             <title type="complete" ref="LIT1834Mahlet"/>
+                            <textLang mainLang="gez"/>
                             <note><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">80</citedRange></bibl>
                                 notes that the number of stanzas and their order agrees with the edition in
                                 <bibl><ptr target="bm:Grohmann1919Marien"/></bibl>.
@@ -59,8 +60,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                             <msItem xml:id="ms_i2">
                                 <locus from="64v" to="85r"/>
                                 <title type="complete" ref="LIT2285Saqoqa"></title>
+                                <textLang mainLang="gez"/>
                                 <note><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">80</citedRange></bibl>
-                                    notes that this attestation of the text contains 55 stanzas.
+                                    notes that this attestation of the text contains 55 stanzas and suggests that the three stanzas saluting body parts
+                                    could .
                                 </note>
                                 <incipit xml:lang="gez">
                                     <gap reason="omitted"/>ቆቃወ፡ ድንግል፡ እጽሕፍ፡ በቀለመ፡ አንብዕ፡ ወአንጠብ<gap reason="omitted"/> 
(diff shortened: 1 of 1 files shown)


Uppsala/UppEt17.xml
```diff
@@ -0,0 +1,243 @@
...
+                            </explicit>
+                        </msItem>
+                            
+                            <msItem xml:id="ms_i2">
+                                <locus from="64v" to="85r"/>
+                                <title type="complete" ref="LIT2285Saqoqa"></title>
+                                <textLang mainLang="gez"/>
+                                <note><bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">80</citedRange></bibl>
+                                    notes that this attestation of the text contains 55 stanzas and suggests that the three stanzas saluting body parts
+                                    could .
```
MagdaKrzyz, line 66: Something has escaped after "could"

Uppsala/UppEt17.xml
```diff
@@ -0,0 +1,243 @@
...
+                            
+                        </objectDesc>
+                        
+                        
+                        <handDesc>
+                            <handNote xml:id="h1" script="Ethiopic">
+                                <desc>Mid-sized hand (<foreign xml:lang="de">von mittlerer Hand</foreign>, 
+                                    <bibl><ptr target="bm:Loefgren1974Katalog"/><citedRange unit="page">82</citedRange></bibl>).
+                                    </desc>
+                                <date notBefore="1800" notAfter="1899">19th century.</date>
```
MagdaKrzyz, line 155: Needed age of the script as below from the guideliness:
`<handNote script="Ethiopic" xml:id="h1" corresp="#p1.2 #p1.3 #a4">
            **<seg type="script">First half of the twentieth-century script.</seg>**
            <seg type="ink">Black, red.</seg>`

Uppsala/UppEt17.xml
```diff
@@ -0,0 +1,243 @@
...
+                            </item>
+                            </list>
+                            
+                        </additions>
+                        
+                        <bindingDesc>
+                            <binding contemporary="true" xml:id="binding">
+                                <decoNote xml:id="b1" type="Boards">Wooden boards.
+                                </decoNote>
+                                <decoNote xml:id="b2" type="Other">Wrapped in fabric.
```
MagdaKrzyz, line 176: "Wrapped in fabric." will appear then under "Binding decoration". Is that ok? I would also give it to "Other" if "Other" would really mean "Other". But it means something other then "Other".

Uppsala/UppEt17.xml
```diff
@@ -0,0 +1,243 @@
...
+                                    secondarily in five places.
+                                </desc>
+                            </item>
+                            </list>
+                            
+                        </additions>
+                        
+                        <bindingDesc>
+                            <binding contemporary="true" xml:id="binding">
+                                <decoNote xml:id="b1" type="Boards">Wooden boards.
```
MagdaKrzyz, line 174: Please ask Zhenia, but I understand that we should not add type to b1. 
Here quotation from Zhenia's message 
"Formerly untyped notes (meaning decoNote without @type) were not visualized (except for b1, which MUST be without type https://betamasaheft.eu/Guidelines/?id=bindingDescription), though the guidelines do not say that a type is obligatory and there are examples with untyped b2 in the guidelines." Issue #2818

MagdaKrzyz (approved):
Muchos Dankes!

Uppsala/UppEt17.xml
karljonaskarlsson: You are right – I had forgotten (or rather, I was copying from a merged file that had it like this…). Maybe b2 could also remained untyped?

Uppsala/UppEt17.xml
karljonaskarlsson: Maybe b2 could remain untyped?

Uppsala/UppEt17.xml
thea-m: Both are examples from the records that were used in the guidelines, their exact formulations are not prescriptive. I know that this is not ideal and very confusing. If anyone has interest and energy to improve the guidelines, this is always welcome :) I'm afraid I don't have the mental bandwith for this task now.

karljonaskarlsson: So the guidelines accept several ways of encoding the date of a hand? :) My "feeling" is that it is good to encode the date also as code, not just write the date in words.

thea-m: Yes. It is possible to describe the hand without date or to insert the date wherever seems useful in the description. My personal default is 
```
   <desc>Large and careful hand.</desc>
                                <seg type="ink">Black, red.</seg>
                                <persName ref="" role="scribe"/>
                                <date notBefore="1900" notAfter="1949"/>
```
I think that for example   `<desc>Large and careful hand of the <date notBefore="1900" notAfter="1949">first half of the twentieth century</date> by the scribe <persName ref="" role="scribe"/></desc>` would be equally possible.

There have been several discussions in the past (Denis' practice differs somewhat from mine): https://github.com/BetaMasaheft/Documentation/issues/865, https://github.com/BetaMasaheft/Documentation/issues/2573

I think that keeping the current flexible system makes sense considering the variety of the sources we work with.

Uppsala/UppEt17.xml
MagdaKrzyz: I am not sure. If you do not insert the type, the description occurs under the type directly above. If there is no type in b1, I do not know what will pop up. 
I have quoted your example in the issue about binding. Maybe for the moment put it were it suits you.

Uppsala/UppEt17.xml
MagdaKrzyz: I don't know what will be the result.

Uppsala/UppEt17.xml
```diff
@@ -0,0 +1,244 @@
...
+                        </bindingDesc>
+                        
+                    </physDesc>
+                    
+                    <history>
+                        <origin>
+                            <origDate notBefore="1800" notAfter="1899">19th century</origDate>
+                        </origin>
+                        
+                        <acquisition>The catalogue does not specify when or how the manuscript reached its current repository.</acquisition>
```
CarstenHoffmannMarburg, line 189: I find, it is not useful to give the information, that there is no information.

PR #2908, opened 2025-04-03
Update TarikaWA mss

CarstenHoffmannMarburg:


Berlin/PetermannIINachtrag24.xml
```diff
@@ -50,11 +88,37 @@ schematypens="http://relaxng.org/ns/structure/1.0"?>
...
+                        <locus from="68va"/>
+                        <seg type="script">18th century</seg>                       
+                        <desc>Careless and small script.</desc>
+                     </handNote>
+                  </handDesc>
                </physDesc>
+               <history>
+                  <origin>
+                     <origDate notBefore="1700" notAfter="1799" evidence="lettering" cert="low">18th century</origDate>
+                     <note>The manuscript is a palimpsest and therefore looks older than it actually is. The earlier text may date from the 14th
```
thea-m, line 118: I think "looks older than it actually is" could be potentially misleading. Do I understand it correctly that the undertext is actually old, and Walda Amid, which is the overtext, might appear older than it is because the parchment itself actually is older? 
This manuscript is mentioned in Ted Erho's article on palimpsests and I think in "Die Entdeckung verlorener Texte: Foto- und Textarbeiten
am Untertext einer altäthiopischen Handschrift", so maybe "without technical means" is now outdated?

thea-m (approved):
Great, thank you!

Berlin/PetermannIINachtrag24.xml
CarstenHoffmannMarburg: Yes, this is how I understand Dillmanns statement. It is probably not really neccessary to speak about the lookout. Thank you for the hint to Teds article. I will update it in a minute.

CarstenHoffmannMarburg: I found the article by Stuckenbruck and Rabin in [Bibliotheksmagazin, 2018, p. 71-76](https://www.bsb-muenchen.de/fileadmin/pdf/publikationen/bibliotheksmagazin/BM2018_2.pdf) (bm:StuckenbruckRabin2018Entdeckung), which is based on Ted Erhos work. Do you know any other contribution by Ted himself, which may eventually give more details and insights on the deciphered texts? 

From the short information in bm:StuckenbruckRabin2018Entdeckung, I learnt, that there have been identified more than ten codicological units with a number of textual units in the undertext. Most of them can be identified with already existing CAe, even though I do not have incipits or further information. I think, they are important witnesses and therefore deserve being encoded in detail. I decided for `<origin>` as the best place to do it. Do you agree?
```
<origin>
                     <origDate notBefore="1650" notAfter="1799" evidence="lettering" cert="low">Late 17th or 18th century</origDate>
                     <note>The manuscript is a palimpsest and the current text is written on an earlier layer of writing containing fragments of the
                        following works:
                        (1) <ref type="work" corresp="LIT3159Cyprian"/> from the 13th century;
                        (2) <ref type="work" corresp="LIT1019Actsof">Act. 9,10-11,22</ref> from the 13th century; 
                        (3) a lectionary on the <ref type="work" corresp="LIT3530OldTes"/> with texts of <ref type="work" corresp="LIT3148Zephan"/>;
                        <ref type="work" corresp="LIT3529Daniel"/>; <ref type="work" corresp="LIT1672Isaiah"/>,
                        <ref type="work" corresp="LIT1567Bookof"/>; <ref type="work" corresp="LIT3500Chroni"/>
                        <ref type="work" corresp="LIT2637Deuteronomy"/>;
                        (4) a <ref type="work" corresp="LIT000000000000000000000000">sermon collection for Pentecost</ref> from the 14th century;
                        (5) a <ref type="work" corresp="LIT00000000000000000000000">homily for Easter</ref> by 
                        <persName ref="PRS8121RetuaH"/> from the 14th century;
                        (6) a <ref type="work" corresp="LIT00000000000000000000000">narration</ref> of the finding of the 
                        <ref type="authFile" corresp="TrueCross">True Cross</ref> and the beginning of the 
                        <ref type="work" corresp="LIT6512Zenobius"/> from the same codex of the 14th century;
                        (7) <ref type="work" corresp="LIT6866Testaments"/> of the 14th/15th century;
                        (8) <ref type="work" corresp="LIT1893Martyr"/> from the 15th century;
                        (9) a number of not identifiable <ref type="authFile" corresp="Chants"/> liturgical chants from the 15th century;
                        (10) <ref type="work" corresp="LIT1340EnochE"/> from the early 16th century, all of which have been described in 
                        <bibl><ptr target="bm:StuckenbruckRabin2018Entdeckung"/></bibl>.</note>
                  </origin>
```

However, I do not know, if we have already, what is called "a sermon collection for Pentecost" (4); "a homily for Easter by Rǝtuʿa Hāymānot" (5); and "a narration of the finding of the True Cross" (6) in Stuckenbruck/Rabins article. Do you think, it is reasonable to create new LIT IDs for these three texts, which cannot be clearly identified without text quotations? Are shall I leave in unmarked just with the references to PRS8121RetuaH and TrueCross?

grafik
grafik

thea-m: More information on the content is available in https://www.degruyterbrill.com/document/doi/10.1515/9783111552170-015/html
Regarding the correct place of encoding, I think a note in origin works as a quick solution. For more detailed encoding (I don't know how important this is to your work), I think an issue should be opened to discuss this, because I found nothing in the guidelines for palimpsest encoding. It would be good to check what other TEI projects are doing. Maybe msPart would work?

ParisBNF/abb/BNFabb20.xml
```diff
@@ -162,22 +165,18 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
-                                                    die Syrer, die Gebiete von Rabiʿa, die Insel Waḥarāmqā, die Chaldäer - das sind die Leute von Babylon, Persien, Khorasan,
-                                                    Feryā, Sind, Indien, Jemen, Ṭayǝf, Tamāmā sowie zwei Meere und die Araber. Diejenigen, die mit Buchstaben
-                                                    zu schreiben wissen, sind acht: Hebräer, Śǝyanāwi, Perser, Inder,
-                                                    Chaldäer, Sini, Ḫāmāzi und Araber. An Gewässern
-                                                    haben sie den Euphrat und den Strom Balaka.</ab>
+                                        <ab>The number of people gathered for the construction of the tower was 72 captains. Among them were of Shem's 
+                                            descendants, 25 men. They lived from the Euphrates to Syria and to the eastern edge. They included the Syrians, the areas
+                                            of Rabiʿā (= Diyar Rabi'a), the island of Wäḥarāmqā (= Ǧarāmiqa), the Chaldeans - that is the people of Babylon - as well
```
karljonaskarlsson, line 175: Should it be "Waḥarāmqā" according to the current transcription?

CarstenHoffmannMarburg: Correct!

PR #2913, opened 2025-04-04
Dr 25 04 anlcr10

thea-m:


RomeALincei/ANLcr10.xml
```diff
@@ -0,0 +1,229 @@
...
+                                        catalogue does not state this clearly).</ab>
+                                </layout>
+                            </layoutDesc>
+                            
+                        </objectDesc>
+                        
+                        <handDesc>
+                            <handNote xml:id="h1" script="Ethiopic">
+                                <desc>Large and careful hand.</desc>
+                                <seg type="ink">Black, red.</seg>
```
MagdaKrzyz, line 97: was it ink or pen?

RomeALincei/ANLcr10.xml
```diff
@@ -0,0 +1,229 @@
...
+                                <persName ref="PRS14783GabraS" role="scribe"/>
+                                <date notBefore="1900" notAfter="1949"/>
+                            </handNote>                  
+                        </handDesc>
+                        
+                        <additions>
+                            <list>
+                                <item xml:id="e1">
```
MagdaKrzyz, line 105: Do you think such info should be in extras and not rather as a note under the hand? I am just asking since I am collecting types for the extras. But to my mind it would more fit to keep it with the hand. Here these two pieces of info are one after another but in the app there are far apart.

RomeALincei/ANLcr10.xml
```diff
@@ -0,0 +1,229 @@
...
+                                    <desc>The manuscript was revised, corrected and completed by
+                                        another hand.</desc>
+                                </item>
+                                
+                                <item xml:id="e2">
+<locus target="#26rb"/>
```
MagdaKrzyz, line 111: Again, maybe this info could go to the hand? This fits there even more. Sorry, I am just thinking now about the range of different information that is to be accommodated by Extras.

RomeALincei/ANLcr10.xml
```diff
@@ -0,0 +1,229 @@
...
+                                </item>
+                                
+                                <item xml:id="e2">
+<locus target="#26rb"/>                                   
+                                    <desc>The name of the scribe, mentioned in the final
+                                        supplication, was <persName ref="PRS14783GabraS" role="scribe"/>.</desc>
+                                </item>
+                                
+                                <item xml:id="e3">
+                                    <desc>There are divisions on the monthly reading for the 13th of
```
MagdaKrzyz, line 117: Here comes the type <desc type="findingAid">, oder?

RomeALincei/ANLcr10.xml
```diff
@@ -0,0 +1,229 @@
...
+                                    <q xml:lang="gez">በረከተ፡ ጸሎቱ፡ ይክዐው፡ ፍጡነ፡ ወያፈድፍድ፡ በላዕለ፡ <persName
+                                        ref="PRS14784TanseaK"><roleName type="title">አቡነ፡</roleName> ተንሥአ፡
+                                        ክርስቶስ፡</persName> ወበላዕለ፡ ኵሎሙ፡ ደቂቀ፡ ቢዘን።.</q>
+                                </item>
+                            </list>
+                        </additions>
+                        
+                        <bindingDesc>
+                            <binding contemporary="non-Ethiopian">
+                                <decoNote xml:id="b1">The manuscript consists of a notebook
```
MagdaKrzyz, line 160: "consists of" or rather "is" a notebook?

MagdaKrzyz (commented):
You have interesting cases!

RomeALincei/ANLcr10.xml
thea-m: The catalogue doesn't specify it

RomeALincei/ANLcr10.xml
thea-m: To me, it belongs to extras, because it doesn't actually describe any hand.

RomeALincei/ANLcr10.xml
thea-m: no type is needed for extras, but I can add it! 
I have edited your comment, because markup is only shown if it is marked up :)
image

thea-m:
@MagdaKrzyz Thank you! Would it be possible in the future to write your comments as part of one review process and not as single comments? This would reduce the amount of notifications and also make it clearer when you are finished with the review :)

RomeALincei/ANLcr10.xml
MagdaKrzyz: ok

RomeALincei/ANLcr10.xml
MagdaKrzyz: It must be a tradition because it is not in the guidelines. But I think this is not the right place. Why should it be separated from the description of the hand? One could easily accommodate in the desc.

RomeALincei/ANLcr10.xml
MagdaKrzyz: but if there is a type, I would think it is good to use it. 
Sure, I forgot about the markup. Thanks!

MagdaKrzyz:
Aaaaa.... so when I click on "single comment" it sends each comment separately! I did not know. I will change it immediately, it is indeed irritating.

MagdaKrzyz (approved):

nafisa-valieva (approved):

PR #2931, opened 2025-04-29
Nv 2025.04.29 abb change

nafisa-valieva:
I have added change elements based on the revision history I could retrieve from the pull requests.


thea-m (changes requested):
I applaud the correct use of change elements, and I feel that first and last names would be appropriate in this context

karljonaskarlsson (approved):

thea-m (approved):

PR #2935, opened 2025-05-01
Magda am 012 dd 012 tooling

MagdaKrzyz:
I hope it is ok if there is another keyword within the keyword.


eu-genia:
No it is not ok. Each ref or term or corresp is a link. Please split.

eu-genia:
BTW terms are printed as links when they are part of `desc`
so for binding elements with terms it would be nice to wrap them in desc as in
`<decoNote xml:id="b7" type="Other"><desc><term key="tooling">Blind tooled ornament on the cover</term> (at least 2 tools were used: <term key="tripleStraightLine">triple line fillet</term>, fourfold line fillet).</desc></decoNote>`
would be printed as 
image

MagdaKrzyz:
Corrected

eu-genia (approved):

MagdaKrzyz:
DD-012 merged unsuccessfully, will try again

PR #2937, opened 2025-05-05
updated the desc

nafisa-valieva:
I have introduced a tiny update to the description, which is relevant to Abb. 100, which I am currently describing.


thea-m (approved):
Very nice!

MagdaKrzyz (commented):
Good!

PR #2944, opened 2025-05-08
Update ESdd025.xml

MagdaKrzyz:
I have done only minor corrections on this file.
Do you understand why the monthly dates in lines 290 and 296 have not been visualized?

### DIFF
--- a/ES/ESdd025.xml
+++ b/ES/ESdd025.xml
@@ -287,13 +287,13 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                      <list>
                         <item xml:id="a1">
                            <locus from="160r" to="160v"/>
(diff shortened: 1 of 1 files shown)


CarstenHoffmannMarburg:
I have no experience with commemoration notes. I think, there should be no diacritics in the tag, therefore, probably, it needs to be written `<date ref="ethiocal:Hamle20"/>`, but not `<date ref="ethiocal:Ḥamle20"/>` and `<date ref="ethiocal:Nahase13"/>`), but not `<date ref="ethiocal:Naḥāse13"/>`).

CarstenHoffmannMarburg (approved):

thea-m (approved):
Carsten is right, and in addition, I think that in these cases, the date element needs content (at least this applies to all examples I could find), so: `<date ref="ethiocal:Nahase13">13 Naḥāse</date>` or something similar. I hope that it will work with this!

MagdaKrzyz:
many thanks to both of you

PR #2947, opened 2025-05-13
Add change notes to BL manuscripts (no. 1)

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2700.


MagdaKrzyz (approved):
I have not thought that it is so important to mark who has suggested some corrections. Is it not too much?

PR #2953, opened 2025-05-15
Add change notes to mss records

CarstenHoffmannMarburg:
cf. https://github.com/BetaMasaheft/Documentation/issues/2976


thea-m (commented):
Very admirable addition of change elements, all look great!
Is  Blorient9178 intended to be part of this PR? Some of its IDs have not been filled in yet, so haven't reviewed this file.

CarstenHoffmannMarburg:
> > Is Blorient9178 intended to be part of this PR? Some of its IDs have not been filled in yet, so haven't reviewed this file.
> 
> No, it was not intended, to push this record in this branch. I have deleted it.

To my surprise, it is still in this list. How can I safely delete after committing to the branch?

thea-m:
A good solution is to "save as" the file to another location on your computer (outside of the BetMas project) as a safety copy (if this is needed), then delete the file from this branch (either in oxygen or in your explorer, this makes no difference). Then you can switch branch and re-save the safety copy into BetMas. Right now, I see in the branch that the file has not yet been deleted. Does this help?

thea-m (approved):

PR #2967, opened 2025-05-20
Update ESdd015.xml

MagdaKrzyz:
I have made some small changes. Line 1174 with the info about inlays has not aapeared on the web, do you know why?


thea-m (approved):

thea-m:
I think that it is not an issue with this particular record, but that in general `<term key="textileInlay"/>` is not visualized (maybe this is true for all terms, I only checked this one now) when it does not have content. The alternatives would be to either include content such as `<term key="textileInlay">textile inlays</term> ` or to open a visualization issue

PR #2973, opened 2025-05-21
Emml6694

MarcinKrawczuk:
It's a very basic encoding but I suppose that some of those hagiographies are relatively rare so it will make a nice addition.


CarstenHoffmannMarburg (approved):

CarstenHoffmannMarburg:
Very good! Thank you!

PR #2977, opened 2025-05-22
2025 05 22 jk update ssb 002

karljonaskarlsson:
updated ESssb002 with new LIT IDs for single-type collections (and the content identification based on my dissertation)


nafisa-valieva (approved):
Beautiful reconstruction of the correct order of folia!

PR #2979, opened 2025-05-22
Create ANLcr13.xml

thea-m:


RomeALincei/ANLcr13.xml
```diff
@@ -0,0 +1,548 @@
...
+                                        አናቅፂሃ፡ ወዓረፍታተ፡ ኅንፃሃ፡ እለ፡ ተሐንጹ፡ ቀዳሚ፡ በእደ፡ አበው፡ ቀደምት፡ ቅዱሳን፡
+                                        ወሐፄራነ፡ አምጣን፡ ወትሑታነ፡ ዓቅም፡ በህቁ።</incipit>
+                                </msItem>
+                                
+                                <colophon xml:id="coloph2" xml:lang="gez">
+                                    <locus from="106ra" to="106rb"/>
+                                    <note>The colophon states that the manuscript was commissioned
+                                        in 1938, under the Italian government, during the tenure of
+                                        the Patriarch of Alexandria <persName ref="PRS5714JohnXIX"/>, the
+                                        Metropolitan of Ethiopia <persName ref="PRS14840Abreham">ʾAbrǝhām</persName>, the ʾƎč̣č̣age
```
MagdaKrzyz, line 189: A question: do we need to extend the encoding of the names? `<persName ref="PRS14840Abreham">**ʾAbrǝhām**</persName>`? I can see that sometimes you expand the encoding and sometimes not. I was told not to expand unless there is a different spelling. Is there any reason why you have expanded the names?

RomeALincei/ANLcr13.xml
```diff
@@ -0,0 +1,548 @@
...
+                 <decoDesc>
+                     <summary>All miniatures are behind a <term key="protectiveCurtain">protective curtain</term>
+                         made of <material key="silk">silk</material> or <material key="textile">cloth</material>.
+                         The miniature painter was <persName ref="PRS14838YohannesA"
+                             role="illustrator"/>.
+                     </summary>
+                 
+                     <decoNote xml:id="d1" type="miniature">
+                     <locus target="#10v"/>
+                     <desc></desc>
```
MagdaKrzyz, line 324: perhaps you can remove desc, or they have any role here?

RomeALincei/ANLcr13.xml
```diff
@@ -0,0 +1,548 @@
...
+                         <q xml:lang="gez">ሥሉስ፡ ቅዱስ፡</q>
+                 </decoNote>    
+                     
+                     <decoNote xml:id="d2" type="miniature">
+                         <locus target="#11r"/>
+                         <desc>The archangel Gabriel lets Madḫānina ʾƎgziʾ ascend miraculously to
```
MagdaKrzyz, line 330: "allows Madḫānina ʾƎgziʾ to ascend"

MagdaKrzyz (approved):
Great job! Thank you!

CarstenHoffmannMarburg (approved):

RomeALincei/ANLcr13.xml
thea-m: Both are possible. The advantages of not expanding (`<persName ref="PRS14840Abreham"/>`) are that it shortens the record, minimizes typos and leads to a consistent visualization of all occurences of the same entity. One disadvantage is that if someone searches for "ʾAbrǝhām", a record that contains `<persName ref="PRS14840Abreham"/>` will not be among the results. Ideally, in this case the person record PRS14840Abreham should be among the results and the searcher will be led from the person record to the other records mentioning this ID via attestations, but some people might only skim the list of results and miss this additional piece of information. 
Here, I had no specific reason :)

MagdaKrzyz: Thanks for the explanation!

PR #2984, opened 2025-05-26
Dr 25 05 an lcr12

thea-m:


RomeALincei/ANLcr12.xml
```diff
@@ -0,0 +1,218 @@
...
+                                
+                                <item xml:id="a3">
+                                    <desc type="Supplication">Supplication for the patriarch, the
+                                        metropolitan, the saints and martyrs etc., written on the recto of the guard leaf following the last
+                                        printed page, probably by <persName
+                                            ref="PRS11501Endreyas"/> shortly after the publication
+                                        of the Psalter. The patriarch mentioned at the beginning is
+                                        probably <persName ref="PRS4380Gabriel"/>, which would be corroborated by
+                                        the palaeography (<persName ref="PRS4381Gabriel"/> seems to
+                                        be too late). A metropolitan is not mentioned.</desc>
```
MagdaKrzyz, line 89: `(<persName ref="PRS4381Gabriel"/> seems to
                                        be too late)` I know what you mean but could you write it a bit more clear. Sth. like: Patriarch XY should be excluded from consideration, since he was in office too late.

PR #2985, opened 2025-05-26
Emml2461 mk 4.11.24

MarcinKrawczuk:
Apparently we're doing some cleaning, so this is a mss record, far from complete, but still contains some new info, including work records so far unattested, so perhaps it could be merged in a preliminary state.


eu-genia (approved):
Thank you!

PR #2996, opened 2025-06-04
Magda dd 016 corrections

MagdaKrzyz:
There were some small mistakes in the first version. There is a problem with visualizing the subtitles. Dorothea, do you think that it is the same issue as with Wuddase Amlak yesterday? I have compared my encoding with yours and Marcin's of Fetha negest and all seems to be kosher. 
fetha


thea-m (approved):

thea-m:
I don't know the reason for this visualisation (I have checked FSUor14, which contains the same textparts, and seems to visualized fine, so there seems to be nothing wrong with the encoding)

MagdaKrzyz:
Thank you!

MagdaKrzyz:
With the second merge some of the items emerged, but there are still some pending. Merging helps! Merging the remedy!

PR #3048, opened 2025-07-31
Nv 2025.07.30 abb150

nafisa-valieva:
New record. I will create and encode PersName when the application is back.

### DIFF
--- a/ParisBNF/abb/BNFabb150.xml
+++ b/ParisBNF/abb/BNFabb150.xml
@@ -59,8 +59,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                             <height unit="mm">125</height>
                             <width unit="mm">100</width>
                           </dimensions>
-                          <measure unit="leaf" type="blank">1
-                            <locus target="#148v"></locus></measure>
+                          <measure unit="leaf" type="blank">1</measure>
+                            <locus target="#148v"/>
                         </extent>
                         <foliation>Foliation in the upper right corner of the folia.</foliation>
                         <collation>
@@ -68,7 +68,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                             <locus target="#10r"/> <locus target="#18r"/><locus target="#26r"/><locus target="#34r"/></signatures>
                           <list>
                             <item xml:id="q1" n="A">
-                              <dim unit="leaf">1, half folio before 1</dim>
+                              <dim unit="leaf">1, stub before 1</dim>
                               <locus target="#1"/>
                             </item>
                             <item xml:id="q2" n="1">
@@ -98,7 +98,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                             </item>
                             <item xml:id="q7" n="6">
                               <num value="6">፮</num>
-                              <dim unit="leaf">7, half folio after 7</dim>
+                              <dim unit="leaf">7, stub after 7</dim>
                               <locus from="42r" to="48v"/>
                             </item>
                           </list>
@@ -109,7 +109,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         <layout columns="2" writtenLines="12 14">   
                           <ab type="ruling">Ruling is partly visible. </ab>
                           <ab type="pricking"> Pricking is partly visible.</ab>
-                          <ab type="ruling" cert="unknown"> The upper line is written above the ruling. </ab>
+                          <ab type="ruling"> The upper line is written above the ruling. </ab>
                         </layout>
                       </layoutDesc>
                     </objectDesc>
@@ -149,7 +149,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                      </list>
                     </additions>
                     <bindingDesc>
-                      <binding xml:id="true">
+                      <binding  xml:id="true">
                         <decoNote xml:id="b1">The front is a wooden board consisting of two pieces. The back board is either invisible in the images or is missing.</decoNote>
                       </binding>
                     </bindingDesc>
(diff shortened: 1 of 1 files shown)


CarstenHoffmannMarburg (approved):

ParisBNF/abb/BNFabb150.xml
```diff
@@ -28,6 +28,140 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+                          <material key="parchment"/>
+                        </support>
+                        <extent>
+                          <measure unit="leaf" quantity="48">48</measure>
+                          <dimensions type="outer">
+                            <height unit="mm">125</height>
+                            <width unit="mm">100</width>
+                          </dimensions>
+                          <measure unit="leaf" type="blank">1
+                            <locus target="#148v"></locus></measure>
```
thea-m, line 63: I think we usually put locus after measure in these cases, but I don't know if it makes a difference

ParisBNF/abb/BNFabb150.xml
```diff
@@ -28,6 +28,140 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+                        </extent>
+                        <foliation>Foliation in the upper right corner of the folia.</foliation>
+                        <collation>
+                          <signatures>Undecorated quire marks  are written in the main hand, in the right corner of the upper margin of <locus target="#2r"/>,
+                            <locus target="#10r"/> <locus target="#18r"/><locus target="#26r"/><locus target="#34r"/></signatures>
+                          <list>
+                            <item xml:id="q1" n="A">
+                              <dim unit="leaf">1, half folio before 1</dim>
```
thea-m, line 71: I don't know if "1, half folio before 1" is allowed here (my feeling is no). I have looked at the images and could not understand the situation well from it, and then gallica stopped working. Purely speculating without having been able to look at the complete set of images, could the half folia in the beginning and end be part of one spine reinforcement? If uncertainty remains, I think it would be better to decide on one numeral in dim, and add a note that explains the situation

ParisBNF/abb/BNFabb150.xml
```diff
@@ -28,6 +28,140 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+                            </item>
+                          </list>
+                          </collation>
+                        <condition key="deficient"/>
+                      </supportDesc>
+                      <layoutDesc>
+                        <layout columns="2" writtenLines="12 14">   
+                          <ab type="ruling">Ruling is partly visible. </ab>
+                          <ab type="pricking"> Pricking is partly visible.</ab>
+                          <ab type="ruling" cert="unknown"> The upper line is written above the ruling. </ab>
```
thea-m, line 112: It does not seem to me that cert="unknown" is necessary

ParisBNF/abb/BNFabb150.xml
```diff
@@ -28,6 +28,140 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+                      <list>
+                        <item xml:id="a1">
+                          <locus target="#1"/>
+                          <desc type="GuestText">An incomplete prayer against eye and head pain. <bibl>
+                            <ptr target="bm:Chaine1912Catalogue"/>
+                            <citedRange unit="page">91</citedRange>
+                            <citedRange unit="number">150</citedRange>
+                          </bibl> defined it as 'two magic formulas'.</desc>
+                          <q xml:lang="gez">ይትባረክ፡ እግዚአብሔር፡ ጸሎተ፡ በእንተ፡ ሕምም፡ አይን፡ ወርዕስ፡ መዕመን፡ ኦሪት፡ ወመድኃኒት፡ ዘበምራቅከ፡ ትከሥት፡ አይንተ፡ ዕዉራን፡</q>
+                          <note>According to <bibl>
```
thea-m, line 131: I think it is easier for the reader if the information from note is integrated into desc. Note should only be a last resort if no other element makes sense

ParisBNF/abb/BNFabb150.xml
```diff
@@ -28,6 +28,140 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+                        </item>
+                        <item xml:id="e3">
+                          <locus target="#2r #48r"/>
+                          <desc type="StampExlibris">A circle stamp, with the name
+                            <persName ref="PRS1071dAbbadi" role="owner"/> in it.</desc>
+                        </item>
+                     </list>
+                    </additions>
+                    <bindingDesc>
+                      <binding xml:id="true">
```
thea-m, line 152: type="Ethiopian"?

ParisBNF/abb/BNFabb150.xml
```diff
@@ -28,6 +28,140 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+                        <decoNote xml:id="b1">The front is a wooden board consisting of two pieces. The back board is either invisible in the images or is missing.</decoNote>
+                      </binding>
+                    </bindingDesc>
+                  </physDesc>
+                  <history>
+                    <origin>
+                      <origDate notBefore="1700" notAfter="1800" source="lettering">18th century</origDate>.  This manuscript was collected by <persName
+                        ref="PRS1071dAbbadi" role="owner"/> in Ethiopia.
+                    </origin>
+                    <acquisition>The manuscript entered <placeName ref="INS0303BNF"/> in
```
thea-m, line 162: the BnF

thea-m (changes requested):
Very nice, thank you!

ParisBNF/abb/BNFabb150.xml
nafisa-valieva: It is encoded, why should I repeat it? 'In ' refers to the year

ParisBNF/abb/BNFabb150.xml
nafisa-valieva: > I think it is easier for the reader if the information from note is integrated into desc. Note should only be a last resort if no other element makes sense
Why would I include into description something that is missing? It is an additional information from the catalogues about something that is not there anymore. I do not see how I can integrate it! But probably this note belong somewhere else?

ParisBNF/abb/BNFabb150.xml
thea-m:
```
     <measure unit="leaf" type="blank">1</measure>
                            <locus target="#148v"/>
```

ParisBNF/abb/BNFabb150.xml
thea-m: then leave it

ParisBNF/abb/BNFabb150.xml
thea-m: I think the article is needed linguistically, but who knows

PR #3081, opened 2025-08-28
updates

thea-m:


ES/ESmkl014.xml
```diff
@@ -190,7 +190,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                  of Heaven for ever. Amen.</foreign>
                            </explicit>
                            <note>
-                              The text seems closer to the so-called second recension (<ref type="work" corresp="LIT1473GadlaL#recension2"/>, but with some more passages in the prologue.
+                              The text seems closer to the so-called second recension (<ref type="work" corresp="LIT7616GadLib2"/>, but with some more passages in the prologue.
```
MagdaKrzyz, line 193: Why "closer"? What is the entity that you compare with? Maybe just "... is close"

ES/ESmgm012.xml
```diff
@@ -41,7 +41,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                      <title type="complete" ref="LIT1473GadlaL"/>
                         <incipit xml:lang="gez"><note>A rhymed prologue</note><locus from="1ra" to="1vb"></locus>        በስመ፡ አብ፡ ወወልድ፡ ወመንፈስ፡ ቅዱስ፡ ፩አምላክ፨ በስመ፡ እግዚአብሔር፡ አብ፡ ዘአምቅድመ፡ ዓለም፡ ህላዌሁ፨
                               ወእስከ፡ ለዓለም፡ ዘኢየሐልቅ፡ ምስፍናሁ። ልዑል፡ በልዕልናሁ፨ ስብሕ፡ በስብሐቲሁ፨ ...</incipit>
-                        <note>The text seems closer to the text as in the so-called <title ref="LIT1473GadlaL#recension2">second recension</title>, but with some more
+                        <note>The text seems closer to the text as in the so-called <title ref="LIT7616GadLib2">second recension</title>, but with some more
```
MagdaKrzyz, line 44: The note is unclear. "The text seems closer to the text" (which text???)

MagdaKrzyz (approved):

ES/ESmgm012.xml
MagdaKrzyz: I see but the note is totally unclear.

CarstenHoffmannMarburg: I think, I can understand, but it may be easier to say: `The text seems to be close to the <title ref="LIT7616GadLib2">second recension</title>, but with some more passages in the initial part.`

I think, it is clear, that we speak here about the Gadla Lalibelā, because this is stated in the title.

CarstenHoffmannMarburg (approved):

PR #3127, opened 2025-10-07
Magda dd 031 corrections

MagdaKrzyz:
There were some minor mistakes, plus I had to add two individuals.


CarstenHoffmannMarburg:
> There were some minor mistakes, plus I had to add two individuals.

I did not see any individuals in the record.

PR #3275, opened 2026-02-10
Magda makm 001 05 02 2026

MagdaKrzyz:


ES/ESmakm001.xml
```diff
@@ -25,146 +25,263 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
-                                 <locus from="107r" to="116v"/> ፲XII፲(10/fols. 107r-116v) </item>
-                              <item xml:id="q13">
+                                 <locus from="107r" to="116v"/></item>
+                              <item xml:id="q13" n="13">
+                                 <num value="12">፲፪</num>
                                  <dim unit="leaf">10</dim>
-                                 <locus target="#10"/>
-                                 <locus from="117r" to="125v"/> ፲፪XIII፲፪(10/fols. 117r-125v) </item>
-                              <item xml:id="q14">
+                                 <locus from="117r" to="125v"/></item>
```
CarstenHoffmannMarburg: If there are 10 leafs, this would end up with folio 126

CarstenHoffmannMarburg (changes requested):

e-schwarz (approved):

ES/ESmakm001.xml
```diff
@@ -25,146 +25,263 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                <msIdentifier>
                   <repository ref="INS0174MAKM"/>
                   <collection>Ethio-SPaRe</collection>
-                  <idno>MAKM-001</idno>
+                  <idno facs="MAKM/001/MAKM-001">MAKM-001</idno>
+                  <!--n is missing -->
                   <altIdentifier>
                      <idno>C4-IV-40</idno>
```
eu-genia: the idnos should be switched, we respect the primary local shelfmark provided, just as the HMML does, as E-S numbers are actually numbers assigned to image sets, not necessarily to mss - ESdd024 ESdd025 are correct examples (ideally, `surrogates `section should be added, see issue https://github.com/BetaMasaheft/Documentation/issues/2514 and the Guidelines - especially important when multiple microfiliming/photographing initiatives may have treated the same objects)

eu-genia: Should be ideally
```
<idno facs="MAKM/001/MAKM-001" n="176">C4-IV-40</idno>
                <altIdentifier>
                     <collection>Ethio-SPaRe</collection>
                     <idno>MAKM-001</idno>
                  </altIdentifier>
```

ES/ESmakm001.xml
MagdaKrzyz: That is right, but folio 118 is not numbered.

CarstenHoffmannMarburg (approved):

ES/ESmakm001.xml
MagdaKrzyz: The mss of DD have yet another number, perhaps given earlier than  C4-IV-40. Do we consider it too?

eu-genia: > The mss of DD have yet another number, perhaps given earlier than C4-IV-40. Do we consider it too?

as altIdentifier of course, all shelfmarks must be mapped and be searchable, the important thing is that if there is a more or less reliable/standard shelfmark associated with a manuscript (and Tourism Commission numbers are the best one gets in Ethiopia) it must be shown as the primary one, clearly differentiated from project numbers (even if they are well established like EMML or Tanasee)

MagdaKrzyz:
@DenisNosnitsin1970 what do you think

DenisNosnitsin1970 (commented):
see comments

ES/ESmakm001.xml
eu-genia: Full collections - not quite true. See the way they treat Gunda Gunde collection

eu-genia:
On reduscussing with Bausi:

As I said, the primary shelfmarks are, when present, the TCA ones. Please restore DD manuscripts as they were and revise all others. Thank you.

Only

```
<idno facs="MAKM/001/MAKM-001" n="176">C4-IV-40</idno>

                <altIdentifier>

                     <collection>Ethio-SPaRe</collection>

                     <idno>MAKM-001</idno>

                  </altIdentifier>
</idno>
```

We align to HMML as in the examples above

PR #3337, opened 2026-04-01
Update ESmakm004.xml

MagdaKrzyz:
Minor corrections


ES/ESmakm004.xml
e-schwarz: add space after "elipsis" and before መሐሪት፡

e-schwarz: Shouldn`t the parchment description be moved under the parchment?

MagdaKrzyz: done. Thanks!

MagdaKrzyz: could be, ok

e-schwarz: I think it should be: supplications formulas in <ref target="#ms_i14"/> on <locus target="#67ra #118rb"/>

e-schwarz (changes requested):

e-schwarz (approved):

ES/ESmakm004.xml
MagdaKrzyz: done

PR #3343, opened 2026-04-15
Delete links in PULgarrettEth017

CarstenHoffmannMarburg:
I found that "Link to images" existed twice on https://betamasaheft.eu/manuscripts/PULgarrettEth017/main, but none worked (while clicking on "Images" works well). I think, it is therefore necessary either to update the links in facsimile or to delete them. I found that images are distributed on https://catalog.princeton.edu/catalog/99123667053506421, which is quoted in listBibl, but I am not sure about the link that would fit into facsimile. Therefore I suggest to delete the links in facsimile.

### DIFF
--- a/princeton/PULgarrettEth017.xml
+++ b/princeton/PULgarrettEth017.xml
@@ -31,7 +31,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                             4. Folio 66v-67r. Miracle of Saint Marqoryos [sic]
                             5. Folio 68r-75r. Miracles of the Infant Jesus 10 miracles.</summary>
                   <msItem xml:id="ms_i1">
-                     <title>Miracles of Mary</title>
+                     <title ref="LIT2384Taamme">Miracles of Mary</title>
                      <msItem xml:id="ms_i1.1">
                         <title>Introduction</title>
                      </msItem>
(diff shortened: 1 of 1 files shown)


princeton/PULgarrettEth017.xml
```diff
@@ -103,19 +103,14 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
-            <language ident="gez"/>
+            <language ident="gez">Gǝʿǝz</language>
          </langUsage>
       </profileDesc>
       <revisionDesc>
          <change when="2024-04-22" who="ES">created stub</change>
+         <change when="2026-04-15" who="CH">deleted dead facsimile links</change>
       </revisionDesc>
    </teiHeader>
-   <facsimile>
```
eu-genia, line 113: the empty facsimile should of course be deleted, the second one works perfectly and should stay

eu-genia: (note that we too will probably put cloudflare on, just like princeton did, it means you have to wait, not that the links are dead)

eu-genia:
I reinstated the correct link to the main repository of the images and added the msDesc from PEMM

eu-genia:
remember to always keep your branches in sync with the remote or you risk overwriting or deleting something (github tries to make sure this does not happen but it cannot always know which version should be taken if you save an old version over a newer one). always fetch and pull in github desktop **before** any pushes or commits.

PR #3353, opened 2026-04-17
Update ESdd033.xml

MagdaKrzyz:


eu-genia:
The change is correct but you will have no visualization because of the quire irregularity

https://github.com/BetaMasaheft/Documentation/issues/3300#issuecomment-4267611730

ES/ESdd033.xml
```diff
@@ -59,6 +59,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                                     <dim unit="leaf">1</dim>
                                     <locus target="#1r"/>
                                     1, stub after 8 
+                                    <note>The stub of <locus target="#1r"/> is inserted after folium 8 of <ref target="#q2"/>. It is inserted sideways and originates
```
eu-genia: if you want to trick visColl so that visualization stays you can also say

```
                               <item xml:id="q1" n="A">
                                    <dim unit="leaf">1</dim>
                                    <locus target="#1r"/>
                                    1   (The respective stub is inserted after the next quire. The protective leaf is inserted sideways and originates from an earlier codex)
                                 </item>
```

Also note that now it sounds as if only stub is inserted sideways, you must specify that the entire leaf is

eu-genia (approved):

eu-genia (changes requested):
On rethinking. I would prefer the collation corrected to q1+q2 becoming q1

ES/ESdd033.xml
```diff
@@ -55,82 +55,78 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
                            <collation>
                               <list>
-                                 <item xml:id="q1" n="A">
-                                    <dim unit="leaf">1</dim>
-                                    <locus target="#1r"/>
-                                    1, stub after 8 
+                                 <item xml:id="q1" n="1">
+                                    <dim unit="leaf">9</dim>
+                                    <locus from="1r" to="9v"/>
+                                    1, stub after 9
```
eu-genia: you can add here the note that the singleton is a recycled endleaf

ES/ESdd033.xml
```diff
@@ -178,9 +174,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         <placeName ref="INS0105DD"/>
                      </origPlace>
                      <origDate notBefore="1400" notAfter="1800" evidence="lettering" cert="medium">The manuscript consists of two codicological units. 
-                        The fragment referred to as <ref target="#p1"/> originates from a different, older, and larger manuscript dating to the 15th century. 
-                        Because of the large size of the folium, it was inserted sideways on the right, so the text appears perpendicular to the main writing surface.
-                        The folium has been used as a guard leaf (<ref target="#q1"/>). 
+                        The fragment referred to as <ref target="#p1"/> originates from a different, older, and larger manuscript dating to the 15th century. It is a single 
+                        folium with a stub, <locus target="#1"/>, which belongs to <ref target="#q1"/> and serves as a guard leaf.
```
eu-genia: on terminology (endleaf) see https://github.com/BetaMasaheft/Documentation/issues/2503

ES/ESdd033.xml
```diff
@@ -253,8 +248,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         </origPlace>
                         <origDate notBefore="1400" notAfter="1499">
                           <ref target="#p1"/> originates from a different, older, and larger manuscript dating to the 15th century. 
-                           Because of the large size of the folium, it was inserted sideways on the right, so the text appears perpendicular to the main writing surface.
-                           The folium has been used as a guard leaf (<ref target="#q1"/>).  
+                            It is a single folium with a stub, <locus target="#1"/>, which belongs to <ref target="#q1"/> and serves as a guard leaf.
```
eu-genia: endleaf

MagdaKrzyz: I understand that Gumpert wanted the term to be as neutral as possible, even though calling a leaf that is at the beginning an "endleaf" is strange. But here, I want to stress its function. If you think we should abandon this practice since we do not know what the function is (but we know) then it would be good to change the terms in the guidelines as well.

eu-genia (approved):

MagdaKrzyz:
The visualization failed:
image

eu-genia:
the note disturbed the calculation, changed to text

eu-genia:
BTW `"The margins of #q1 have been trimmed."` in condition - does it refer to the new q1 or to the old q1, that is fol. 1

PR #3357, opened 2026-04-21
Update

MagdaKrzyz:
I was wrong to delete the word "endband" in those files. I have now restored the word.


CarstenHoffmannMarburg (approved):
I would suggest to add change notes to each record even in cases of small changes.

MagdaKrzyz:
uuuu... I will do it next week from the computer where I opened all the files.

e-schwarz (approved):

MagdaKrzyz:
I added info about the change.

PR #3379, opened 2026-05-12
Update ESmakm012.xml

MagdaKrzyz:


ES/ESmakm012.xml
```diff
@@ -74,7 +74,8 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                      </msItem>
                
                      <explicit type="embedded" xml:lang="gez" corresp="LIT3989Aragawi#NoteTranslationExplicit">
-                        <locus target="#180va"/>ተተርጐመ፡ ዝንቱ፡ መጽሐፍ፡ እምልሳነ፡ ዓረቢ፡ ኅበ፡
+                        <locus target="#180va"/>
+                        <foreign xml:lang="gez">ተተርጐመ፡ ዝንቱ፡ መጽሐፍ፡ እምልሳነ፡ ዓረቢ፡ ኅበ፡
```
CarstenHoffmannMarburg: This explicit is already marked with xml:lang="gez". I think, it is not neccessary to add foreign.

MagdaKrzyz: Sure, so please look in the app how it is visualized. The three texts are displayed together without any boundaries. How to improve this?

ES/ESmakm012.xml
```diff
@@ -255,7 +257,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         </item>
                         
                         <item xml:id="e1"> 
-                           <locus target="#2vb #3vb #5ra #6va #12va #14rab #15rab #15vb #16rab #19vb #22vab #14rb #24vb #31rb #46rab #57rb #87vb #88rab #93ra #123rab #130rab #139rab #143 rab #146rab #159rab"/>
+                           <locus target="#2vb #3vb #5ra #6va #12va #14rab #15rab #15vb #16rab #19vb #22vab #14rb #24vb #31rb #46rab #57rb #87vb #88rab #93ra #123rab #130rab #139rab #143rab #146rab #159rab"/>
```
CarstenHoffmannMarburg: I have the suspicion, that 14rab is an incorrect target, because it refers to both 143ra and 143rb. I think, it would be correct to refer to either both columns (#143ra #143rb) or to the full page (#14r). The same for the other elements.

CarstenHoffmannMarburg (approved):

ES/ESmakm012.xml
CarstenHoffmannMarburg: I am not sure, whether this change will solve the situation or whether it will rather stay the same. Another option, which comes to my mind is to encode the text and its translation as a div in body. However, in this case, the information would be hidden and retrievable behind the button 'Text'. Hardly anybody would find it. To my opinion, it will be better to leave it as it is.

ES/ESmakm012.xml
MagdaKrzyz: I see, good. I will change it.

MagdaKrzyz: I have checked how it works and actually when you click on a page number it will not take you to the page you want anyway. Also when you click a correct number like 12v. How about 12rv? I need to split it too?

MagdaKrzyz: Look what I did.

ES/ESmakm012.xml
```diff
@@ -255,7 +257,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         </item>
                         
                         <item xml:id="e1"> 
-                           <locus target="#2vb #3vb #5ra #6va #12va #14rab #15rab #15vb #16rab #19vb #22vab #14rb #24vb #31rb #46rab #57rb #87vb #88rab #93ra #123rab #130rab #139rab #143 rab #146rab #159rab"/>
+                           <locus target="#2vb #3vb #5ra #6va #12va #14r #15r #15vb #16r #19vb #22v #14rb #24vb #31rb #46r #57rb #87vb #88r #93ra #123r #130r #139r #143r #146r #159r"/>
```
CarstenHoffmannMarburg, line 260: I have the suspicion, that 14rab is an incorrect target, because it refers to both 143ra and 143rb. I think, it would be correct to refer to either both columns (#143ra #143rb) or to the full page (#14r). The same for the other elements.

ES/ESmakm012.xml
CarstenHoffmannMarburg: I see, possibly there are no practical consequences from this change. However, I am more comfortable with your latest encoding.

CarstenHoffmannMarburg (approved):

PR #3393, opened 2026-05-24
Update CBD931.xml

e-schwarz:


e-schwarz:
Unfortunately, there are no images of the ms, and Cerulli does not quote the Image of Raphael, therefore, I could not encode it.

Dublin/CBD931.xml
```diff
@@ -21,40 +24,217 @@ schematypens="http://relaxng.org/ns/structure/1.0"
...
+                        <title ref="LIT4155MiraclesMichael#Miracle11"/>
+                     </msItem>
+                     <msItem xml:id="ms_i2.12">
+                        <locus from="141r"/>
+                        <title ref="LIT4155MiraclesMichael#Miracle12"/>
+                     </msItem>
+                  </msItem>
+                  <msItem xml:id="ms_i3">
+                     <locus from="143v"/>
+                     <title>Malkǝʾa Rufāʾel</title>
```
CarstenHoffmannMarburg: I think the indication is good enough to mark it up with LIT0255Malke.

e-schwarz: thank you, done

CarstenHoffmannMarburg (approved):

PR #3436, opened 2026-06-24
Update Hasura Masqal manuscript

CarstenHoffmannMarburg:
I updated three records with LIT5696MagicPrayer, which is a general record for Hasura masqal manuscripts according to my suggestion in https://github.com/BetaMasaheft/Documentation/issues/3350.

I found that the dimensions must be incorrect and I will open an issue for it in a minute.

### DIFF
--- a/EMIP/magicscrolls/EMIPms00484/EMIPms00484.xml
+++ b/EMIP/magicscrolls/EMIPms00484/EMIPms00484.xml
@@ -34,6 +34,31 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                <msContents>
                   <summary/>
                   <msItem xml:id="ms_i1">
+                     <title ref="LIT5696MagicPrayer"/>
+                     <textLang mainLang="gez"/>
+                  </msItem>
+                  <msItem xml:id="ms_i2">
+                     <title ref="LIT5888AsmatPrayer"/>
+                     <textLang mainLang="gez"/>
+                  </msItem>
+                  <msItem xml:id="ms_i3">
+                     <title ref="LIT5888AsmatPrayer"/>
+                     <textLang mainLang="gez"/>
+                  </msItem>
+                  <msItem xml:id="ms_i4">
+                     <title ref="LIT5888AsmatPrayer"/>
+                     <textLang mainLang="gez"/>
+                  </msItem>
+                  <msItem xml:id="ms_i5">
+                     <title ref="LIT5888AsmatPrayer"/>
+                     <textLang mainLang="gez"/>
+                  </msItem>
+                  <msItem xml:id="ms_i6">
+                     <title ref="LIT5888AsmatPrayer"/>
+                     <textLang mainLang="gez"/>
+                  </msItem>
+                  <msItem xml:id="ms_i7">
+                     <title ref="LIT5888AsmatPrayer"/>
                      <textLang mainLang="gez"/>
                   </msItem>
                </msContents>
@@ -47,7 +72,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                         <extent>
                            <measure unit="strips">3</measure>
                            <dimensions type="outer" unit="mm">
-                              <height>186</height>
+                              <height>186</height><!-- these dimensions can not be correct -->
                               <width>8.7</width>
                               <depth/>
                            </dimensions>
@@ -88,6 +113,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <change when="2022-09-23" who="RL">Corrected include code errors</change>
          <change when="2022-09-26" who="RL">Removed text body code lines</change>
          <change when="2024-08-28" who="RL">formatted text and facsimile files</change>
+         <change when="2026-06-24" who="CH">Add msContents with LIT5696MagicPrayer</change>
       </revisionDesc>
    </teiHeader>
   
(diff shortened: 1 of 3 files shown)


EMIP/magicscrolls/EMIPms00510/EMIPms00510.xml
```diff
@@ -31,9 +31,34 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
                <msContents>
-                  <summary> </summary>
+                  <summary>Magic prayers</summary>
                   <msItem xml:id="ms_i1">
-                     <textLang mainLang="gez">Gǝ‘ǝz</textLang>
+                     <title ref="LIT5888AsmatPrayer"/>
+                     <textLang mainLang="gez"/>
+                  </msItem>
+                  <msItem xml:id="ms_i2">
+                     <title ref="LIT5888AsmatPrayer"/>
```
MagdaKrzyz, line 40: In such cases, when we give a general record, we should also provide an incipt here. Otherwise, we will not be able to identify the prayer at all. The same for the rest of the texts here

EMIP/magicscrolls/EMIPms00532/EMIPms00532.xml
```diff
@@ -32,8 +32,25 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                   <idno facs="EMIP/Scrolls/MS0532/" n="7">Tweed MS 64</idno>
                </msIdentifier>
                <msContents>
-                  <summary/>
+                  <summary>Magic prayers</summary>
                   <msItem xml:id="ms_i1">
+                     <title ref="LIT5888AsmatPrayer"/>
```
MagdaKrzyz, line 37: These are not your descriptions, but to my mind the general record without any incipt does not bring much.

EMIP/magicscrolls/EMIPms00484/EMIPms00484.xml
```diff
@@ -34,6 +34,31 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                <msContents>
                   <summary/>
                   <msItem xml:id="ms_i1">
+                     <title ref="LIT5696MagicPrayer"/>
+                     <textLang mainLang="gez"/>
+                  </msItem>
+                  <msItem xml:id="ms_i2">
+                     <title ref="LIT5888AsmatPrayer"/>
```
MagdaKrzyz, line 41: In such cases, when we give a general record, we should also provide an incipt here. Otherwise, we will not be able to identify the prayer at all. The same for the rest of the texts here.

EMIP/magicscrolls/EMIPms00510/EMIPms00510.xml
CarstenHoffmannMarburg: Citing meaningful incipits for each of the EMIP manuscripts—which I intend to update here and elsewhere—would mean a significant amount of work. The sole aim of this pull request is to add the entry "LIT5696MagicPrayer" to those records where I have identified such a prayer. Furthermore, a thorough revision of the content descriptions for all EMIPms records would require investigating potentially identical or similar texts in other records. That workload would be excessive and is not currently on our list of priorities. The full texts are available as Transkribus data; consequently, they are more easily searchable than other prayer texts—even though the majority of the records lack the usual level of descriptive precision. I believe @eu-genia is skeptical about investing too much energy in updating EMIPms-records as she made clear recently for another matter (cf. https://github.com/BetaMasaheft/Documentation/issues/3353).

eu-genia: I am more than sceptical. This is not our task as Bm.

MagdaKrzyz: First I made a remark about LIT5888AsmatPrayer, not about LIT5696MagicPrayer. 
I suggested only to add incipits so that when someone comes across a similar text, can better identify it.
Otherwise, I do not understand what for to add the Ids at all: in magic scrolls we can expect magic prayers and asmats. 
If this is not the task of BM this is also fine for me.

MagdaKrzyz (approved):

EMIP/magicscrolls/EMIPms00510/EMIPms00510.xml
eu-genia: It is not the task of Bm to edit EMIP records unless there are errors that hurt us

MagdaKrzyz: I see, so there were errors that hurt us and Carsten was supposed to remove them. Great!

CarstenHoffmannMarburg: I believe it is a valuable information to mark identical prayers in other manuscripts to make clear, that a certain work does not occur only once.

I wanted to add LIT5696MagicPrayer and I wanted to make clear that it is the first out of 7 prayers. If I do not add ms_i2 to ms_i7, but merely add LIT5696MagicPrayer to ms_i1, the user would come to the conclusion, that the entire scroll has only one long prayer, what is not the case. This is why I put LIT5888AsmatPrayer or LIT6609ProtectPrayer as provisorical placeholders. Other researchers, who want to describe the texts more precisely can build upon this structure.

If you think, that this is not the best way to indicate it, I can also come to add a note in the summary instead, saying that it is one out of several magic prayers. In this case, I would express verbally, what I can describe analytically. This might even be quicker.

PR #3490, opened 2026-08-07
Update BLorient11294#ms_i1

CarstenHoffmannMarburg:
According to Magdas request in https://github.com/BetaMasaheft/Documentation/issues/3375#issuecomment-5219464538.

### DIFF
--- a/LondonBritishLibrary/orient/BLorient11294.xml
+++ b/LondonBritishLibrary/orient/BLorient11294.xml
@@ -33,9 +33,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     <msContents><summary/>
                         <msItem xml:id="ms_i1">
                             <locus from="2ra" to="57va"/>
-                            <title ref="LIT2048Nagara"/>
+                            <title>Text on incarnation</title>
                             <textLang mainLang="am" otherLangs="gez"/>       
-                            <note>The spaces intended for the rubrics have been left blank on <locus from="5r" to="56r"/>.</note>
+                            <note>Different to <ref type="work" corresp="LIT2048Nagara"/>. The spaces intended for the rubrics have been left
+                                blank on <locus from="5r" to="56r"/>.</note>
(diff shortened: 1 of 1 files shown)


LondonBritishLibrary/orient/BLorient11294.xml
```diff
@@ -33,9 +33,10 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     <msContents><summary/>
                         <msItem xml:id="ms_i1">
                             <locus from="2ra" to="57va"/>
-                            <title ref="LIT2048Nagara"/>
+                            <title>Text on incarnation</title>
```
eu-genia: It would be safer to restore the label provided by the cataloguer, as we are supposed to reproduce the catalogue by Strelcyn unless we have strong reasons to update/modify. I guess we do not have the images for this one and the incipit does sound a bit generic.

Can you check the catalogue again?

(In the past we occasionally also referred to texts referring to one of the Pillars with a sub-ID (e.g. `LIT1097Ammest#Incarnation`) as it happens that only some of the Pillars are present)

eu-genia:
Note also that for another ms you created a new ID LIT6940IncarnationAmh - also to be checked as the question-answer structure is the one of the "standard" AAM texts (they are a catechism)

MagdaKrzyz:
I would like Carsten to wait a bit until I finish my work on the Pillars. Perhaps then I will be able to tell a bit more about the text.

LondonBritishLibrary/orient/BLorient11294.xml
CarstenHoffmannMarburg: The description by Strelcyn is the following:

grafik

I will update the title and the note.

LondonBritishLibrary/orient/BLorient11294.xml
```diff
@@ -33,9 +33,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     <msContents><summary/>
                         <msItem xml:id="ms_i1">
                             <locus from="2ra" to="57va"/>
-                            <title>Text on incarnation</title>
+                            <title>Nagara hāymānot</title>
                             <textLang mainLang="am" otherLangs="gez"/>       
-                            <note>Different to <ref type="work" corresp="LIT2048Nagara"/>. The spaces intended for the rubrics have been left
+                            <note>According to <bibl><ptr target="bm:Strelcyn1978BritishLibrary"/><citedRange unit="page">80</citedRange>
```
eu-genia, line 38: OK but let us wait for Magda as she has asked

PR #3515, opened 2026-09-07
Update ESdd009.xml

MagdaKrzyz:
Minor corrections

### DIFF
--- a/ES/ESdd009.xml
+++ b/ES/ESdd009.xml
@@ -115,7 +115,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         <locus from="109rb" to="113ra"/>
                         <title type="complete">Homily for Sunday</title>
                      </msItem>
-                     <msItem xml:id="ms_i2.10">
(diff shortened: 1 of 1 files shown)


ES/ESdd009.xml
```diff
@@ -115,7 +115,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                         <locus from="109rb" to="113ra"/>
                         <title type="complete">Homily for Sunday</title>
                      </msItem>
-                     <msItem xml:id="ms_i2.10">
+                     <msItem xml:id="ms_i3">
```
CarstenHoffmannMarburg, line 120: When you want to define this msItem as ms_i3, you need to close ms_i2.

CarstenHoffmannMarburg (changes requested):

ES/ESdd009.xml
MagdaKrzyz: done

CarstenHoffmannMarburg (approved):

